PR #18, opened 2021-05-21
Moving the ODFE Settings to Legacy Settings and adding the new settings compatible with Opensearch 

aditjind:
*Issue #, if available:* Resolves 

- [x] Rename, and Backwards Compatibility for Settings from #14 

*Description of changes:* 

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/settings/LegacyOpenDistroAlertingSettings.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/settings/LegacyOpenDistroAlertingSettings.kt
@@ -1,15 +1,12 @@
-package com.amazon.opendistroforelasticsearch.alerting.settings
+package org.opensearch.alerting.settings
 
-import com.amazon.opendistroforelasticsearch.alerting.AlertingPlugin
 import org.opensearch.common.settings.Setting
 import org.opensearch.common.unit.TimeValue
 import java.util.concurrent.TimeUnit
 
-
 /**
  * Legacy Opendistro settings specific to [AlertingPlugin]. These settings include things like history index max age, request timeout, etc...
  */
-
 
 class LegacyOpenDistroAlertingSettings {
 
@@ -101,10 +98,12 @@ class LegacyOpenDistroAlertingSettings {
 
         val MAX_ACTION_THROTTLE_VALUE = Setting.positiveTimeSetting(
             "opendistro.alerting.action_throttle_max_value",
-            Setting.Property.NodeScope, Setting.Property.Dynamic, Setting.Property.Deprecated)
+            Setting.Property.NodeScope, Setting.Property.Dynamic, Setting.Property.Deprecated
+        )
 
         val FILTER_BY_BACKEND_ROLES = Setting.boolSetting(
             "opendistro.alerting.filter_by_backend_roles",
-            Setting.Property.NodeScope, Setting.Property.Dynamic, Setting.Property.Deprecated)
+            Setting.Property.NodeScope, Setting.Property.Dynamic, Setting.Property.Deprecated
+        )
     }
 }
--- a/alerting/src/main/kotlin/org/opensearch/alerting/settings/LegacyOpenDistroAlertingSettings.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/settings/LegacyOpenDistroAlertingSettings.kt
@@ -51,7 +51,7 @@ class LegacyOpenDistroAlertingSettings {
         )
 
         val ALERT_BACKOFF_MILLIS = Setting.positiveTimeSetting(
-            "opensearch.alerting.alert_backoff_millis",
+            "opendistro.alerting.alert_backoff_millis",
             TimeValue.timeValueMillis(50),
             Setting.Property.NodeScope, Setting.Property.Dynamic, Setting.Property.Deprecated
         )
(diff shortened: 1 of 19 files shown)


core/src/main/kotlin/org/opensearch/alerting/core/settings/LegacyOpenDistroScheduledJobSettings.kt
```diff
@@ -0,0 +1,48 @@
...
+
+class LegacyOpenDistroScheduledJobSettings {
+
+    companion object {
+        val SWEEPER_ENABLED = Setting.boolSetting(
+            "opendistro.scheduled_jobs.enabled",
+            true,
+            Setting.Property.NodeScope, Setting.Property.Dynamic, Setting.Property.Deprecated
+        )
+        val REQUEST_TIMEOUT = Setting.positiveTimeSetting(
```
dblock, line 18: Add a line break above ;)

core/src/main/kotlin/org/opensearch/alerting/core/settings/ScheduledJobSettings.kt
```diff
@@ -37,32 +36,38 @@ class ScheduledJobSettings {
...
+            "plugins.scheduled_jobs.sweeper.backoff_millis",
+            LegacyOpenDistroScheduledJobSettings.SWEEP_BACKOFF_MILLIS,
+            Setting.Property.NodeScope, Setting.Property.Dynamic
+        )
 
         val SWEEP_BACKOFF_RETRY_COUNT = Setting.intSetting(
-                "opendistro.scheduled_jobs.retry_count",
-                3,
-                Setting.Property.NodeScope, Setting.Property.Dynamic)
+            "plugins.scheduled_jobs.retry_count",
```
dblock, line 56: I know this is not new, but these names scare me. We have jobscheduler.retry_count in https://github.com/opensearch-project/job-scheduler/blob/1eb8e28823c20de3d3840f3fe755f21a11c5e301/src/main/java/org/opensearch/jobscheduler/JobSchedulerSettings.java#L44 which sounds awfully similar.

I wonder whether this is the time we change it to something more specific to alerting? By adding `alerting` into the name for example?

core/src/main/kotlin/org/opensearch/alerting/core/settings/LegacyOpenDistroScheduledJobSettings.kt
aditjind: Adding klint formatting to the whole project. It will take care of all these changes for future. ;)

skkosuri-amzn: @aditjind we got ktlint enabled as part of gradle build. Is that missing now?

alerting/src/main/kotlin/org/opensearch/alerting/settings/LegacyOpenDistroAlertingSettings.kt
```diff
@@ -0,0 +1,109 @@
...
+            "opendistro.alerting.request_timeout",
+            Setting.Property.NodeScope, Setting.Property.Dynamic, Setting.Property.Deprecated
+        )
+
+        val MAX_ACTION_THROTTLE_VALUE = Setting.positiveTimeSetting(
+            "opendistro.alerting.action_throttle_max_value",
+            Setting.Property.NodeScope, Setting.Property.Dynamic, Setting.Property.Deprecated
+        )
+
+        val FILTER_BY_BACKEND_ROLES = Setting.boolSetting(
```
lezzago, line 104: Just want to confirm that the setting is false by default if it is not set.

core/src/main/kotlin/org/opensearch/alerting/core/settings/ScheduledJobSettings.kt
skkosuri-amzn: After the chasing the history, the origin design goal is to have same settings to all Job Scheduler based plugins. The central idea seems to be one set of settings to control all jobs run by various plugins. Currently, Alerting, ISM, AD use the same settings. Second part is, when scheduling code is seperated from Alerting to a seperate plugin (Job Schedular) the duplicate code is not removed from alerting. Created https://github.com/opensearch-project/alerting/issues/19 to track this post GA.

skkosuri-amzn: @dblock, I am fine to keep status-quo for this based on my above analysis.

skkosuri-amzn (approved):

core/src/main/kotlin/org/opensearch/alerting/core/settings/ScheduledJobSettings.kt
dblock: But aside of duplicate code the actual setting names are different:

`jobscheduler.retry_count`
`scheduled_jobs.retry_count`

This wasn't the original intent. When did we break it and what do we do about it?

skkosuri-amzn: Good callout 👍 . Setting names are different, this is not broken (compared with odfe versions). My 2C is keep status-quo in view of ODFE.1.13-OpenSearch.1.0 compatability and address the names as part of #19

dblock: Your call. I was just thinking it might be easier to do this now since we're introducing a new setting that fallback to a legacy one.

core/src/main/kotlin/org/opensearch/alerting/core/settings/LegacyOpenDistroScheduledJobSettings.kt
aditjind: Updating the version and applied the formatting. (https://github.com/opensearch-project/alerting/pull/20)

lezzago (approved):

Issue #94, opened 2021-06-02
[BUG]"Propagate exception from publishing Email notifications to Action run result"


aditjind:
<a href="https://github.com/skkosuri-amzn"><img src="https://avatars.githubusercontent.com/u/59816283?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [skkosuri-amzn](https://github.com/skkosuri-amzn)**
_Monday Mar 01, 2021 at 18:06 GMT_
_Originally opened as https://github.com/opendistro-for-elasticsearch/alerting/issues/359_

----

**Describe the bug**
Propagate exception from publishing Email notifications to Action run result

**To Reproduce**
1. Configure non-reachable email smtp server.
2. Configure and Trigger an alert.
3. Alert will show as Active even though it failed to send the email.

**Expected behavior**
1. Show alert status as Error and show error message.

Issue #74, opened 2021-06-02
Support date math in index names


aditjind:
<a href="https://github.com/smowky"><img src="https://avatars.githubusercontent.com/u/36858658?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [smowky](https://github.com/smowky)**
_Wednesday Sep 23, 2020 at 10:05 GMT_
_Originally opened as https://github.com/opendistro-for-elasticsearch/alerting/issues/254_

----

**Is your feature request related to a problem? Please describe.**
I’m using elasticsearch ILM to rotate my indexes and now alerts are searching through all indexes metricbeat-* which is quite expensive operation.
It would be nice to have a possibility to search only for last 2 day indexes instead

**Describe the solution you'd like**
Would be nice define search index as:
<logstash-{now/d-2d}>,<logstash-{now/d-1d}>,<logstash-{now/d}>
or
%3Clogstash-%7Bnow%2Fd-2d%7D%3E%2C%3Clogstash-%7Bnow%2Fd-1d%7D%3E%2C%3Clogstash-%7Bnow%2Fd%7D%3E

**Describe alternatives you've considered**
Alternative solutions is using aliases , which is be default not as easy since elasticsearch ILM actions does not have a possibility do add/remove alias

aditjind:
<a href="https://github.com/aditjind"><img src="https://avatars.githubusercontent.com/u/13850971?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [aditjind](https://github.com/aditjind)**
_Wednesday Jun 02, 2021 at 03:34 GMT_

----

Hi @smowky, 

ES already has the support [built in](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/date-math-index-names.html) for Date Math Index Name Resolution to search over a range of indices. Since alerting uses the [Node Client](https://www.javadoc.io/doc/org.elasticsearch/elasticsearch/5.0.0/org/elasticsearch/client/support/AbstractClient.html#execute-org.elasticsearch.action.Action-Request-) to index and search this should be possible with-in the current implementation. 

Please check once with our latest release.

aditjind:
Hi @smowky

[...]

PR #113, opened 2021-06-20
Adding the ITs for the support of Date Math Index Resolution and a fix for moving the alert to the error state in case of not being able to publish the emails

aditjind:
*Issue #, if available:* #74 https://github.com/opensearch-project/alerting/issues/94

*Description of changes:* A null check was required to prevent the destinations which don't have a URL or a HOST as a part of the Destination Params. This resulted in validation failures whenever Destinations such as SNS are added. Added the null checks and also updated the UTs

An urgent update (hot fix) was done as a part of [this](https://github.com/opensearch-project/alerting/commit/496c9546889cdad5f30c483cf04d18b870531460) commit to unblock.

This also updates the integration test to create the index which is encoded in date math format. (#74)

This also contains the fix for storing the alert in the error condition whenever the action is unable to send the email to the destination.  

Changes are tested locally by hitting the APIs and checking the response of the unreachably configured SMPT server. 

*CheckList:*

### DIFF
--- a/alerting/src/test/kotlin/org/opensearch/alerting/MonitorRunnerIT.kt
+++ b/alerting/src/test/kotlin/org/opensearch/alerting/MonitorRunnerIT.kt
@@ -52,6 +52,7 @@ import org.opensearch.index.query.QueryBuilders
 import org.opensearch.rest.RestStatus
 import org.opensearch.script.Script
 import org.opensearch.search.builder.SearchSourceBuilder
+import java.net.URLEncoder
 import java.time.Instant
 import java.time.ZonedDateTime
 import java.time.format.DateTimeFormatter
@@ -59,6 +60,7 @@ import java.time.temporal.ChronoUnit
 import java.time.temporal.ChronoUnit.DAYS
 import java.time.temporal.ChronoUnit.MILLIS
 import java.time.temporal.ChronoUnit.MINUTES
+import kotlin.collections.HashMap
 
 class MonitorRunnerIT : AlertingRestTestCase() {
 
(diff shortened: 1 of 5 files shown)


alerting/src/test/kotlin/org/opensearch/alerting/MonitorRunnerIT.kt
```diff
@@ -83,11 +85,14 @@ class MonitorRunnerIT : AlertingRestTestCase() {
     }
 
     fun `test execute monitor returns search result`() {
-        val testIndex = createTestIndex()
+        val testIndex = "<my-index-{now/d{YYYY.MM.dd|+12:00}}>"
+        val encodedTestIndex = createTestIndex(
+            URLEncoder.encode(testIndex, "utf-8")
+        )
```
lezzago, line 91: Can you make this into its own separate test so we know we are testing for this specific functionality?

lezzago (changes requested):

alerting/src/test/kotlin/org/opensearch/alerting/MonitorRunnerIT.kt
aditjind: Although I'd be happy to, but I don't see the value addition in adding more code (redundant?). We already have the standard index names tested via other tests. Thoughts?

lezzago: Personally I think the redundancy here is ok as its to show 2 different use cases and to a new developer, its clearer what we are testing in case they need to make changes to it later on. Also if many of the tests are repeating the code, they we can abstract out that to another function to reduce redundancy.

aditjind: Fair Point. Have added the changes to another test which was named Date math but was not using the date math indices. This might have been a miss while the test would have been written.

lezzago (approved):

Issue #129, opened 2021-07-21
Clear usages of ActionGet API call


aditjind:
**Describe the bug**

The usage of `actionGet` is not recommended since it blocks the thread and if the program is executing in a single thread it can create a deadlock.

PR #130, opened 2021-07-21
Removing All Usages of Action Get Method Calls and adding the listeners

aditjind:
*Issue #, if available:* #129 

*Description of changes:*

This commit addresses the usages of the actionGet API calls which are blocking in nature and block the execution of the program. This is replaced by the callback listeners for the API call response. 

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/alerts/AlertIndices.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/alerts/AlertIndices.kt
@@ -27,9 +27,10 @@
 package org.opensearch.alerting.alerts
 
 import org.apache.logging.log4j.LogManager
-import org.apache.lucene.index.IndexNotFoundException
 import org.opensearch.ResourceAlreadyExistsException
+import org.opensearch.action.ActionListener
 import org.opensearch.action.admin.cluster.state.ClusterStateRequest
+import org.opensearch.action.admin.cluster.state.ClusterStateResponse
 import org.opensearch.action.admin.indices.alias.Alias
 import org.opensearch.action.admin.indices.create.CreateIndexRequest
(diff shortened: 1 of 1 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/alerts/AlertIndices.kt
```diff
@@ -331,26 +358,48 @@ class AlertIndices(
...
+                    override fun onResponse(acknowledgedResponse: AcknowledgedResponse?) {
+                        if (acknowledgedResponse != null) {
+                            if (!acknowledgedResponse.isAcknowledged) {
+                                logger.error("Could not delete one or more Alerting history indices: $index")
+                            }
                         }
-                    } catch (e: IndexNotFoundException) {
+                    }
+                    override fun onFailure(e: Exception) {
                         logger.debug("$index was already deleted. ${e.message}")
```
dbbaughe, line 399: Do you want to wrap this in a if IndexNotFoundException check? Otherwise seems we say it was already deleted on every exception.

aditjind: I too thought about it but then on a delete index request we don't anticipate any other exception apart from index deleted exception. If the index is not deleted we get the response as `Acknowledged False`.

dbbaughe: Hmm but what about network exceptions and timeouts and others?

aditjind: Right we can modify the log message keeping the exception same?

dbbaughe (approved):

PR #149, opened 2021-08-23
Add ActionExecutionPolicy

qreshi:

*Issue #, if available:*

*Description of changes:*
Adds ActionExecutionPolicy which will be used for Action configuration in Bucket-Level Monitors.

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/model/action/Action.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/model/action/Action.kt
@@ -48,25 +48,27 @@ data class Action(
     val messageTemplate: Script,
     val throttleEnabled: Boolean,
     val throttle: Throttle?,
-    val id: String = UUIDs.base64UUID()
+    val id: String = UUIDs.base64UUID(),
+    val actionExecutionPolicy: ActionExecutionPolicy = ActionExecutionPolicy.getDefaultConfiguration()
 ) : Writeable, ToXContentObject {
 
     init {
         if (subjectTemplate != null) {
-            require(subjectTemplate.lang == Action.MUSTACHE) { "subject_template must be a mustache script" }
+            require(subjectTemplate.lang == MUSTACHE) { "subject_template must be a mustache script" }
         }
-        require(messageTemplate.lang == Action.MUSTACHE) { "message_template must be a mustache script" }
+        require(messageTemplate.lang == MUSTACHE) { "message_template must be a mustache script" }
     }
 
     @Throws(IOException::class)
-    constructor(sin: StreamInput) : this(
+    constructor(sin: StreamInput): this(
         sin.readString(), // name
         sin.readString(), // destinationId
         sin.readOptionalWriteable(::Script), // subjectTemplate
         Script(sin), // messageTemplate
         sin.readBoolean(), // throttleEnabled
         sin.readOptionalWriteable(::Throttle), // throttle
-        sin.readString() // id
+        sin.readString(), // id
+        ActionExecutionPolicy(sin) // actionExecutionPolicy
     )
 
     override fun toXContent(builder: XContentBuilder, params: ToXContent.Params): XContentBuilder {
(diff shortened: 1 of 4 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/model/action/Action.kt
```diff
@@ -171,7 +179,8 @@ data class Action(
                 requireNotNull(messageTemplate) { "Action message template is null" },
                 throttleEnabled,
                 throttle,
-                id = requireNotNull(id)
+                id = requireNotNull(id),
+                actionExecutionPolicy = requireNotNull(actionExecutionPolicy) { "Action execution policy is null" }
```
rishabhmaurya, line 183: What happen when for old Actions which doesn't have the actionExecutionPolicy? Is this change backward compatible?

alerting/src/main/kotlin/org/opensearch/alerting/model/action/ActionExecutionScope.kt
```diff
@@ -0,0 +1,179 @@
...
+                                    ensureExpectedToken(Token.START_ARRAY, xcp.currentToken(), xcp)
+                                    val allowedCategories = AlertCategory.values().map { it.toString() }
+                                    while (xcp.nextToken() != Token.END_ARRAY) {
+                                        val alertCategory = xcp.text()
+                                        if (!allowedCategories.contains(alertCategory)) {
+                                            throw IllegalStateException("Actionable alerts should be one of $allowedCategories")
+                                        }
+                                        alertFilter.add(AlertCategory.valueOf(alertCategory))
+                                    }
+                                }
```
rishabhmaurya, line 72: What's the expectation in else block here? Shouldn't it fail if `perAlertFieldName` is not `ACTIONABLE_ALERTS_FIELD`?

alerting/src/main/kotlin/org/opensearch/alerting/model/action/ActionExecutionScope.kt
```diff
@@ -0,0 +1,179 @@
...
+                            }
+                        }
+                    }
+                    PER_EXECUTION_FIELD -> {
+                        type = Type.PER_EXECUTION
+                        while (xcp.nextToken() != Token.END_OBJECT) {}
```
rishabhmaurya, line 78: Is it for empty block?

alerting/src/main/kotlin/org/opensearch/alerting/model/action/Action.kt
qreshi: So Query-Level Monitors don't actually use the `ActionExecutionPolicy` as of now. If we make the change in the future to allow it, we'd update the default based on Monitor type to be compatible with both at that time.

alerting/src/main/kotlin/org/opensearch/alerting/model/action/ActionExecutionScope.kt
qreshi: Yeah, keeping this an empty block in case we did want to add configuration information for `PER_EXECUTION` in the future

alerting/src/main/kotlin/org/opensearch/alerting/model/action/ActionExecutionScope.kt
qreshi: Yeah, I'll add an `else` to make the `when` exhaustive.

alerting/src/main/kotlin/org/opensearch/alerting/model/action/Action.kt
rishabhmaurya: To be on a safer side, we should try to keep the behavior for existing query-level monitors as it is i.e. to not add any default field for `actionExecutionPolicy` in the data class for query-level monitors. This will ensure both forward and backward compatibility issues.

qreshi: Yeah, that makes sense. I'll merge in this change for now since it doesn't reflect the full usage of `ActionExecutionPolicy`. Once we get to the PR with the changes to `MonitorRunner`, I think we can update `actionExecutionPolicy` to be nullable and set the default configuration at runtime if it is not set.

rishabhmaurya (approved):

PR #163, opened 2021-08-31
Fix bug in paginating multiple bucket paths for Bucket-Level Monitor

qreshi:

*Issue #, if available:*

*Description of changes:*
For Bucket-Level Monitors, different Triggers can potentially have different bucket paths that they are evaluating the conditions against. For paginated aggregation queries, different bucket paths could have different page counts meaning that they will completed independently as the data is being paginated. This change makes it so the after_key can be independently kept for each bucket path to handle the case where one bucket path completes pagination while the others still have data to be iterated through.

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/util/AggregationQueryRewriter.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/util/AggregationQueryRewriter.kt
@@ -15,6 +15,7 @@ import org.opensearch.action.search.SearchResponse
 import org.opensearch.alerting.model.BucketLevelTrigger
 import org.opensearch.alerting.model.InputRunResults
 import org.opensearch.alerting.model.Trigger
+import org.opensearch.alerting.model.TriggerAfterKey
 import org.opensearch.search.aggregations.AggregationBuilder
 import org.opensearch.search.aggregations.AggregatorFactories
 import org.opensearch.search.aggregations.bucket.SingleBucketAggregation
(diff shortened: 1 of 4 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/util/AggregationQueryRewriter.kt
```diff
@@ -82,11 +87,37 @@ class AggregationQueryRewriter {
...
+                            if (prevTriggerAfterKey.lastPage) {
+                                // If the previous TriggerAfterKey already hit the last page, pass along the after key it used to get there
+                                bucketLevelTriggerAfterKeys[trigger.id] = prevTriggerAfterKey
+                            } else if (afterKey == null) {
+                                // If the previous TriggerAfterKey had not reached the last page and the after key for the current result
+                                // is null, then the last page has been reached so the after key that was used to get there is stored
+                                bucketLevelTriggerAfterKeys[trigger.id] = TriggerAfterKey(prevTriggerAfterKey.afterKey, true)
+                            } else {
+                                // Otherwise, update the after key to the current one
+                                bucketLevelTriggerAfterKeys[trigger.id] = TriggerAfterKey(afterKey, false)
```
aditjind, line 114: Do you think we should simplify and cleanup these if else blocks. Couple of solutions which come to my mind are abstracting it out in a utils function or maybe using a switch case or maybe ternary operators. I leave it upto you to make the decision.

aditjind (approved):

alerting/src/main/kotlin/org/opensearch/alerting/util/AggregationQueryRewriter.kt
qreshi: Updated it to a `when` statement

PR #174, opened 2021-09-16
Removing default integtest.sh.

dblock:
Coming from https://github.com/opensearch-project/opensearch-build/issues/497, removing default integtest.sh.


dblock:
The tests are failing for some other reason, is someone looking at them?

PR #199, opened 2021-10-08
Add backwards compatibility tests

qreshi:
*Issue #, if available:* #147

*Description of changes:*
* Adds backwards compatibility tests for mixed cluster, rolling upgrade and full restart scenarios
* Updates `DEVELOPER_GUIDE` to add instructions for running bwc tests
* Adds bwc tests to the multi-node GitHub Actions tests workflow

A couple improvements that can probably be made in future iterations:
* A hardcoded Monitor in the legacy format (ODFE 1.13.1) is being used for the request on the old version, this is because the `toXContent` method that is being used in the current version of the plugin test is not forwards compatible, so we would need a workaround or some legacy `toXContent` method for `Monitor` if we want to use the `randomMonitor()` helper method for creating the Monitor request
* We can manually update the Monitor execution time to artificially fast forward time when trying to execute the job again instead of sleeping since it inflate the test execution time
* **[DONE]** ~~We can generate the current version of the artifact dynamically and place it in the expected file path when running the bwc tests so we don't have to place an artifact there each time we move the versions along (ex. the 1.1 Alerting version being used is under `src/test/resources/bwc/alerting/1.1.0.0-SNAPSHOT/opensearch-alerting-1.1.0.0-SNAPSHOT.zip`)~~

*CheckList:*

### DIFF
--- a/.github/workflows/multi-node-test-workflow.yml
+++ b/.github/workflows/multi-node-test-workflow.yml
@@ -78,3 +78,8 @@ jobs:
           else
             echo "Security plugin is NOT available skipping this run as tests without security have already been run"
           fi
+
+      - name: Run Alerting Backwards Compatibility Tests
+        run: |
+          echo "Running backwards compatibility tests ..."
+          ./gradlew bwcTestSuite -Dtests.security.manager=false
--- a/.github/workflows/multi-node-test-workflow.yml
+++ b/.github/workflows/multi-node-test-workflow.yml
@@ -82,4 +82,4 @@ jobs:
       - name: Run Alerting Backwards Compatibility Tests
         run: |
           echo "Running backwards compatibility tests ..."
-          ./gradlew bwcTestSuite -Dtests.security.manager=false
+          ./gradlew bwcTestSuite
--- a/alerting/build.gradle
+++ b/alerting/build.gradle
@@ -88,7 +88,7 @@ dependencies {
     implementation "com.github.seancfoley:ipaddress:5.3.3"
 
     testImplementation "org.jetbrains.kotlin:kotlin-test:${kotlin_version}"
-    testCompile "org.mockito:mockito-core:2.23.0"
+    testCompile "org.mockito:mockito-core:3.12.4"
 }
 
 javadoc.enabled = false // turn off javadoc as it barfs on Kotlin code
--- a/alerting/build.gradle
+++ b/alerting/build.gradle
@@ -311,7 +311,7 @@ task bwcTestSuite(type: StandaloneRestIntegTestTask) {
     exclude '**/*IT*'
     dependsOn tasks.named("${baseName}#mixedClusterTask")
     dependsOn tasks.named("${baseName}#rollingUpgradeClusterTask")
-//    dependsOn tasks.named("${baseName}#fullRestartClusterTask")
+    dependsOn tasks.named("${baseName}#fullRestartClusterTask")
 }
 
 run {
--- a/alerting/build.gradle
+++ b/alerting/build.gradle
@@ -24,7 +24,9 @@
  *   permissions and limitations under the License.
  */
 
+import java.util.concurrent.Callable
 import org.opensearch.gradle.test.RestIntegTestTask
+import org.opensearch.gradle.testclusters.StandaloneRestIntegTestTask
 
 apply plugin: 'java'
 apply plugin: 'idea'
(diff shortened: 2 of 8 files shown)


lezzago:
Is there a reason we are using the snapshot version of alerting 1.1 zip?

qreshi:
> Is there a reason we are using the snapshot version of alerting 1.1 zip?

I updated the newest artifact to be generated automatically but I think it will always be the snapshot since that's the default for most recent point-in-time artifact being built.

alerting/src/test/kotlin/org/opensearch/alerting/bwc/AlertingBackwardsCompatibilityIT.kt
```diff
@@ -0,0 +1,220 @@
...
+                    createBasicMonitor()
+                }
+                ClusterType.MIXED -> {
+                    assertTrue(pluginNames.contains("opensearch-alerting"))
+                    verifyMonitorExists(LEGACY_OPENDISTRO_ALERTING_BASE_URI)
+                    // Waiting a minute to ensure the Monitor ran again at least once before checking if the job is running
+                    // on time
+                    // TODO: Should probably change the next execution time of the Monitor manually instead since this inflates
+                    //  the test execution by a lot
+                    Thread.sleep(60000)
```
skkosuri-amzn, line 72: Can we avoid sleep(...) ?

alerting/src/test/kotlin/org/opensearch/alerting/bwc/AlertingBackwardsCompatibilityIT.kt
```diff
@@ -0,0 +1,220 @@
...
+        val responseMap = getAsMap(uri)["nodes"] as Map<String, Map<String, Any>>
+        for (response in responseMap.values) {
+            val plugins = response["plugins"] as List<Map<String, Any>>
+            val pluginNames = plugins.map { plugin -> plugin["name"] }.toSet()
+            when (CLUSTER_TYPE) {
+                ClusterType.OLD -> {
+                    assertTrue(pluginNames.contains("opendistro-alerting"))
+                    createBasicMonitor()
+                }
+                ClusterType.MIXED -> {
```
skkosuri-amzn, line 65: How can we make sure that monitor has run both on old node and new node in mixed mode?

alerting/build.gradle
```diff
@@ -164,6 +166,151 @@ integTest {
...
     }
+
+    if (System.getProperty("tests.rest.bwcsuite") == null) {
+        filter {
+            excludeTestsMatching "org.opensearch.alerting.bwc.*IT"
+        }
+    }
+}
+
+String bwcVersion = "1.13.1.0"
```
lezzago, line 177: Do we want to add all the older versions (and add the zips) and test BWC against all of those? This would mean we also test against OS 1.0 and OS 1.1 since this is on OS 1.2.

alerting/src/test/kotlin/org/opensearch/alerting/bwc/AlertingBackwardsCompatibilityIT.kt
qreshi: I had originally planned on adding a utility method like what ISM had added:

https://github.com/opensearch-project/index-management/blob/ab122799283a58bb3f5c6f6d8576c7ba5875f13b/src/test/kotlin/org/opensearch/indexmanagement/indexstatemanagement/IndexStateManagementRestTestCase.kt#L382

As well as a `waitFor` wrapper so we could add assertions that could wait for the Monitor to execute and confirm state afterwards:

https://github.com/opensearch-project/index-management/blob/ab122799283a58bb3f5c6f6d8576c7ba5875f13b/src/test/kotlin/org/opensearch/indexmanagement/TestHelpers.kt#L133-L149

However, it looks like the JobScheduler implementation of the schedule information stores things like `start_time` which is convenient to modify for the particular scheduled job you'd like to speed up in test scenarios:

https://github.com/opensearch-project/job-scheduler/blob/main/spi/src/main/java/org/opensearch/jobscheduler/spi/schedule/IntervalSchedule.java#L54

Alerting's scheduled job implementation does not do this and instead uses `enabled_time` to resolve the next expected execution time only if the `expectedPreviousExecutionTime` is not set, otherwise it uses that:

https://github.com/opensearch-project/alerting/blob/e0e41df2b2e36c187edd3c45e214451208b10309/core/src/main/kotlin/org/opensearch/alerting/core/model/Schedule.kt#L332

The `expectedPreviousExecutionTime` is stored in `scheduledJobIdToInfo` which is part of a concurrent hashmap which is stored in memory as far as I'm aware:
https://github.com/opensearch-project/alerting/blob/e0e41df2b2e36c187edd3c45e214451208b10309/core/src/main/kotlin/org/opensearch/alerting/core/schedule/JobScheduler.kt#L52

This would make modification of an already running Monitor to execute sooner less trivial than expected. I'd like to leave it out of scope of this particular PR for now and have it be a possible follow-up for a test refactor.

alerting/src/test/kotlin/org/opensearch/alerting/bwc/AlertingBackwardsCompatibilityIT.kt
qreshi: We can't necessarily guarantee it has run on both old and new node in a mixed node cluster. However, this test is mostly ensuring the Monitor is still continuing to run and we are still able to access the API in a mixed cluster state.

alerting/src/test/kotlin/org/opensearch/alerting/bwc/AlertingBackwardsCompatibilityIT.kt
```diff
@@ -0,0 +1,220 @@
...
+import org.opensearch.alerting.LEGACY_OPENDISTRO_ALERTING_BASE_URI
+import org.opensearch.alerting.makeRequest
+import org.opensearch.alerting.model.Monitor
+import org.opensearch.common.settings.Settings
+import org.opensearch.common.xcontent.XContentType
+import org.opensearch.index.query.QueryBuilders
+import org.opensearch.rest.RestStatus
+import org.opensearch.search.builder.SearchSourceBuilder
+
+class AlertingBackwardsCompatibilityIT : AlertingRestTestCase() {
```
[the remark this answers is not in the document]
qreshi: We definitely want to cover at least ODFE 1.13 for now since it's a big change and an expected upgrade path to OpenSearch. As for the other versions, I've reached out to a couple folks so I can see how OpenSearch core does it so we can strive for parity with that. I still think the existing BWC tests will be useful for upcoming releases though, so we shouldn't necessarily block this PR on that, unless people prefer doing BWC tests manually until this is resolved.

.github/workflows/multi-node-test-workflow.yml
```diff
@@ -78,3 +78,8 @@ jobs:
           else
             echo "Security plugin is NOT available skipping this run as tests without security have already been run"
           fi
+
+      - name: Run Alerting Backwards Compatibility Tests
+        run: |
+          echo "Running backwards compatibility tests ..."
+          ./gradlew bwcTestSuite -Dtests.security.manager=false
```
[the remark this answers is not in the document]
qreshi: I had originally copied the command over from the AD plugin's example.

I did run this locally without `-Dtests.security.manager=false` and it does pass so looks like we can remove it.

alerting/src/test/kotlin/org/opensearch/alerting/bwc/AlertingBackwardsCompatibilityIT.kt
qreshi: Hmm, looks like these tests do get run automatically as part of `./gradlew build` so this would impact build times when running all tests. I was hoping to exclude them unless running something like `./gradlew bwcTestSuite` to avoid the overhead for the time being but `./gradlew build` will likely pick up all tests by nature.

skkosuri-amzn: Thanks for diving deep into this. Currently `gradlew build` takes around 4mins, this sleep would add one more min.

qreshi: Actually, since the `bwcTestSuite` executes the following scenarios:

1. `mixedCluster` - Hits the `MIXED` case
2. `rollingUpgrade` - Hits the `MIXED` and `UPGRADED` case
3. `fullRestart` - Hits the `UPGRADED` case

We'd be adding 4 minutes of sleep across all the tests since both the `MIXED` and `UPGRADED` cases are adding a sleep of 1 minute before verifying.

alerting/build.gradle
qreshi: Answered a similar question to one of @skkosuri-amzn's comments. We need to check OpenSearch core's BWC test matrix for older versions so we can have parity with that but for now I'm at least including against ODFE `1.13.1` since it's a common upgrade path which spans OpenDistro and OpenSearch.

skkosuri-amzn:
@qreshi Please confirm if `./grawdlew build  -x  bwcTestSuite` execludes the bwc tests.

qreshi:
> @qreshi Please confirm if `./grawdlew build -x bwcTestSuite` excludes the bwc tests.

I checked and it looks like it's not excluding the dependent tasks which run the BWC tests. This could be because they're defined as `StandaloneRestIntegTestTask`. We might have to exclude all the tasks explicitly.

skkosuri-amzn (approved):

lezzago (approved):

PR #201, opened 2021-10-11
Remove user from the responses

skkosuri-amzn:
*Issue #, if available:*
Remove user from the responses

*Description of changes:*
Remove user from the responses

All unit and integ passed. All security tests also passed. `./gradlew :alerting:integTest -Dtests.rest.cluster=localhost:9200 -Dtests.cluster=localhost:9200 -Dtests.clustername=opensearch -Dhttps=true -Dsecurity=true -Duser=admin -Dpassword=admin`

*CheckList:*


alerting/src/test/kotlin/org/opensearch/alerting/MonitorRunnerIT.kt
```diff
@@ -861,7 +861,7 @@ class MonitorRunnerIT : AlertingRestTestCase() {
             assertEquals("Alert not saved", 1, alerts.size)
             verifyAlert(alerts.single(), monitor, ERROR)
 
-            Assert.assertTrue(alerts.single().errorMessage?.contains("The destination address is invalid") as Boolean)
+            Assert.assertNotNull(alerts.single().errorMessage)
```
lezzago, line 864: Why is this check being modified? It doesn't seem like the error message should have changed.

lezzago: Ah ok that makes sense. Thanks.

qreshi: If this fails sometimes but your `assertNotNull` check doesn't it means that there is an error message but not the one we expect. That would mean the Alert is failing for another reason. Did you get a chance to see what the error was in these scenarios?

qreshi: @skkosuri-amzn By the way, I realized even as far back as our plugin version being `1.0.0`, the [Pull and Run Docker]() step in our Multi-node test workflow has been failing with `Error response from daemon: manifest for opensearchstaging/opensearch:<version> not found: manifest unknown: manifest unknown`. Is that a known issue?

Issue #203, opened 2021-10-13
[BUG] Cannot acknowledge more than 10 alerts at once.


AWSHurneyt:
**Describe the bug**
The dashboards on the `Alerts by trigger`, trigger-specific flyout, `Monitors` tab modal, and `Monitor Details` page do not support acknowledging more than 10 alerts at once. If more than 10 are selected, only 10 will be acknowledged. The same is true when issuing the command using `DevTools`.

**Other plugins installed**
N/A

**To Reproduce**
For the 3 dashboards and 1 modal mentioned above...
1. Create a monitor and trigger that will generate more than 10 alerts.
2. Go to the dashboard/modal you want to test.
3. Check off more than 10 alerts.
4. Press the `Acknowledge` button.
5. Go to the `Alerts by trigger` dashboard or `Monitors` tab to see the total number of acknowledged alerts.

**Expected behavior**
It seems reasonable to support acknowledging more than 10 alerts at once. If users should be restricted to acknowledging at most 10 alerts at once though, there should be some feedback explaining that limit.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: IOS
 - Browser: Chrome
 - Version: 10.15.7

**Additional context**

Issue #126, opened 2021-07-07
[BUG] exception instead of proper error


celesteking:
Getting an exception when trying to update a monitor with copy-pasted data from `GET _opendistro/_alerting/monitors/<monitor_id>`:

```json
PUT _opendistro/_alerting/monitors/NyRggXoBcuZ-vCEN3j30
{
  "_id" : "NyRggXoBcuZ-vCEN3j30",
  "_version" : 7,
  "_seq_no" : 23,
  "_primary_term" : 1,
  "monitor" : {
    "type" : "monitor",
    "schema_version" : 3,
ETC
```
Input request body doesn't conform to spec, but user shouldn't be confused by irrelevant error. Instead, a proper error should be emitted like "supplied data is incorrect, check docs" or whatever.

```
[2021-07-07T14:41:03,379][WARN ][r.suppressed             ] [elax1] path: /_opendistro/_alerting/monitors/NyRggXoBcuZ-vCEN3j30, params: {pretty=true, monitorID=NyRggXoBcuZ-vCEN3j30}
kotlin.UninitializedPropertyAccessException: lateinit property name has not been initialized
        at com.amazon.opendistroforelasticsearch.alerting.model.Monitor$Companion.parse(Monitor.kt:225) ~[?:?]
        at com.amazon.opendistroforelasticsearch.alerting.model.Monitor$Companion.parse$default(Monitor.kt:174) ~[?:?]
        at com.amazon.opendistroforelasticsearch.alerting.resthandler.RestIndexMonitorAction.prepareRequest(RestIndexMonitorAction.kt:76) ~[?:?]
        at org.elasticsearch.rest.BaseRestHandler.handleRequest(BaseRestHandler.java:94) ~[elasticsearch-7.10.2.jar:7.10.2]
        at com.amazon.opendistroforelasticsearch.security.filter.OpenDistroSecurityRestFilter$1.handleRequest(OpenDistroSecurityRestFilter.java:116) ~[?:?]
        at org.elasticsearch.rest.RestController.dispatchRequest(RestController.java:258) [elasticsearch-7.10.2.jar:7.10.2]
        at org.elasticsearch.rest.RestController.tryAllHandlers(RestController.java:340) [elasticsearch-7.10.2.jar:7.10.2]
        at org.elasticsearch.rest.RestController.dispatchRequest(RestController.java:191) [elasticsearch-7.10.2.jar:7.10.2]
[...]

PR #205, opened 2021-10-13
Fixed a bug that was preventing the AcknowledgeAlerts API from acknowledging more than 10 alerts at once.

AWSHurneyt:
*Issue #, if available:*

1. #203 
2. Alerting-Dashboard-Plugin [PR #126](https://github.com/opensearch-project/alerting-dashboards-plugin/issues/126)

*Description of changes:*
1. Fixed a bug that was preventing the AcknowledgeAlerts API from acknowledging more than 10 alerts at once. Because `TransportAcknowledgeAlertAction::findActiveAlerts` wasn't using the number of provided alerts for its `size`, the `search` API call was using the default `size` of `10`.

*CheckList:*

### DIFF
--- a/release-notes/opensearch-alerting.release-notes-1.1.0.0.md
+++ b/release-notes/opensearch-alerting.release-notes-1.1.0.0.md
@@ -26,8 +26,6 @@ Compatible with OpenSearch 1.1.0
 * Fix bug in paginating multiple bucket paths for Bucket-Level Monitor ([#163](https://github.com/opensearch-project/alerting/pull/163))
 * Various bug fixes for Bucket-Level Alerting ([#164](https://github.com/opensearch-project/alerting/pull/164))
 * Return only monitors for /monitors/_search ([#162](https://github.com/opensearch-project/alerting/pull/162))
-* Add valid search filters ([#191](https://github.com/opensearch-project/alerting/pull/191))
-* Fixed a bug that was preventing the AcknowledgeAlerts API from acknowledging more than 10 alerts at once ([#205](https://github.com/opensearch-project/alerting/pull/205/))
 
 ### Infrastructure
 


AWSHurneyt:
> Let's change the target branch to `main` to merge it in there first. This will also run the GitHub CI since we have it to only run against `main` (which we should also change later).

Will do.

alerting/src/test/kotlin/org/opensearch/alerting/resthandler/MonitorRestApiIT.kt
```diff
@@ -569,6 +569,129 @@ class MonitorRestApiIT : AlertingRestTestCase() {
         assertFalse("Alert in state ${activeAlert.state} found in failed list", failedResponseList.contains(activeAlert.id))
     }
 
+    fun `test acknowledging more than 10 alerts at once`() {
+        // GIVEN
+        putAlertMappings() // Required as we do not have a create alert API.
+        val monitor = createRandomMonitor(refresh = true)
+        val alertsToAcknowledge = arrayOf(
```
qreshi, line 576: To reduce redundancy for these you can refactor it to:
```
(1..15).map { createAlert(randomAlert(monitor).copy(state = Alert.State.ACTIVE))  }
```

alerting/src/test/kotlin/org/opensearch/alerting/resthandler/MonitorRestApiIT.kt
```diff
@@ -569,6 +569,129 @@ class MonitorRestApiIT : AlertingRestTestCase() {
...
+        val acknowledgedAlerts = responseMap["success"] as List<String>
+        assertTrue("Expected $expectedAcknowledgedCount alerts to be acknowledged successfully.", acknowledgedAlerts.size == expectedAcknowledgedCount)
+
+        val acknowledgedAlertsList = acknowledgedAlerts.toString()
+        alertsToAcknowledge.forEach { alert -> assertTrue("Alert with ID ${alert.id} not found in failed list.", acknowledgedAlertsList.contains(alert.id)) }
+
+        val failedResponse = responseMap["failed"] as List<String>
+        assertTrue("Expected 0 alerts to fail acknowledgment.", failedResponse.isEmpty())
+
+        val failedResponseList = failedResponse.toString()
```
qreshi, line 610: Since we're asserting above that there are no alerts in the `failed` response, I think we can remove this check since the test would have failed prior to reaching here in the case where there is something to check.

release-notes/opensearch-alerting.release-notes-1.1.0.0.md
```diff
@@ -26,6 +26,8 @@ Compatible with OpenSearch 1.1.0
 * Fix bug in paginating multiple bucket paths for Bucket-Level Monitor ([#163](https://github.com/opensearch-project/alerting/pull/163))
 * Various bug fixes for Bucket-Level Alerting ([#164](https://github.com/opensearch-project/alerting/pull/164))
 * Return only monitors for /monitors/_search ([#162](https://github.com/opensearch-project/alerting/pull/162))
+* Add valid search filters ([#191](https://github.com/opensearch-project/alerting/pull/191))
+* Fixed a bug that was preventing the AcknowledgeAlerts API from acknowledging more than 10 alerts at once ([#205](https://github.com/opensearch-project/alerting/pull/205/))
```
qreshi, line 30: You might actually want to hold off on updating any `1.1` release notes here.

I think since we're adding patches to the `1.1` branch after the OS 1.1 release, we'll want to update the `1.1` version from `1.1.0.0` to `1.1.1.0` to reflect the patches. In the future if there is a OS 1.1.1 release, then our patched version of Alerting 1.1 would go in there. So it'd actually be going into another release note I guess since we keep them down to the build version (these things wouldn't be done in `main`, we'd do them on the `1.1` branch itself).

This does bring up a question though on our process here. We should increment the patch version if we end up pushing changes to a minor version branch after an OpenSearch release but do we then create release notes whenever we update our patch version or only if an updated version of the plugin is going into an OpenSearch release?

@skkosuri-amzn @lezzago Do you guys agree with the process above and any comments regarding the version update/release notes?

AWSHurneyt: That's a good point. I can definitely remove the notes or move them to a new note depending on what's decided. To confirm, we're referring to the 191 and 205 entries, correct?

AWSHurneyt: Removed those 2 entries for now while we discuss adding a separate note.

Issue #89, opened 2021-06-02
[BUG] Monitor jobs do not reschedule when flipping the SWEEPER_ENABLED setting


aditjind:
<a href="https://github.com/qreshi"><img src="https://avatars.githubusercontent.com/u/47198598?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [qreshi](https://github.com/qreshi)**
_Thursday Jan 07, 2021 at 02:23 GMT_
_Originally opened as https://github.com/opendistro-for-elasticsearch/alerting/issues/333_

----

When the `SWEEPER_ENABLED` setting is set to `false`, all Alerting jobs are descheduled as expected. However, once this setting is set back to `true`, the expectation is that existing enabled Monitors would be rescheduled but this is currently not the case.

The issue seems to be occurring in the `sweep()` logic. This is indirectly invoked from the `enable()` method when the `SWEEPER_ENABLED` setting is set to `true`. At this time, any jobs owned by the shard in question are skipped if the `newVersion` being passed in is less than or equal to the `currentVersion`. So when the sweeper is re-enabled, all the jobs that went through no change are skipped over and never rescheduled.

Here is a snippet of the code in question:
https://github.com/opendistro-for-elasticsearch/alerting/blob/8daf9e64ccfd4219441002a821a2bb48bbeeaaf9/core/src/main/kotlin/com/amazon/opendistroforelasticsearch/alerting/core/JobSweeper.kt#L348-L361

The solution would involve accounting for the case where the job version is unchanged only when coming from a scenario where the jobs are previously descheduled and coming from the `enable()` code path, so that the logic where the jobs are scheduled can occur. Not accounting for the aforementioned scenario could lead to existing jobs being repeatedly descheduled and rescheduled during the routine background runs, which should be avoided.

aditjind:
<a href="https://github.com/pkriete"><img src="https://avatars.githubusercontent.com/u/32901?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [pkriete](https://github.com/pkriete)**
_Thursday Feb 11, 2021 at 00:00 GMT_

----

We would love to see this fixed soon. We ran into this issue on AWS ES after toggling `opendistro.scheduled_jobs.enabled` to `false` during an outage. Setting it back to `true` results in a completely opaque failure mode where everything looks like it's running but nothing happens. We have customer facing alerts, the ability to turn them off when they would otherwise be incorrect is critical for us.

[...]

PR #243, opened 2021-11-18
Update sweep logic to re-schedule unchanged jobs when SWEEPER_ENABLED is toggled

qreshi:

*Issue #, if available:* #89

*Description of changes:*
* Fixed the case where if an enabled job is de-scheduled when `SWEEPER_ENABLED` is set to `false`, jobs that were unchanged were not re-scheduled when `SWEEPER_ENABLED` was set to back to `true`
* Added a manual sweep of all shards when `initBackgroundSweep()` is called since the runnable being scheduled in the threadpool is only executed for the first time after the `SWEEPER_PERIOD` (which defaults to 5 minutes) has passed and this could lead to delays in jobs that should run more frequently than that
* Added a test to capture the changes
* Confirmed with an anti-test that the written test failed as expected without the fix

*CheckList:*

### DIFF
--- a/alerting/src/test/kotlin/org/opensearch/alerting/resthandler/MonitorRestApiIT.kt
+++ b/alerting/src/test/kotlin/org/opensearch/alerting/resthandler/MonitorRestApiIT.kt
@@ -1090,17 +1090,15 @@ class MonitorRestApiIT : AlertingRestTestCase() {
     }
 
     private fun isMonitorScheduled(monitorId: String, alertingStatsResponse: Map<String, Any>): Boolean {
-        var monitorScheduled = false
         val nodesInfo = alertingStatsResponse["nodes"] as Map<String, Any>
         for (nodeId in nodesInfo.keys) {
             val nodeInfo = nodesInfo[nodeId] as Map<String, Any>
             val jobsInfo = nodeInfo["jobs_info"] as Map<String, Any>
             if (jobsInfo.keys.contains(monitorId)) {
-                monitorScheduled = true
-                break
+                return true
             }
         }
 
-        return monitorScheduled
+        return false
     }
 }
--- a/alerting/src/test/kotlin/org/opensearch/alerting/resthandler/MonitorRestApiIT.kt
+++ b/alerting/src/test/kotlin/org/opensearch/alerting/resthandler/MonitorRestApiIT.kt
@@ -271,7 +271,7 @@ class MonitorRestApiIT : AlertingRestTestCase() {
         }
     }
 
-    /* Enable this test case after issue issue#269 is fixed.
+    /* Enable this test case after checking for disallowed destination during Monitor creation is added in
     fun `test creating a monitor with a disallowed destination type fails`() {
         try {
             // Create a Chime Destination
@@ -284,7 +284,8 @@ class MonitorRestApiIT : AlertingRestTestCase() {
                 chime = chime,
                 slack = null,
                 customWebhook = null,
-                email = null)
+                email = null
+            )
             val chimeDestination = createDestination(destination = destination)
 
             // Remove Chime from the allow_list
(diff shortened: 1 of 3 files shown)


alerting/src/test/kotlin/org/opensearch/alerting/resthandler/MonitorRestApiIT.kt
```diff
@@ -1045,4 +1082,25 @@ class MonitorRestApiIT : AlertingRestTestCase() {
...
+
+    private fun isMonitorScheduled(monitorId: String, alertingStatsResponse: Map<String, Any>): Boolean {
+        var monitorScheduled = false
+        val nodesInfo = alertingStatsResponse["nodes"] as Map<String, Any>
+        for (nodeId in nodesInfo.keys) {
+            val nodeInfo = nodesInfo[nodeId] as Map<String, Any>
+            val jobsInfo = nodeInfo["jobs_info"] as Map<String, Any>
+            if (jobsInfo.keys.contains(monitorId)) {
+                monitorScheduled = true
+                break
```
lezzago, line 1100: Why not just return true here and on the bottom of the function, have it return false?

alerting/src/test/kotlin/org/opensearch/alerting/resthandler/MonitorRestApiIT.kt
```diff
@@ -887,23 +889,62 @@ class MonitorRestApiIT : AlertingRestTestCase() {
...
+    }
+
+    fun `test monitor stats when disabling and re-enabling scheduled jobs with existing monitor`() {
+        // Enable Monitor jobs
+        enableScheduledJob()
+        val monitorId = createMonitor(randomQueryLevelMonitor(enabled = true), refresh = true).id
+
+        var alertingStats = getAlertingStats()
+        assertEquals("Scheduled job is not enabled", true, alertingStats[ScheduledJobSettings.SWEEPER_ENABLED.key])
+        assertEquals("Scheduled job index does not exist", true, alertingStats["scheduled_job_index_exists"])
```
lezzago, line 902: Use `assertTrue` instead

alerting/src/test/kotlin/org/opensearch/alerting/resthandler/MonitorRestApiIT.kt
```diff
@@ -887,23 +889,62 @@ class MonitorRestApiIT : AlertingRestTestCase() {
...
+        assertTrue(
+            "Monitor [$monitorId] was not found scheduled based on the alerting stats response: $alertingStats",
+            isMonitorScheduled(monitorId, alertingStats)
+        )
+
+        // Disable Monitor jobs
+        disableScheduledJob()
+
+        alertingStats = getAlertingStats()
+        assertEquals("Scheduled job is still enabled", false, alertingStats[ScheduledJobSettings.SWEEPER_ENABLED.key])
```
lezzago, line 918: use `assertFalse` instead

alerting/src/test/kotlin/org/opensearch/alerting/resthandler/MonitorRestApiIT.kt
```diff
@@ -887,23 +889,62 @@ class MonitorRestApiIT : AlertingRestTestCase() {
...
+        )
+
+        // Re-enable Monitor jobs
+        enableScheduledJob()
+
+        // Sleep briefly so sweep can reschedule the Monitor
+        Thread.sleep(2000)
+
+        alertingStats = getAlertingStats()
+        assertEquals("Scheduled job is not enabled", true, alertingStats[ScheduledJobSettings.SWEEPER_ENABLED.key])
```
lezzago, line 931: use `assertTrue` instead

alerting/src/test/kotlin/org/opensearch/alerting/resthandler/MonitorRestApiIT.kt
qreshi: Leaving this as `assertEquals` because `assertTrue` doesn't infer type and `alertStats` is a Map<String, Any> so all the actual items being passed in would need to have `as Boolean` added to them (it complains otherwise) which would get a little bothersome. I think that's why the other tests wrote it that way as well.

alerting/src/test/kotlin/org/opensearch/alerting/resthandler/MonitorRestApiIT.kt
qreshi: Good point, I think there was an assertion in the method originally which is why it was saved as a variable.

zengyan-amazon (approved):

skkosuri-amzn (approved):

AWSHurneyt (approved):

Issue #172, opened 2021-09-09
Add setting to limit max Actions per Trigger


qreshi:
**Is your feature request related to a problem? Please describe.**
The number of Actions that can be used per Trigger is currently unbounded. It has been observed in the past that for Query-Level Monitors, very large Notification Action counts can lead to compile errors for Painless scripts (this needs to be investigated more). With the introduction of Bucket-Level Monitors as well, it could be useful to limit the Action count to prevent Action executions exploding.

**Describe the solution you'd like**
Introduce some user-configurable dynamic setting to limit the maximum number of Actions per Trigger. Since Alerting already exists with Actions being unbounded, if this setting is introduced, the default will likely need to be `-1` or something similar to be unbounded as well so that existing Monitors don't start failing by going over a newly introduced threshold.

**Describe alternatives you've considered**

**Additional context**

PR #275, opened 2022-01-12
Limitactions1.2.3

charliezhangaws:
*Issue #, if available:*
https://github.com/opensearch-project/alerting/issues/172
*Description of changes:*
added a user-configurable limit in Query Level.

I have retrieved the limit settings in both configuration files (AlertingSetting.kt and LegacyOpenDistroAlertingSettings.kt)
LegacyOpenDistroAlertingSettings.kt line 29-33
AlertingSetting.kt line 28-32

imported limit settings into MonitorRunner.kt Line 45

set an attribute called threshold in the MonitorRunner to iterate through the list of triggers and check the length of the list of actions. Line 295

added a line of conditional control to MonitorRunner.kt to ensure there is a protection mechanism. Line 296

*CheckList:*


.github/workflows/multi-node-test-workflow.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         with:
           java-version: 14
       - name: Run integration tests with multi node config
-        run: ./gradlew integTest -PnumNodes=3 -Dopensearch.version=1.3.0-SNAPSHOT
```
qreshi, line 35: `main` is currently on version `1.3.0` so you'll want to develop the changes against that since this will be merged into `main`

alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
```diff
@@ -25,6 +25,12 @@ class AlertingSettings {
             Setting.Property.NodeScope, Setting.Property.Dynamic
         )
 
+        val ALERTING_TRIGGER_MAX_ACTIONS = Setting.intSetting(
+            "plugins.alerting.trigger.max_actions",
+            LegacyOpenDistroAlertingSettings.ALERTING_TRIGGER_MAX_ACTIONS,
```
qreshi, line 30: Any new settings can have their default constants defined here in `AlertingSettings` instead of legacy since that was for existing settings when we migrated to OpenSearch. You can refer to `MAX_ACTIONABLE_ALERT_COUNT` below as an example.

alerting/src/main/kotlin/org/opensearch/alerting/settings/LegacyOpenDistroAlertingSettings.kt
```diff
@@ -26,6 +26,12 @@ class LegacyOpenDistroAlertingSettings {
             Setting.Property.NodeScope, Setting.Property.Dynamic, Setting.Property.Deprecated
         )
 
+        val ALERTING_TRIGGER_MAX_ACTIONS = Setting.intSetting(
```
qreshi, line 29: This can be removed since the setting is only being introduced in an OpenSearch version.

alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunner.kt
```diff
@@ -291,7 +292,11 @@ object MonitorRunner : JobRunner, CoroutineScope, AbstractLifecycleComponent() {
 
             if (triggerService.isQueryLevelTriggerActionable(triggerCtx, triggerResult)) {
                 val actionCtx = triggerCtx.copy(error = monitorResult.error ?: triggerResult.error)
-                for (action in trigger.actions) {
+                var threshold = ALERTING_TRIGGER_MAX_ACTIONS.get(settings)
+                threshold = if (threshold < trigger.actions.size) threshold else trigger.actions.size
+
+                // val threshold = if(ALERTING_TRIGGER_MAX_ACTIONS.get(settings) < trigger.actions.size) ALERTING_TRIGGER_MAX_ACTIONS.get(settings) else trigger.actions.size
+                for (action in trigger.actions.slice(0 until threshold)) {
```
qreshi, line 299: Since the default value for `ALERTING_TRIGGER_MAX_ACTIONS` is `-1` won't this slice just return nothing meaning no actions will be evaluated?

I'm assuming the default behavior we want is to evaluate all actions when the setting is `-1` so this logic will need to account for that.

You'll also want to add the logic to implement `ALERTING_TRIGGER_MAX_ACTIONS` for Bucket-Level Monitors.

Also, let's add some tests to cover the behavior of this new setting. If you want to add integration tests for Monitor execution behavior, you can write it in [MonitorRunnerIT](https://github.com/opensearch-project/alerting/blob/main/alerting/src/test/kotlin/org/opensearch/alerting/MonitorRunnerIT.kt)

alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunner.kt
```diff
@@ -291,7 +292,11 @@ object MonitorRunner : JobRunner, CoroutineScope, AbstractLifecycleComponent() {
 
             if (triggerService.isQueryLevelTriggerActionable(triggerCtx, triggerResult)) {
                 val actionCtx = triggerCtx.copy(error = monitorResult.error ?: triggerResult.error)
-                for (action in trigger.actions) {
+                var threshold = ALERTING_TRIGGER_MAX_ACTIONS.get(settings)
+                threshold = if (threshold < trigger.actions.size) threshold else trigger.actions.size
+
+                // val threshold = if(ALERTING_TRIGGER_MAX_ACTIONS.get(settings) < trigger.actions.size) ALERTING_TRIGGER_MAX_ACTIONS.get(settings) else trigger.actions.size
```
qreshi, line 298: Can remove commented code

charliezhangaws: Got it!

charliezhangaws: > Can remove commented code

Got it!

alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunner.kt
charliezhangaws: > Since the default value for `ALERTING_TRIGGER_MAX_ACTIONS` is `-1` won't this slice just return nothing meaning no actions will be evaluated?
> 
> I'm assuming the default behavior we want is to evaluate all actions when the setting is `-1` so this logic will need to account for that.
> Got it! I added `&& threshold >= 0` right after `threshold < trigger.actions.size`. how is that sounds like or i will still have some logical problem?
> You'll also want to add the logic to implement `ALERTING_TRIGGER_MAX_ACTIONS` for Bucket-Level Monitors.
> Got it! So basically, what I'm going to do is add the same thing like what i did in the Query Level right?
> Also, let's add some tests to cover the behavior of this new setting. If you want to add integration tests for Monitor execution behavior, you can write it in [MonitorRunnerIT](https://github.com/opensearch-project/alerting/blob/main/alerting/src/test/kotlin/org/opensearch/alerting/MonitorRunnerIT.kt)
I tested it in the OpensearchDashboard. Do i still need to add integration tests for Monitor execution behavior?

alerting/src/main/kotlin/org/opensearch/alerting/settings/LegacyOpenDistroAlertingSettings.kt
charliezhangaws: > This can be removed since the setting is only being introduced in an OpenSearch version.

Got it!

alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
charliezhangaws: > Any new settings can have their default constants defined here in `AlertingSettings` instead of legacy since that was for existing settings when we migrated to OpenSearch. You can refer to `MAX_ACTIONABLE_ALERT_COUNT` below as an example.

Got it! So the `legacy` was something like old version before migated to Opensearch? Can i ask where did it get migrated here? is that from Elastic Search?

alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunner.kt
```diff
@@ -291,7 +300,8 @@ object MonitorRunner : JobRunner, CoroutineScope, AbstractLifecycleComponent() {
 
             if (triggerService.isQueryLevelTriggerActionable(triggerCtx, triggerResult)) {
                 val actionCtx = triggerCtx.copy(error = monitorResult.error ?: triggerResult.error)
-                for (action in trigger.actions) {
+                var threshold = if (triggerMaxActions < trigger.actions.size && triggerMaxActions >= 0) triggerMaxActions else trigger.actions.size
```
lezzago, line 303: So we want to guarantee all the actions execute even if the number of actions are more than the max actions setting?

alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
```diff
@@ -25,6 +25,12 @@ class AlertingSettings {
             Setting.Property.NodeScope, Setting.Property.Dynamic
         )
 
+        val ALERTING_TRIGGER_MAX_ACTIONS = Setting.intSetting(
+            "plugins.alerting.trigger.max_actions",
+            -1,
+            Setting.Property.NodeScope, Setting.Property.Dynamic //Setting.Validator<List<T>> validator
```
lezzago, line 31: Remove commented code

build.gradle
```diff
@@ -62,7 +62,7 @@ task ktlintFormat(type: JavaExec, group: "formatting") {
     args "-F", "alerting/**/*.kt", "elastic-api/**/*.kt", "core/**/*.kt"
 }
 
-check.dependsOn ktlint
+//check.dependsOn ktlint
```
lezzago, line 65: undo the comment

build.gradle
```diff
@@ -82,7 +82,7 @@ allprojects {
     }
     plugins.withId('org.jetbrains.kotlin.jvm') {
         compileKotlin.kotlinOptions.jvmTarget = compileTestKotlin.kotlinOptions.jvmTarget = "1.8"
-        compileKotlin.dependsOn ktlint
+        //compileKotlin.dependsOn ktlint
```
lezzago, line 85: undo the comment

lezzago:
Closing PR in favor of #434

PR #283, opened 2022-01-21
Add back opendistro prefixed sweeper enabled field to alerting stats response

qreshi:

*Issue #, if available:*

*Description of changes:*
The `opendistro` prefixed field response for sweeper/scheduled jobs enabled in the [Alerting Stats API](https://opensearch.org/docs/latest/monitoring-plugins/alerting/api/#monitor-stats) was inadvertently changed to `plugins` during the migration to OpenSearch. This change reintroduces the legacy field back alongside the existing `plugins` one to maintain backwards compatibility and to not break behavior for anyone using the current field.

This PR also updates the existing tests asserting on the response of the stats API to use hardcoded values for the scheduled jobs enabled field instead of referring to the `Setting.key` so it can catch a change like this in the future. We should also see if any other tests are currently checking API responses this way. Fixing those, however, is out of scope for this particular PR.

*CheckList:*

### DIFF
--- a/alerting/src/test/kotlin/org/opensearch/alerting/AlertingRestTestCase.kt
+++ b/alerting/src/test/kotlin/org/opensearch/alerting/AlertingRestTestCase.kt
@@ -69,8 +69,8 @@ abstract class AlertingRestTestCase : ODFERestTestCase() {
     protected val numberOfNodes = System.getProperty("cluster.number_of_nodes", "1")!!.toInt()
     protected val isMultiNode = numberOfNodes > 1
 
-    protected val statsResponseOpendistroSweeperEnabledSetting = "opendistro.scheduled_jobs.enabled"
-    protected val statsResponseOpenSearchSweeperEnabledSetting = "plugins.scheduled_jobs.enabled"
+    protected val statsResponseOpendistroSweeperEnabledField = "opendistro.scheduled_jobs.enabled"
+    protected val statsResponseOpenSearchSweeperEnabledField = "plugins.scheduled_jobs.enabled"
 
     override fun xContentRegistry(): NamedXContentRegistry {
         return NamedXContentRegistry(
--- a/alerting/src/test/kotlin/org/opensearch/alerting/AlertingRestTestCase.kt
+++ b/alerting/src/test/kotlin/org/opensearch/alerting/AlertingRestTestCase.kt
@@ -69,6 +69,9 @@ abstract class AlertingRestTestCase : ODFERestTestCase() {
     protected val numberOfNodes = System.getProperty("cluster.number_of_nodes", "1")!!.toInt()
     protected val isMultiNode = numberOfNodes > 1
 
+    protected val statsResponseOpendistroSweeperEnabledSetting = "opendistro.scheduled_jobs.enabled"
+    protected val statsResponseOpenSearchSweeperEnabledSetting = "plugins.scheduled_jobs.enabled"
+
     override fun xContentRegistry(): NamedXContentRegistry {
         return NamedXContentRegistry(
             mutableListOf(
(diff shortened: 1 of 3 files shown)


alerting/src/test/kotlin/org/opensearch/alerting/AlertingRestTestCase.kt
```diff
@@ -69,6 +69,9 @@ abstract class AlertingRestTestCase : ODFERestTestCase() {
     protected val numberOfNodes = System.getProperty("cluster.number_of_nodes", "1")!!.toInt()
     protected val isMultiNode = numberOfNodes > 1
 
+    protected val statsResponseOpendistroSweeperEnabledSetting = "opendistro.scheduled_jobs.enabled"
+    protected val statsResponseOpenSearchSweeperEnabledSetting = "plugins.scheduled_jobs.enabled"
```
getsaurabh02, line 73: nit: should we rather call it ``Field`` instead of ``Setting``, given this is part of api response structure.

alerting/src/test/kotlin/org/opensearch/alerting/resthandler/MonitorRestApiIT.kt
```diff
@@ -886,8 +886,7 @@ class MonitorRestApiIT : AlertingRestTestCase() {
         disableScheduledJob()
 
         val responseMap = getAlertingStats()
-        // assertEquals("Cluster name is incorrect", responseMap["cluster_name"], "alerting_integTestCluster")
-        assertEquals("Scheduled job is not enabled", false, responseMap[ScheduledJobSettings.SWEEPER_ENABLED.key])
+        assertAlertingStatsSweeperEnabled(responseMap, false)
```
getsaurabh02, line 889: Can we add a Resource File with the Rest API response, and assert on those values instead. You may want to check how it is achieved similarly in Rest ITs of the core. I feel it's more manageable that way and you will not have to define the explicit fields in the test here.

getsaurabh02 (commented):
LGTM however you may want to consider the suggestion around Rest API ITs.

alerting/src/test/kotlin/org/opensearch/alerting/resthandler/MonitorRestApiIT.kt
qreshi: Personally, I think asserting here is cleaner since for some tests we just want to check specific values and not the entire response. If we had a resource file then we'd need to handcraft the entire responses for new cases that aren't covered by the old ones. That would be more appropriate for unit tests where we're mocking but would add overhead here.

alerting/src/test/kotlin/org/opensearch/alerting/AlertingRestTestCase.kt
qreshi: I guess `statsResponseOpenSearchSweeperEnabledSettingField` is the **most** accurate since it is just the setting that we're outputting here but it's quite verbose. Do you have a preference between that and just `Field`?

getsaurabh02: Either way it looks good. Only keep it less verbose I might just keep ``Field``

alerting/src/test/kotlin/org/opensearch/alerting/resthandler/MonitorRestApiIT.kt
getsaurabh02: You can also have resource file versioned since any new API changes would only happen with a change in the minor version. Also, the validation would be between the objects in resource file and the API response here, not necessarily requiring any mock. In order to check for any specific values, we can always filter the object in consideration, given it will be a json structure. 

Just that having a resource file would defines one source of truth to compare the API responses against. However, it is just an open idea, and might not necessarily be included in the scope of this PR. Feel free to skip it for now.

qreshi: Fair point. Let's skip it for this PR but can you create an issue capturing this and we can continue discussions there and have others weigh in on the option.

lezzago (approved):

rishabhmaurya (approved):

Issue #241, opened 2021-11-15
[BUG] Flakey tests in github workflows related to :alerting:alertingBwcCluster#twoThirdsUpgradedClusterTask


leeyun-amzn:
**Describe the bug**
The task `:alerting:alertingBwcCluster#twoThirdsUpgradedClusterTask` is flakey when executing the test workflow and multi-node test workflow. It sometimes fails, and sometimes succeeds on re-run.

**Other plugins installed**
None

**To Reproduce**
Running the test workflow and multi-node workflow

**Expected behavior**
Should not have flake behavior.

**Screenshots**
N/A

**Desktop (please complete the following information):**
Using Github action in this repository.

**Additional context**
Log files:
For multi-node test workflow [logs_multi_node_bwc_flakey.zip](https://github.com/opensearch-project/alerting/files/7541970/logs_multi_node_bwc_flakey.zip)
 
For test workflow [logs_test_workflow_bwc_flakey.zip](https://github.com/opensearch-project/alerting/files/7541967/logs_test_workflow_bwc_flakey.zip)

qreshi:
Closing as this has been resolved.

PR #296, opened 2022-01-31
Fix flaky BWC test

qreshi:

*Issue #, if available:* #241

*Description of changes:*
The BWC tests will occasionally fail on the `twoThirdsUpgradedClusterTask`. It seems that there is some race condition that can lead to some test cluster changes to not go through due to the sleep being done in the mixed cluster scenario. For the time being, the sleep will be removed from `MIXED_CLUSTER` scenario as this was shown to remove flakiness.

The sleeps were initially put in to give time for a Monitor execution in between assertions but the current checks in mixed state will suffice. Once we have an easy way to artificially forward Monitor executions (which will likely be after integration with Job Scheduler since the next execution time isn't stored in the Monitor job currently) then we can add additional checks without relying on sleeps.

**Some local verification**:
### Multiple test runs before change
```
(22-01-31 17:22:48) <0> [~/workspace/opensearch-alerting]
qreshi % for i in {1..10}; do ./gradlew bwcTestSuite &> /dev/null || echo "Run $i failed"; done
Run 1 failed
Run 4 failed
Run 6 failed
Run 7 failed
Run 9 failed
```

### Multiple test runs after removal of sleep
```
(22-01-29 1:06:34) <0> [~/workspace/opensearch-alerting]
qreshi % for i in {1..10}; do ./gradlew bwcTestSuite &> /dev/null || echo "Run $i failed"; done
```

*CheckList:*

### DIFF
--- a/alerting/src/test/kotlin/org/opensearch/alerting/bwc/AlertingBackwardsCompatibilityIT.kt
+++ b/alerting/src/test/kotlin/org/opensearch/alerting/bwc/AlertingBackwardsCompatibilityIT.kt
@@ -59,11 +59,6 @@ class AlertingBackwardsCompatibilityIT : AlertingRestTestCase() {
                 ClusterType.MIXED -> {
                     assertTrue(pluginNames.contains("opensearch-alerting"))
                     verifyMonitorExists(LEGACY_OPENDISTRO_ALERTING_BASE_URI)
-                    // Waiting a minute to ensure the Monitor ran again at least once before checking if the job is running
-                    // on time
-                    // TODO: Should probably change the next execution time of the Monitor manually instead since this inflates
-                    //  the test execution by a lot
-                    Thread.sleep(60000)
                     // TODO: Need to move the base URI being used here into a constant and rename ALERTING_BASE_URI to
                     //  MONITOR_BASE_URI
                     verifyMonitorStats("/_opendistro/_alerting")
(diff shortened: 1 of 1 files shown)


alerting/src/test/kotlin/org/opensearch/alerting/bwc/AlertingBackwardsCompatibilityIT.kt
```diff
@@ -59,18 +59,17 @@ class AlertingBackwardsCompatibilityIT : AlertingRestTestCase() {
...
                     verifyMonitorStats("/_opendistro/_alerting")
                 }
                 ClusterType.UPGRADED -> {
                     assertTrue(pluginNames.contains("opensearch-alerting"))
                     verifyMonitorExists(ALERTING_BASE_URI)
+                    // TODO: Change the next execution time of the Monitor manually instead since this inflates
+                    //  the test execution by a lot (might have to wait for Job Scheduler plugin integration first)
+                    // Waiting a minute to ensure the Monitor ran again at least once before checking if the job is running
+                    // on time
                     Thread.sleep(60000)
```
getsaurabh02, line 73: Usually a consistent way to avoid sleeps is by using ``assertBusy``, given if we expect some change of state within a stipulated time threshold. It allows to runs the code block for the provided interval, waiting for no assertions to trip, and return as soon as all the assertions are met. So wondering of there exists some state related to monitor execution, such as count, that we can rely upon as a trigger here, and then verify the stats response.

qreshi: The reason for the sleep was that the minimum Monitor execution interval is `1 minute` and we were going to ensure the Monitor executed again before checking the stats to make sure that any late running Monitor manifested itself. The test itself is not perfect and we'd normally want to just fast forward the job execution like we do for the Index Management test but Alerting uses an older job scheduling implementation where the next job execution time is not stored on the document but rather in some `ConcurrentHashMap` object that is not test friendly.

getsaurabh02: Thanks for clarifying @qreshi. So my suggestion was, is there is a way (API/Object) which can specify if monitor was run again (such as a running-counter). If yes, then we can do [assertBusy](https://github.com/opensearch-project/OpenSearch/blob/main/test/framework/src/main/java/org/opensearch/test/OpenSearchTestCase.java#L1053-L1055), which provides an exponential backoff mechanism to wait until the condition is met.  
If not, I understand having sleep is the only valid option.

getsaurabh02 (approved):

lezzago (approved):

Issue #355, opened 2022-03-21
[BUG] Security Integration tests are failing for 2.0


lezzago:
**Describe the bug**
The security Integration tests are failing for 2.0. This is due to the [security plugin](https://github.com/opensearch-project/security/) not being upgraded to 2.0 yet, so the cluster with security for 2.0 cannot come up and causes the security integration tests to fail.

**Other plugins installed**
Security plugin

**Expected behavior**
Once the security plugin has upgraded to 2.0, the tests should pass. If not, more investigation will then be required.

getsaurabh02:
Related to https://github.com/opensearch-project/alerting/pull/354

sbcd90:
Issue resolved by PR #407 . Closing this issue.

Issue #351, opened 2022-03-14
[2.0.0][Distribution][BUG] Failed to apply plugin class 'org.opensearch.gradle.info.GlobalBuildInfoPlugin'.


peterzhuamazon:
```
2022-03-14 18:54:21 INFO     Sanity testing alerting
2022-03-14 18:54:21 INFO     Executing "git init" in /tmp/tmplk9kqyh6/alerting
2022-03-14 18:54:21 INFO     Executing "git remote add origin https://github.com/opensearch-project/alerting.git" in /tmp/tmplk9kqyh6/alerting
2022-03-14 18:54:21 INFO     Executing "git fetch --depth 1 origin main" in /tmp/tmplk9kqyh6/alerting
2022-03-14 18:54:21 INFO     Executing "git checkout FETCH_HEAD" in /tmp/tmplk9kqyh6/alerting
2022-03-14 18:54:21 INFO     Executing "git rev-parse HEAD" in /tmp/tmplk9kqyh6/alerting
2022-03-14 18:54:21 INFO     Checked out https://github.com/opensearch-project/alerting.git@main into /tmp/tmplk9kqyh6/alerting at 9487c00a388dbc3d9a5bce602a2b70b0ec20f90f
2022-03-14 18:54:21 INFO     Executing "./gradlew properties -Dopensearch.version=2.0.0-SNAPSHOT -Dbuild.snapshot=true" in /tmp/tmplk9kqyh6/alerting

FAILURE: Build failed with an exception.

* Where:
Build file '/tmp/tmplk9kqyh6/alerting/alerting/build.gradle' line: 13

* What went wrong:
A problem occurred evaluating project ':alerting'.
> Failed to apply plugin class 'org.opensearch.gradle.info.GlobalBuildInfoPlugin'.
   > Could not create plugin of type 'GlobalBuildInfoPlugin'.
      > Could not generate a decorated class for type GlobalBuildInfoPlugin.
         > org/gradle/internal/jvm/inspection/JvmInstallationMetadata

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org/

BUILD FAILED in 27s
2022-03-14 18:54:52 ERROR    CI Manifest check failed
Traceback (most recent call last):
  File "./src/run_ci.py", line 25, in <module>
    sys.exit(main())
  File "./src/run_ci.py", line 21, in main
[...]

PR #357, opened 2022-03-21
Upgrade Alerting to 2.0

lezzago:
*Issue #, if available:*
#313 - OpenSearch 2.0 Upgrade
#325 - Remove JDK 8 support
#351 - Fix issue by upgrading to 2.0
*Description of changes:*
- Build Alerting for OS 2.0
- Security integ test fails due to this [issue](https://github.com/opensearch-project/alerting/issues/355)

*CheckList:*


core/build.gradle
```diff
@@ -9,14 +9,14 @@ apply plugin: 'jacoco'
...
     compileOnly "org.opensearch:opensearch:${opensearch_version}"
-    compile "org.jetbrains.kotlin:kotlin-stdlib:${kotlin_version}"
-    compile 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.1.1'
-    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:${kotlin_version}"
-    compile "com.cronutils:cron-utils:9.1.6"
-    compile "org.opensearch.client:opensearch-rest-client:${opensearch_version}"
-    compile 'com.google.googlejavaformat:google-java-format:1.10.0'
-    compile "org.opensearch:common-utils:${common_utils_version}"
-    compile 'commons-validator:commons-validator:1.7'
+    api "org.jetbrains.kotlin:kotlin-stdlib:${kotlin_version}"
```
rishabhmaurya, line 12: why `api` and not `implementation` configuration? Are we exporting all of these libraries?

alerting/build.gradle
```diff
@@ -50,24 +50,24 @@ configurations.all {
...
 }
 
 dependencies {
     compileOnly "org.opensearch.plugin:opensearch-scripting-painless-spi:${versions.opensearch}"
 
     // OpenSearch Nanny state
-    compile "org.jetbrains.kotlin:kotlin-stdlib:${kotlin_version}"
-    compile "org.jetbrains.kotlin:kotlin-stdlib-common:${kotlin_version}"
-    compile "org.jetbrains:annotations:13.0"
+    api "org.jetbrains.kotlin:kotlin-stdlib:${kotlin_version}"
```
rishabhmaurya, line 61: why not `implementation`?

core/build.gradle
lezzago: At first I was trying to get the build working with gradle 7, but didnt realize that it would work if some of these can be converted to `implementation`. I converted as many of the dependencies from `api` to `implementation`

rishabhmaurya: I think alerting should just build fine. Now it will not export these dependencies transitively and thus will not be available to any other module, dependent on alerting, at compile time. But looking at these dependencies, I think that most of them, if not all, will not be required by other modules as transitive dependencies at compile time.

rishabhmaurya (approved):

PR #363, opened 2022-03-25
Seggregate monitor runner logic for separation of concerns

getsaurabh02:
*Description of changes:*

Segregate the monitor runner logic based on monitor execution logic for separation of concerns and improved testability. Also provide applicability of dry run for direct executions from the transport actions.

*CheckList:*
Testing the changes currently and will update once done. 
Few minor semantic refactoring to be followed in the same PR with fresh commits (including ktlint)


alerting/src/main/kotlin/org/opensearch/alerting/BucketLevelMonitor.kt
```diff
@@ -0,0 +1,350 @@
...
+import org.opensearch.alerting.script.BucketLevelTriggerExecutionContext
+import org.opensearch.alerting.script.TriggerExecutionContext
+import org.opensearch.alerting.util.getActionExecutionPolicy
+import org.opensearch.alerting.util.getBucketKeysHash
+import org.opensearch.alerting.util.getCombinedTriggerRunResult
+import org.opensearch.alerting.util.isAllowed
+import org.opensearch.common.Strings
+import java.time.Instant
+
+class BucketLevelMonitor: MonitorRunner {
```
qreshi, line 30: NP: Can we call these classes something like BucketLevelMonitor**Runner** instead to disambiguate from the data model objects?

Also you can make all of these runner types `objects` so they're singletons if you don't expect there to be more than one instance of it.

PR #367, opened 2022-03-29
Update mapping and imports for breaking changes for 2.0

lezzago:

*Issue #, if available:*
- Resolve build issues with OS 2.0
*Description of changes:*
- Update mapping and imports for breaking changes for 2.0
*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/alerts/AlertIndices.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/alerts/AlertIndices.kt
@@ -212,7 +212,7 @@ class AlertIndices(
         if (existsResponse.isExists) return true
 
         val request = CreateIndexRequest(index)
-            .mapping(MAPPING_TYPE, alertMapping(), XContentType.JSON)
+            .mapping(alertMapping())
             .settings(Settings.builder().put("index.hidden", true).build())
 
         if (alias != null) request.alias(Alias(alias))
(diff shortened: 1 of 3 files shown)


qreshi:
It looks like this PR already has most of the changes we need to build with 2.0.0-alpha1.

I made the changes locally to get the build to pass and the diff can be found [here](https://github.com/opensearch-project/alerting/compare/main...qreshi:bump-to-alpha1). Do you mind rebasing with `main` and including the diff linked above in this PR and we can get it merged in?

I added the build qualifier in the GitHub Actions runs because otherwise it pulls in the non-alpha1 version of common-utils. Also, later on we can look into removing `7.10.2` as the old version in the BWC tests since I don't think we're expecting to support upgrades from ODFE -> OS 2.0

qreshi:
It looks like this PR already has most of the changes we need to build with 2.0.0-alpha1.

I made the changes locally to get the build to pass and the diff can be found [here](https://github.com/opensearch-project/alerting/compare/main...qreshi:bump-to-alpha1). Do you mind rebasing with `main` and including the diff linked above in this PR and we can get it merged in?

I added the build qualifier in the GitHub Actions runs because otherwise it pulls in the non-alpha1 version of common-utils. Also, later on we can look into removing `7.10.2` as the old version in the BWC tests since I don't think we're expecting to support upgrades from ODFE -> OS 2.0

PR #376, opened 2022-03-31
Completely fix docker pull and install plugin

peterzhuamazon:
Signed-off-by: Peter Zhu <zhujiaxi@amazon.com>

*Issue #, if available:*
#348

*Description of changes:*
Completely fix docker pull and install plugin

*CheckList:*

### DIFF
--- a/.github/workflows/security-test-workflow.yml
+++ b/.github/workflows/security-test-workflow.yml
@@ -43,22 +43,20 @@ jobs:
           plugin_version=`echo $plugin|awk -F- '{print $3}'| cut -d. -f 1-4`
           qualifier=`echo $plugin|awk -F- '{print $4}'| cut -d. -f 1-1`
           candidate_version=`echo $plugin|awk -F- '{print $5}'| cut -d. -f 1-1`
-          docker_version=$version-$qualifier-candidate_version
+          docker_version=$version-$qualifier
 
-          if [ -z $candidate_version ]; then
-              candidate_version=$qualifier && qualifier=""
-              docker_version=$version
-          fi
+          [[ -z $candidate_version ]] && candidate_version=$qualifier && qualifier=""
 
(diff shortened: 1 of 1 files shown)


peterzhuamazon:
Expected failure on the security as the docker image itself does not have security and pa plugin.

peterzhuamazon:
The pulling mechanic is running well now as well as installation.

AWSHurneyt (approved):

.github/workflows/security-test-workflow.yml
```diff
@@ -36,29 +36,27 @@ jobs:
...
           qualifier=`echo $plugin|awk -F- '{print $4}'| cut -d. -f 1-1`
           candidate_version=`echo $plugin|awk -F- '{print $5}'| cut -d. -f 1-1`
-          docker_version=$version-$qualifier-candidate_version
+          docker_version=$version-$qualifier
 
-          if [ -z $candidate_version ]; then
-              candidate_version=$qualifier && qualifier=""
-              docker_version=$version
-          fi
+          [[ -z $candidate_version ]] && candidate_version=$qualifier && qualifier=""
```
qreshi, line 48: Is this setting `candidate_version` to `qualifier`'s value when it's empty? I don't see us using `$candidate_version` anymore below since `$plugin` is used to fetch the plugin zip, can it be removed?

peterzhuamazon: I would still keep it here just in case we change the strategy later to include candidate in docker image.
This is also good for us to track all these vars and its values.
No harm.

qreshi: Typically we should remove unused code and use the git history to reintroduce it if we need it. Not a blocker though, I'll approve.

qreshi (approved):

Issue #105, opened 2021-06-09
Integrate Alerting with the Notification plugin


praveensameneni:
The Alerting plugin plans to integrate with the Notification plugin in 2.0. This will remove the `notification` subproject logic and will delegate to the Notification plugin for managing things like destination configurations and publishing the messages for the  Notification Actions that Monitors have today.

qreshi:
With this integration comes some breaking changes, here is what we are expecting to happen:

### For 2.0
#### Breaking Changes
* All Destination API will be removed (except perhaps the Get Destinations API so that users can see what Destinations have yet to migrate, UI can also show a read-only state using this API, etc.), any CRUD operations regarding Notification Channels will have to be done through the Notification plugin API
* Notification jar logic in Alerting will be removed (this is mostly breaking for something like ISM which will have to use the Notification plugin to publish now which is expected)

#### Non-Breaking but worth noting
* A background process will be running to migrate any Destinations to Notification Channels on the users behalf (deleting Destinations that have successfully migrated)
* The Monitor execution will automatically use the Notification Channel for Action execution if it was migrated, falling back to Destinations (via an internal passthrough API) in the cases where migration has not yet happened or failed
* Any Destination specific plugin settings will be deprecated but not removed and have Notification versions which will inherit from the Alerting setting value (to ensure a smooth transition)
  * This includes the keystore settings being used for Email Destinations (we’ll be keeping both for now, Notifications’ takes precedence and users will need to make sure they’re moving over to the Notification namespace during 2.x)


### For 3.0
#### Breaking Changes
* The background process for Destination -> Notification Channel migration will be removed (we don’t expect to handle migrating from OS 1.x -> 3.0 so the user will need to go through 2.x -> 3.0 and ensure their Destinations have migrated prior to upgrade)
* Monitor runner fallback can be removed as we don’t expect to have Destinations anymore
* Get Destinations API will be removed
* Alerting Destination keystore settings (for email) will be removed
 
#### Non-Breaking but worth noting
* All other Alerting Destination settings will be removed (these are non-breaking because in 2.x the Notification versions of these settings are taking precedence anyway)

dblock:
I believe this was done, @qreshi correct me if I am wrong? Closing.

qreshi:
Not done yet :)

dblock:
What's left?

PR #379, opened 2022-04-01
Add automated migration for Destinations to Notification configs

qreshi:
*Issue #, if available:* #105

*Description of changes:*
This PR adds a background job that will migrate Destinations to Notification configs as part of Alerting's integration with the Notification plugin for a more seamless transition to 2.0.

* The background job will be triggered on a cluster change event (if we feel the need to have further control on this, we can revisit adding a plugin setting to re-enable or permanently disable this job as well)
* The background job once scheduled will run every minute and attempt to move all Destinations to Notification configs, once all Destinations are moved over, the job will disable itself
* The migration job has a lock that is used to ensure that there is no more than one instance of the job running at a time
* Once a Destination is migrated over to a Notification config, the Destination will be deleted
* The migrated Notification configs will be created with the same ID as the Destinations so that the Monitor doesn't have to have any changes to its references in the Action definition (in a later PR, the MonitorRunner will be updated to be able to execute Actions seamlessly regardless of whether or not the underlying config has been migrated)

*CheckList:*


aditjind:
> The MonitorRunner will be updated to be able to execute Actions seamlessly regardless of whether or not the underlying config has been migrated. 

Does this mean `MonitorRunner` will have the support to execute actions by reading configs from both the `NotificationConfig` and the `AlertingConfig`? Any specific reason we need to support both since we already have a job with features such as locking support etc.

alerting/src/test/kotlin/org/opensearch/alerting/util/destinationmigration/DestinationMigrationUtilServiceIT.kt
```diff
@@ -0,0 +1,105 @@
...
+                      "destination" : ${destination.toJsonString()}
+                    }
+                """.trimIndent()
+                indexDoc(SCHEDULED_JOBS_INDEX, destination.id, dest)
+                ids.add(destination.id)
+            }
+
+            // Create cluster change event and wait for migration service to complete migrating data over
+            client().updateSettings("indices.recovery.max_bytes_per_sec", "40mb")
+            Thread.sleep(120000)
```
lezzago, line 84: Is there a way to reduce this time without making it flaky?

alerting/src/main/kotlin/org/opensearch/alerting/util/destinationmigration/DestinationMigrationUtilService.kt
```diff
@@ -0,0 +1,221 @@
...
+    companion object {
+
+        private val logger = LogManager.getLogger(DestinationMigrationUtilService::class)
+
+        @Volatile
+        private var runningLock = false // In case 2 migrateDestinations() processes are running
+
+        // Used in DestinationMigrationCoordinator to cancel scheduled process
+        @Volatile
+        var finishFlag = false
```
lezzago, line 53: Did we want to make this flag a cluster setting to get reset later on?

alerting/src/main/kotlin/org/opensearch/alerting/util/destinationmigration/DestinationMigrationUtilService.kt
```diff
@@ -0,0 +1,221 @@
...
+            }
+            try {
+                runningLock = true
+
+                val destinationsToMigrate = retrieveDestinationsToMigrate(client)
+                logger.info("Need to migrate ${destinationsToMigrate.size} destinations")
+                if (destinationsToMigrate.isEmpty()) {
+                    finishFlag = true
+                    runningLock = false
+                    return
```
lezzago, line 72: Need to delete the old destinations after the migration still

alerting/src/main/kotlin/org/opensearch/alerting/util/destinationmigration/DestinationMigrationUtilService.kt
qreshi: Let's treat that as a potential enhancement for now. If we see a need for it, we can follow-up to add it.

alerting/src/test/kotlin/org/opensearch/alerting/util/destinationmigration/DestinationMigrationUtilServiceIT.kt
qreshi: Hmm, the job will wait 1 minute before actually running for the first time. So if we think once running it doesn't take too long, we might be able to reduce it to as little over 1 minute as possible.

I won't treat it as a blocker for merging if that's fine with you but we can follow-up on it during general testing.

qreshi:
> > The MonitorRunner will be updated to execute Actions seamlessly whether we have migrated the underlying config.
> 
> Does this mean `MonitorRunner` will have the support to execute actions by reading configs from both the `NotificationConfig` and the `AlertingConfig`? Any specific reason we need to support both since we already have a job with features such as locking support etc.

That's correct, the MonitorRunner will have changes as a follow-up to this PR so that the Monitor can continue to run at any time regardless of which Destinations has been migrated or not.

The `runningLock` for `DestinationMigrationCoordinator` is not going to halt Monitor executions during migration, that's just to ensure that multiple instances of the migration job aren't running concurrently.

alerting/src/test/kotlin/org/opensearch/alerting/util/destinationmigration/DestinationMigrationUtilServiceIT.kt
lezzago: sure thats fine

alerting/src/main/kotlin/org/opensearch/alerting/util/destinationmigration/NotificationApiUtils.kt
```diff
@@ -0,0 +1,153 @@
...
+import org.opensearch.commons.notifications.action.DeleteNotificationConfigRequest
+import org.opensearch.commons.notifications.action.DeleteNotificationConfigResponse
+import org.opensearch.commons.notifications.action.GetNotificationConfigRequest
+import org.opensearch.commons.notifications.action.GetNotificationConfigResponse
+import org.opensearch.commons.notifications.action.SendNotificationRequest
+import org.opensearch.commons.notifications.action.SendNotificationResponse
+import org.opensearch.commons.notifications.action.UpdateNotificationConfigRequest
+import org.opensearch.commons.notifications.action.UpdateNotificationConfigResponse
+
+class NotificationApiUtils {
```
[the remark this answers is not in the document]
lezzago: There will only be a single implementation on Alerting for this and this is needed to talk to Notification, so I dont see any benefit with abstracting it

lezzago (approved):

alerting/src/main/kotlin/org/opensearch/alerting/util/destinationmigration/NotificationApiUtils.kt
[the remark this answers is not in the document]
aditjind: Cleaner code maybe? I mean we can keep it as it is for now.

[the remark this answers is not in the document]
qreshi: Yeah, like Ashish said, there won't be multiple implementations of this so an interface wouldn't be too helpful for the utils.

AWSHurneyt (approved):

PR #401, opened 2022-04-13
Integrate with Notifications plugin for Alerting backend

qreshi:
*Issue #, if available:* #105

*Description of changes:*
This PR depends on https://github.com/opensearch-project/common-utils/pull/158 and https://github.com/opensearch-project/notifications/pull/406 so the GitHub Actions are expected to fail without at least `common-utils` being merged in

The main changes in this PR are:
* Integrating with the Notifications plugin, supporting execution of the notification actions before/after migration of the Destination configs to Notifications
  * In the case of the Destination failing migration, it will be sent through the passthrough API (this avoids hard failures on a write block but still requires the Notification plugin to be installed since the passthrough API does not work without it)
* Fixes some bugs
  * For the Destination migration, `email_group` configs were being checked with a string match which could also appear in the Email Destination (this would cause it to use the wrong parser so the fetching of the different config types has been split up)
  * The Alerting email keystore settings were not correctly declaring the fallback setting to the legacy `opendistro*.` variants this would cause runtime errors when it tried to load the keystore settings and failed on falling back to the old setting
    * This has been fixed (although it should not be used anymore since the Notification plugin has its own namespace for these settings that will take precedence, Alerting's will be used as a fallback for that)

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunner.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunner.kt
@@ -647,33 +647,12 @@ object MonitorRunner : JobRunner, CoroutineScope, AbstractLifecycleComponent() {
         }
     }
 
-    private suspend fun runAction(action: Action, ctx: QueryLevelTriggerExecutionContext, dryrun: Boolean): ActionRunResult {
+    private suspend fun runAction(action: Action, ctx: TriggerExecutionContext, dryrun: Boolean): ActionRunResult {
         return try {
-            if (!isActionActionable(action, ctx.alert)) {
+            if (ctx is QueryLevelTriggerExecutionContext && !isActionActionable(action, ctx.alert)) {
                 return ActionRunResult(action.id, action.name, mapOf(), true, null, null)
             }
-            val actionOutput = mutableMapOf<String, String>()
-            actionOutput[SUBJECT] = if (action.subjectTemplate != null) compileTemplate(action.subjectTemplate, ctx) else ""
-            actionOutput[MESSAGE] = compileTemplate(action.messageTemplate, ctx)
-            if (Strings.isNullOrEmpty(actionOutput[MESSAGE])) {
-                throw IllegalStateException("Message content missing in the Destination with id: ${action.destinationId}")
-            }
-            if (!dryrun) {
-                // TODO: Inject user here so only Destination/Notifications that the user has permissions to are retrieved
-                withContext(Dispatchers.IO) {
-                    actionOutput[MESSAGE_ID] = getConfigAndSendNotification(action, actionOutput[SUBJECT], actionOutput[MESSAGE]!!)
-                }
-            }
-            ActionRunResult(action.id, action.name, actionOutput, false, currentTime(), null)
-        } catch (e: Exception) {
-            ActionRunResult(action.id, action.name, mapOf(), false, currentTime(), e)
-        }
-    }
 
-    // TODO: This is largely a duplicate of runAction above for BucketLevelTriggerExecutionContext for now.
-    //   After suppression logic implementation, if this remains mostly the same, it can be refactored.
-    private suspend fun runAction(action: Action, ctx: BucketLevelTriggerExecutionContext, dryrun: Boolean): ActionRunResult {
-        return try {
             val actionOutput = mutableMapOf<String, String>()
             actionOutput[SUBJECT] = if (action.subjectTemplate != null) compileTemplate(action.subjectTemplate, ctx) else ""
             actionOutput[MESSAGE] = compileTemplate(action.messageTemplate, ctx)
--- a/alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunner.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunner.kt
@@ -35,6 +35,7 @@ import org.opensearch.alerting.model.action.ActionExecutionScope
 import org.opensearch.alerting.model.action.AlertCategory
 import org.opensearch.alerting.model.action.PerAlertActionScope
 import org.opensearch.alerting.model.action.PerExecutionActionScope
+import org.opensearch.alerting.model.destination.Destination
 import org.opensearch.alerting.model.destination.DestinationContextFactory
 import org.opensearch.alerting.opensearchapi.InjectorContextElement
 import org.opensearch.alerting.opensearchapi.retry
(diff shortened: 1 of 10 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/util/destinationmigration/DestinationMigrationUtilService.kt
```diff
@@ -64,17 +64,25 @@ class DestinationMigrationUtilService {
             try {
                 runningLock = true
 
-                val destinationsToMigrate = retrieveDestinationsToMigrate(client)
-                logger.info("Need to migrate ${destinationsToMigrate.size} destinations")
-                if (destinationsToMigrate.isEmpty()) {
+                val emailAccountsToMigrate = retrieveConfigsToMigrate(client, "email_account")
```
rishabhmaurya, line 67: Can you add a comment for this method if is thread safe? Is it idempotent? Can this be scheduled on other nodes of the cluster?

alerting/src/main/kotlin/org/opensearch/alerting/util/destinationmigration/DestinationMigrationUtilService.kt
```diff
@@ -64,17 +64,25 @@ class DestinationMigrationUtilService {
...
+                if (configsToMigrate.isEmpty()) {
                     finishFlag = true
                     runningLock = false
                     return
                 }
-                val migratedDestinations = createNotificationChannelIfNotExists(client, destinationsToMigrate)
-                logger.info("Migrated ${migratedDestinations.size} destinations")
-                val failedDeletedDestinations = deleteOldDestinations(client, migratedDestinations)
-                logger.info("Failed to delete ${failedDeletedDestinations.size} destinations from migration process cleanup")
+                val migratedConfigs = createNotificationChannelIfNotExists(client, configsToMigrate)
```
rishabhmaurya, line 82: This might be scope of this PR, but what if the user who created the notification config, is deleted? Will it still migrate it? Or will it fail and keep retrying forever?

alerting/src/main/kotlin/org/opensearch/alerting/util/destinationmigration/DestinationMigrationUtilService.kt
```diff
@@ -64,17 +64,25 @@ class DestinationMigrationUtilService {
...
+                val destinationsToMigrate = retrieveConfigsToMigrate(client, "destination")
+                val configsToMigrate = emailAccountsToMigrate + emailGroupsToMigrate + destinationsToMigrate
+                logger.info(
+                    "Need to migrate ${emailAccountsToMigrate.size} email accounts, " +
+                        "${emailGroupsToMigrate.size} email groups and " +
+                        "${destinationsToMigrate.size} destinations " +
+                        "(${configsToMigrate.size} configs total)"
+                )
+                if (configsToMigrate.isEmpty()) {
                     finishFlag = true
```
rishabhmaurya, line 78: Do we persist this flag? Is this job supposed to run every time the process restarts even if the migration has completed? Wanted to understand the expected lifecycle of this job. It would have been easier if there was a comment on lifecycle!

alerting/src/main/kotlin/org/opensearch/alerting/util/destinationmigration/DestinationMigrationUtilService.kt
qreshi: Yeah, if you look a few lines above, you'll see a `runningLock` with a comment. That's used to ensure that only a single instance of `migrateDestinations()` is ever running.

The `DestinationMigrationCoordinator` is what schedules this job and it uses a similar mechanism (and is also only scheduled by the elected cluster manager node)

alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunner.kt
```diff
@@ -670,41 +671,67 @@ object MonitorRunner : JobRunner, CoroutineScope, AbstractLifecycleComponent() {
...
-        } catch (e: Exception) {
-            ActionRunResult(action.id, action.name, mapOf(), false, currentTime(), e)
+        if (config.destination == null && config.channel == null) {
+            throw IllegalStateException("Unable to find a Notification Channel or Destination config with id [${action.id}]")
+        }
+
+        // Adding a check on TEST_ACTION Destination type here to avoid supporting it as a LegacyBaseMessage type
+        // just for Alerting integration tests
+        if (config.destination?.isTestAction() == true) {
+            return "test action"
```
rishabhmaurya, line 684: can we define this constant somewhere else and use it both at the time assert and here.

alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunner.kt
```diff
@@ -670,41 +671,67 @@ object MonitorRunner : JobRunner, CoroutineScope, AbstractLifecycleComponent() {
...
+        }
+
+        if (config.destination?.isAllowed(allowList) == false) {
```
rishabhmaurya, line 687: can we rename `allowList ` to `destinationsAllowList` for better readability.

alerting/src/main/kotlin/org/opensearch/alerting/util/destinationmigration/DestinationMigrationUtilService.kt
qreshi: You mean the user we're trying to inject in the thread context for the object is deleted?

Hmm, Security might handle that and treat it as no permissions? We'll want to test that when we test Alerting/Notif security integration. That's out of scope for this PR for now but a good callout.

alerting/src/main/kotlin/org/opensearch/alerting/util/destinationmigration/DestinationMigrationUtilService.kt
qreshi: Yeah so the job will finish and deschedule if there are no more Destination configs to migrate.

On a cluster change event, it would reschedule and check if there are any Destination configs (assuming the user didn't restore from snapshot or something this should change since they can't create new ones anymore) and will disable itself again.

rishabhmaurya (approved):

alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunner.kt
getsaurabh02: +1

alerting/src/main/kotlin/org/opensearch/alerting/settings/DestinationSettings.kt
```diff
@@ -85,6 +96,14 @@ class DestinationSettings {
             return concreteSetting.get(settings)
         }
 
+        private fun <T> fallback(key: String, affixSetting: AffixSetting<T>, regex: String, replacement: String): Setting<T>? {
+            return if ("_na_" == key) {
```
getsaurabh02, line 100: same as above

getsaurabh02 (approved):

qreshi:
As discussed, the non-blocking comments will be addressed in a follow-up.

Also, the only failing GitHub Actions now that the dependent PRs are merged in is the security workflow which is being fixed in #407, merging in for now.

PR #405, opened 2022-04-15
Add Trigger condition parser and resolver for Doc Level Alerts

getsaurabh02:
Add Trigger condition parser which parses the provided trigger condition expression in the Reverse Polish Notation and then 
resolves the final set of Document Ids which matches the given expression for Document Level Alerts.
 
 Users can configure the Trigger condition using the expression strings as below:
 

**Always True**
``
return true
``

**Always False**
``
return false
``

**Query Expressions**
```
* (query[name=sigma-123] && query[name=sigma-456]) || (!query[id=aert34df] && query[tag=sev2])
* query[tag=sev1]
* (query[name=sigma-123] && query[tag=sev2]) || (!query[tag=sev3] && query[id=dfd457iy] && query[tag=“sev1"])
```
Operartors supported are:
```
* (
* )
* &&
* ||
* !
```

Query Identifiers supported are:
```
* name ---> query[name=sigma-123]
* tag ---> query[tag=sev2]
* id ---> query[id=aert34df]
```

Signed-off-by: Saurabh Singh <sisurab@amazon.com>

*Issue #, if available:*

*Description of changes:*

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/TriggerService.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/TriggerService.kt
@@ -8,6 +8,7 @@ package org.opensearch.alerting
 import org.apache.logging.log4j.LogManager
 import org.opensearch.alerting.aggregation.bucketselectorext.BucketSelectorIndices.Fields.BUCKET_INDICES
 import org.opensearch.alerting.aggregation.bucketselectorext.BucketSelectorIndices.Fields.PARENT_BUCKET_PATH
+import org.opensearch.alerting.core.model.DocLevelQuery
 import org.opensearch.alerting.model.AggregationResultBucket
 import org.opensearch.alerting.model.Alert
 import org.opensearch.alerting.model.BucketLevelTrigger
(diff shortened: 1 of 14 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/TriggerService.kt
```diff
@@ -60,31 +63,18 @@ class TriggerService(val scriptService: ScriptService) {
...
+                    triggeredDocs.addAll(value)
                 }
-                val triggered = scriptService.compile(trigger.condition, TriggerScript.CONTEXT)
-                    .newInstance(params)
-                    .execute(dummyExecutionContext)
-                logger.info("trigger val: $triggered")
-                if (triggered) triggeredDocs.add(doc)
+            } else if (!trigger.condition.idOrCode.equals(NEVER_RUN.idOrCode)) {
+                triggeredDocs = TriggerExpressionParser(trigger.condition.idOrCode).parse()
+                    .evaluate(queryToDocIds).toMutableList()
```
sbcd90, line 77: the `if-else if` block moves the `object equals check` to runtime. Kotlin allows this check to be made at compile time using `when` control statements.

```
when (trigger.condition.idOrCode) {
                ALWAYS_RUN.idOrCode  -> {
                    for (value in queryToDocIds.values) {
                        triggeredDocs.addAll(value)
                    }
                }
                NEVER_RUN.idOrCode -> {
                    triggeredDocs = TriggerExpressionParser(trigger.condition.idOrCode).parse()
                        .evaluate(queryToDocIds).toMutableList()
                }
            }
```
this will allow catching runtime errors we faced today at compile time only.

alerting/src/main/kotlin/org/opensearch/alerting/triggercondition/parsers/TriggerExpressionParser.kt
```diff
@@ -0,0 +1,53 @@
...
+
+import org.opensearch.alerting.triggercondition.resolvers.TriggerExpressionRPNResolver
+import org.opensearch.alerting.triggercondition.tokens.TriggerExpressionOperator
+
+/**
+ * The postfix (Reverse Polish Notation) parser.
+ * Uses the Shunting-yard algorithm to parse a mathematical expression
+ * @param triggerExpression String containing the trigger expression for the monitor
+ */
+class TriggerExpressionParser(
```
sbcd90, line 16: should the class name be changed to `TriggerRPNExpressionParser`? 
`parse()` returns `TriggerExpressionRPNResolver`
tomorrow if someone wants to have another implementation of interface `ExpressionParser`

sbcd90 (approved):

alerting/src/main/kotlin/org/opensearch/alerting/triggercondition/parsers/TriggerExpressionParser.kt
getsaurabh02: Since currently there is only one concrete implementation of the Trigger Expression Parser have kept the name as is for clarity. However, the base implementation which holds the RPN logic is named as the ``TriggerExpressionRPNBaseParser``. 
Later addition of new base implementations are possible using the new strategy, while the concrete implementation could still expose them via different method interfaces.

alerting/src/main/kotlin/org/opensearch/alerting/TriggerService.kt
getsaurabh02: Thanks for the suggestion @sbcd90. Agree, control statement could make the flow look much cleaner and I have used it in the Parser/Evaluation logic. However, here since this is not an instance check, but rather the value check with string comparison, I am not sure having control statement would add any value here in terms of performance. Given string value comparison still need to be performed at runtime.

Since we weren't using all the three possible values here, I chose to keep it as if-else-of block instead.

AWSHurneyt (approved):

PR #407, opened 2022-04-16
Fix security integration tests for 2.0

sbcd90:
Signed-off-by: Subhobrata Dey <sbcd90@gmail.com>

*Issue #, if available:*
[[BUG] Security Integration tests are failing for 2.0](https://github.com/opensearch-project/alerting/issues/355)

*Description of changes:*

Fix security integration tests for 2.0

*CheckList:*

### DIFF
--- a/alerting/src/test/kotlin/org/opensearch/alerting/AlertingRestTestCase.kt
+++ b/alerting/src/test/kotlin/org/opensearch/alerting/AlertingRestTestCase.kt
@@ -596,9 +596,17 @@ abstract class AlertingRestTestCase : ODFERestTestCase() {
         client.makeRequest("POST", "$ALERTING_BASE_URI/_execute", params, monitor.toHttpEntityWithUser())
 
     protected fun indexDoc(index: String, id: String, doc: String, refresh: Boolean = true): Response {
+        return indexDoc(client(), index, id, doc, refresh)
+    }
+
+    protected fun indexDocWithAdminClient(index: String, id: String, doc: String, refresh: Boolean = true): Response {
+        return indexDoc(adminClient(), index, id, doc, refresh)
+    }
+
+    private fun indexDoc(client: RestClient, index: String, id: String, doc: String, refresh: Boolean = true): Response {
         val requestBody = StringEntity(doc, APPLICATION_JSON)
         val params = if (refresh) mapOf("refresh" to "true") else mapOf()
-        val response = client().makeRequest("PUT", "$index/_doc/$id", params, requestBody)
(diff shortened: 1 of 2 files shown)


praveensameneni (approved):

alerting/src/test/kotlin/org/opensearch/alerting/AlertingRestTestCase.kt
```diff
@@ -598,7 +598,7 @@ abstract class AlertingRestTestCase : ODFERestTestCase() {
     protected fun indexDoc(index: String, id: String, doc: String, refresh: Boolean = true): Response {
         val requestBody = StringEntity(doc, APPLICATION_JSON)
         val params = if (refresh) mapOf("refresh" to "true") else mapOf()
-        val response = client().makeRequest("PUT", "$index/_doc/$id", params, requestBody)
+        val response = adminClient().makeRequest("PUT", "$index/_doc/$id", params, requestBody)
```
qreshi, line 601: Do you know if changing this to `adminClient()` has any side-effects to the other tests that were using `indexDoc` already? As in, they were trying to index against a regular index with some user in a security test and we expect it to only pass if they have permissions but now it'll always work with `adminClient()`?

If we want to keep this generic, we can overload the function to keep the old behavior and have another method for the `DestinationMigrationUtilServiceIT` cases:

```kotlin
protected fun indexDoc(index: String, id: String, doc: String, refresh: Boolean = true): Response {
    return indexDoc(client(), index, id, doc, refresh)
}

protected fun indexDocWithAdminClient(index: String, id: String, doc: String, refresh: Boolean = true): Response {
	return indexDoc(adminClient(), index, id, doc, refresh)
}

private fun indexDoc(client: RestClient, index: String, id: String, doc: String, refresh: Boolean): Response {
    val requestBody = StringEntity(doc, APPLICATION_JSON)
    val params = if (refresh) mapOf("refresh" to "true") else mapOf()
    val response = client().makeRequest("PUT", "$index/_doc/$id", params, requestBody)
    assertTrue(
        "Unable to index doc: '${doc.take(15)}...' to index: '$index'",
        listOf(RestStatus.OK, RestStatus.CREATED).contains(response.restStatus())
    )
    return response
}
```

Then you can use `indexDocWithAdminClient` in `DestinationMigrationUtilServiceIT`

sbcd90: thanks Mohammad for pointing that out. changed it now.

qreshi (approved):

Issue #254, opened 2021-12-10
Add support for multi-bucket aggregation in "Per bucket monitor"


davidshtian:
**Is your feature request related to a problem? Please describe.**
From the description of the docs [here](https://opensearch.org/docs/latest/monitoring-plugins/alerting/monitors/#create-triggers), it seems that only **_composite aggregation_** query is supported in **_Per bucket monitor_**, while sometime users might need [multi-bucket](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-multi-terms-aggregation.html#search-aggregations-bucket-multi-terms-aggregation) aggregation to implement features like sorting or parameters like min_doc_count, etc. 

And then compare the values for each bucket, it might be similar to the Elasticsearch [watch array compare condition](https://www.elastic.co/guide/en/elasticsearch/reference/current/condition-array-compare.html) feature.

**Describe the solution you'd like**
To support multi-bucket aggregation in monitor triggers extraction query.

**Describe alternatives you've considered**
N/A.

**Additional context**
N/A.

davidshtian:
Besides, there're very few description and examples within docs for alerting monitor and trigger, and it might be hard for OpenSearch users to debug and troubleshooting of configurations. Thanks~

rishabhmaurya:
Multi-bucket aggregation is supported in bucket monitors. I can see you are referring to multi-term aggregation in the question, assuming that's what you meant, then you're right as we don't support `multi_terms` aggregation yet in opensearch. It is part of xpack analytics plugins in elasticsearch and opensearch don't have anything similar yet.
Also, its not just composite aggregation, but all bucket, metric aggregations [here](https://github.com/opensearch-project/OpenSearch/tree/main/server/src/main/java/org/opensearch/search/aggregations) are supported with bucket selector.

I agree that documentation needs better examples, but while we fix it, below are some of the examples related to selecting a bucket from search results in bucket monitors, which people might find useful - 

```
curl -X GET "localhost:9200/_search?pretty" -H 'Content-Type: application/json' -d'                                                                                                                      ~
{
   "size": 0,
   "aggs": {
      "sports_agg_multi": {
         "terms": {"field": "sport"},
         "aggs": {
            "goals_stats": { "stats": { "field": "goals" } }
         }
      },
      "trigger_bucket_selector1": {
         "bucket_selector_ext": {
            "buckets_path": {
              "avg": "goals_stats.avg"
            },
            "script": {
[...]

PR #428, opened 2022-04-26
Fix UnsupportedOperation error while alert categorization in BucketLevel monitor

rishabhmaurya:
*Issue #, if available:*
https://github.com/opensearch-project/alerting/issues/254
*Description of changes:*
`toMap()`  has special implementation when size is 1 and returns a readOnly map, which causes trouble later in the code while modifying this map. Explicitly converting it to a `MutableMap`
```
public fun <K, V> Iterable<Pair<K, V>>.toMap(): Map<K, V> {
    if (this is Collection) {
        return when (size) {
            0 -> emptyMap()
            1 -> mapOf(if (this is List) this[0] else iterator().next())
            else -> toMap(LinkedHashMap<K, V>(mapCapacity(size)))
        }
    }
    return toMap(LinkedHashMap<K, V>()).optimizeReadOnlyMap()
}
```

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
@@ -95,8 +95,8 @@ class AlertService(
             (
                 foundAlerts[trigger.id]?.mapNotNull { alert ->
                     alert.aggregationResultBucket?.let { it.getBucketKeysHash() to alert }
-                }?.toMap() ?: mutableMapOf()
-                ) as MutableMap<String, Alert>
+                }?.toMap()?.toMutableMap() ?: mutableMapOf()
+                )
         }
     }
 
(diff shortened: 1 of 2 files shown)


qreshi:
Is it possible to add a test to check for the fix? If not, did you try to reproduce the scenario from the issue manually and confirm that it's not happening anymore?

qreshi (approved):

praveensameneni (approved):

rishabhmaurya:
Added unit test

rishabhmaurya:
> Is it possible to add a test to check for the fix? If not, did you try to reproduce the scenario from the issue manually and confirm that it's not happening anymore?

qreshi:
Looks like the CI workflow failed on some flaky test for the BWC test that I haven't seen before. The assertion implies there were two nodes in the cluster and not three. We can make a note of it and merge in for now since it's not related to this change.

PR #441, opened 2022-05-10
Fix minor bugs and support per alert action execution for Document Level Monitors

lezzago:
*Issue #, if available:*
N/A
*Description of changes:*
- Fix minor bugs
  - Validate query name or tags for DocLeveQueries
  - Fix Document Level monitor wildcard index support
- Support per alert action execution for Document Level Monitors
- Creates a new metadata document to help support future throttling use cases and robustness of Document Level Monitors
*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
@@ -33,8 +33,8 @@ import org.opensearch.alerting.opensearchapi.string
 import org.opensearch.alerting.opensearchapi.suspendUntil
 import org.opensearch.alerting.script.DocumentLevelTriggerExecutionContext
 import org.opensearch.alerting.util.AlertingException
+import org.opensearch.alerting.util.defaultToPerExecutionAction
 import org.opensearch.alerting.util.getActionExecutionPolicy
-import org.opensearch.alerting.util.updateMonitor
 import org.opensearch.alerting.util.updateMonitorMetadata
 import org.opensearch.client.Client
 import org.opensearch.cluster.routing.ShardRouting
@@ -83,22 +83,24 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
         try {
             validate(monitor)
         } catch (e: Exception) {
-            logger.info("Failed to start Document-level-monitor. Error: ${e.message}")
+            logger.error("Failed to start Document-level-monitor. Error: ${e.message}")
             return monitorResult.copy(error = AlertingException.wrap(e))
         }
+
+        monitorCtx.docLevelMonitorQueries!!.initDocLevelQueryIndex()
+        monitorCtx.docLevelMonitorQueries!!.indexDocLevelQueries(
+            monitor = monitor,
+            monitorId = monitor.id,
+            indexTimeout = monitorCtx.indexTimeout!!
+        )
 
         val docLevelMonitorInput = monitor.inputs[0] as DocLevelMonitorInput
         val index = docLevelMonitorInput.indices[0]
         val queries: List<DocLevelQuery> = docLevelMonitorInput.queries
 
-        val monitorMetadata = if (monitor.metadataId.isNullOrEmpty()) createMonitorMetadata(monitor.id)
-        else {
-            try {
-                getMonitorMetadata(monitorCtx.client!!, monitorCtx.xContentRegistry!!, monitor.metadataId)
-            } catch (e: Exception) {
-                logger.warn("Monitor metadata with id, ${monitor.metadataId}, is missing for monitor: ${monitor.id}")
-                createMonitorMetadata(monitor.id)
-            }
+        var monitorMetadata = getMonitorMetadata(monitorCtx.client!!, monitorCtx.xContentRegistry!!, "${monitor.id}-metadata")
+        if (monitorMetadata == null) {
+            monitorMetadata = createMonitorMetadata(monitor.id)
         }
 
         val isTempMonitor = dryrun || monitor.id == Monitor.NO_ID
@@ -128,7 +130,8 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
             indices.forEach { indexName ->
                 // Prepare lastRunContext for each index
                 val indexLastRunContext = lastRunContext.getOrPut(indexName) {
-                    createRunContext(monitorCtx.clusterService!!, monitorCtx.client!!, indexName)
+                    val indexCreatedRecently = createdRecently(monitor, indexName, periodStart, periodEnd, getIndexResponse)
+                    createRunContext(monitorCtx.clusterService!!, monitorCtx.client!!, indexName, indexCreatedRecently)
                 }
 
                 // Prepare updatedLastRunContext for each index
@@ -202,14 +205,6 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
         // Don't update monitor if this is a test monitor
         if (!isTempMonitor) {
             updateMonitorMetadata(monitorCtx.client!!, monitorCtx.settings!!, monitorMetadata.copy(lastRunContext = updatedLastRunContext))
-            if (monitor.metadataId.isNullOrEmpty() || monitor.metadataId != monitorMetadata.id) {
-                updateMonitor(
-                    monitorCtx.client!!,
-                    monitorCtx.xContentRegistry!!,
-                    monitorCtx.settings!!,
-                    monitor.copy(metadataId = monitorMetadata.id)
-                )
-            }
         }
 
         // TODO: Update the Document as part of the Trigger and return back the trigger action result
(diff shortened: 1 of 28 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -70,6 +72,7 @@ import org.opensearch.tasks.Task
 import org.opensearch.transport.TransportService
 import java.io.IOException
 import java.time.Duration
+import java.util.*
```
sbcd90, line 75: remove `java.util.*` import

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -502,6 +524,11 @@ class TransportIndexMonitorAction @Inject constructor(
                     )
                     return
                 }
+
+                if (indexMetadataRequest != null) {
+                    val indexMetadataResponse: IndexResponse = client.suspendUntil { client.index(indexMetadataRequest, it) }
```
sbcd90, line 529: `indexMetadataResponse` is unused.

sbcd90 (approved):

alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunner.kt
```diff
@@ -141,4 +146,41 @@ abstract class MonitorRunner {
 
         return NotificationActionConfigs(destination, channel)
     }
+
+    protected fun defaultToPerExecutionAction(
```
getsaurabh02, line 150: should we move these to be part of a utility class instead?

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -86,28 +91,39 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
         val queries: List<DocLevelQuery> = docLevelMonitorInput.queries
 
+        val monitorMetadata = if (monitor.metadataId.isNullOrEmpty()) createMonitorMetadata(monitor.id)
+        else {
+            try {
+                getMonitorMetadata(monitorCtx.client!!, monitorCtx.xContentRegistry!!, monitor.metadataId)
+            } catch (e: Exception) {
+                logger.warn("Monitor metadata with id, ${monitor.metadataId}, is missing for monitor: ${monitor.id}")
+                createMonitorMetadata(monitor.id)
+            }
```
getsaurabh02, line 101: This is not idempotent as read failures due to multiple reason can result into conflicting write operations? Given monitor metadata has critical state information, such as lastRunContext, we should rather bail out and surface the error.

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -86,28 +91,39 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
-
-            // If the input index is an alias, creating a list of all indices associated with that alias;
-            // else creating a list containing the single index input
-            val indices = if (isAlias) getAliasesResponse.aliases.keys().map { it.value } else listOf(index)
+            // cleanup old indices that are not monitored anymore from the same monitor
+            for (ind in updatedLastRunContext.keys) {
+                if (!indices.contains(ind)) {
+                    updatedLastRunContext.remove(ind)
+                }
+            }
```
getsaurabh02, line 126: If the same index gets added again, the next monitor execution will loose its previous context. Tradeoff of not cleaning up is then there is no good way to stop it from growing otherwise. I am fine with setting up this behavior, however calling out here to ensure that we have thought about it.

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -196,13 +201,15 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
-            updateMonitor(monitorCtx.client!!, monitorCtx.xContentRegistry!!, monitorCtx.settings!!, updatedMonitor)
+            updateMonitorMetadata(monitorCtx.client!!, monitorCtx.settings!!, monitorMetadata.copy(lastRunContext = updatedLastRunContext))
+            if (monitor.metadataId.isNullOrEmpty() || monitor.metadataId != monitorMetadata.id) {
+                updateMonitor(
+                    monitorCtx.client!!,
+                    monitorCtx.xContentRegistry!!,
+                    monitorCtx.settings!!,
+                    monitor.copy(metadataId = monitorMetadata.id)
+                )
+            }
```
getsaurabh02, line 212: Why do we need to update the monitor now, since metadata is already updated with mutable information?

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
lezzago: Yea I understand your concern. I think it is more important to clean it up as this can continuously grow without stopping. Also if an index is removed from this and readded, I think it is fair to assume that is now considered as new data coming into the monitor.

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
lezzago: We update the monitor if the metadata was not created for some reason or the metadata doc that this monitor was pointing to was never created initially, so a new metadata doc was created and this monitor needs to be updated to point to that metadata doc.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
lezzago: done

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
lezzago: fixed

alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunner.kt
lezzago: done

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
lezzago: I refined the exceptions to look for config index not existing and if it could not find the result.

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
lezzago: I also created an issue to find out how to cleanup old rolled over indices with issue #444

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
lezzago: This has been fixed by making the metadata id based on the monitor id so this is no longer needed

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -412,9 +412,19 @@ class TransportIndexMonitorAction @Inject constructor(
                     )
                     return
                 }
+                metadata = metadata.copy(monitorId = indexResponse.id)
+
+                val metadataIndexRequest = IndexRequest(SCHEDULED_JOBS_INDEX)
```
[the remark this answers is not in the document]
lezzago: During the monitor execution, we will have to recreate the metadata. This is being handled by the Document Level Monitor as its the only monitor type currently utilizing it.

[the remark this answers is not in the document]
qreshi: Would you mind making a comment about that here? That's a general error case that we seem to only be handling in Doc-Level Monitors. If we add its usage in other types, we'll need to remember to account for it. Ideally since this is being added to Monitors in general, we should add it to some base initialization logic so all Monitor types handle it by default (with the option to override the default values or something). We don't have to do that in this PR but the comment will be helpful in the meantime.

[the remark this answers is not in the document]
getsaurabh02: I would highly insist on abstracting out the Metadata creation as a separate logic with clear interfaces and exceptions, Same can be re-used across if other monitors start using it. 
Let me know if that is something we can only do as a followup.

getsaurabh02 (approved):

Issue #522, opened 2022-08-16
[FEATURE] Enhancing Monitor Configs to support custom data sources.


eirsep:
**Is your feature request related to a problem?**

Today alerting implements multiple flavors of Monitor Runners to offer different alerting capabilities, such as Query Level Monitor, Bucket Level Monitors and Document Level Monitors. The execution workflow for these monitors are very coupled with their direct functional needs. There is also a tight coupling between the configuration and execution logic. The data sources for both reads and writes are hard coded, with assumption that they would always remain same from the Alerting purview. There is currently no way to store alerts or findings for different users/stakeholders of alerting plugin in different indices. 

**What solution would you like?**
Enhancing Monitor Config

Enhancing monitor configs provides more users with more control over various configurations in each monitor and options to decide input and output sources. These are *optional* params which users of the plugins (such as Security Analytics) can use to injects their own configurations such as source data, query index, findings index etc. helping with separation of concerns in data layer for Alerting Plugin. 

The new addition in the config will be stored as a part of the monitor document in the index itself.

**Monitor configs allow users of alerting to define optional configurable params to override the following defaults:**

**Data Sources** 

We envision Monitors to have multiple data sources, in addition to the current source index. It can read from either indices with user data. Alerting could also read data from findings index which other monitors are writing into. Users can pass source type to be findings instead of the default indices. Alerting would then parse findings index documents to pick up the list of docs to be queried on and treat that as source of the monitor.

Monitors would also be able to consume results of another monitor by consuming findings index documents as source data (TODO)
*Source Type Values: FINDINGS, INDICES*
 *Default Source type:* 

**Query Index** 

Allows users to pass an index where queries will be stored. This override is essential in enabling users of Alerting plugin to build and store rules which can be loaded on demand from rule index.
The current behavior today can be modeled into a default config which can be overridden in the create monitor request for each monitor. 

**Findings Index**
Allows users to create custom set of indices of findings and its history. Users could harness this option to store their alerts

**Alerts Index**

Allows users to create custom set of indices for alerts,  alerts history. Users could harness this option to store their alerts in a separate place.

PR #526, opened 2022-08-17
Adds support in monitor configs for custom alerts, findings, queries indices

eirsep:
*Issue #, if available:*
#522 

*Description of changes:*
- Adds support in monitor configs for accepting custom alerts, findings, queries indices in transport layer
- If any data sources are not passed, the default system indices would be used for storing alerts/queries/findings.
 
*CheckList:*


core/src/main/kotlin/org/opensearch/alerting/core/model/DataSources.kt
```diff
@@ -0,0 +1,101 @@
...
+import org.opensearch.common.io.stream.Writeable
+import org.opensearch.common.xcontent.ToXContent
+import org.opensearch.common.xcontent.ToXContentObject
+import org.opensearch.common.xcontent.XContentBuilder
+import org.opensearch.common.xcontent.XContentParser
+import org.opensearch.common.xcontent.XContentParserUtils
+import java.io.IOException
+
+data class DataSources(
+    val queryIndex: String? = null,
```
getsaurabh02, line 19: Can this be a list too? Or we are assuming it will always be one index to begin with?
Also, for other monitors such as bucket level holds the OS query directly. We will need to think about them as well.

core/src/main/kotlin/org/opensearch/alerting/core/model/DataSources.kt
```diff
@@ -0,0 +1,101 @@
...
+    val queryIndexMapping: String? = null,
```
getsaurabh02, line 20: What is the purpose of the index mapping as a data source here?

core/src/main/resources/mappings/scheduled-jobs.json
```diff
@@ -118,6 +118,46 @@
...
+              "type": "text",
+              "fields": {
+                "keyword": {
+                  "type": "keyword",
+                  "ignore_above": 256
+                }
+              }
+            }
+          }
+        },
```
getsaurabh02, line 160: What are the pros and cons of keeping as part of the metadata/config vs as a separate object part of the same index?

core/src/main/kotlin/org/opensearch/alerting/core/model/DataSources.kt
eirsep: Let's begin with single query index to which all queries of a monitor are written. If different queries are to be written to different indices, users can always different monitors with the different custom query indices.

core/src/main/kotlin/org/opensearch/alerting/core/model/DataSources.kt
eirsep: Provides flexibility to add additional mapping properties including analyzers

core/src/main/resources/mappings/scheduled-jobs.json
eirsep: Pros :
1. We load all metadata and configs at once for the data sources. If kept as a separate object, we would have to read another index and logically link it back to the monitor pojo. 
2. We can re-use data sources POJO for more complicated scenarios such as a possible composite monitor wanting to override a monitor's behaviour by linking monitor id with its own desired data source object to override behaviour of the monitor.

eirsep:
@getsaurabh02 @lezzago Plz review the latest changes.

core/src/main/kotlin/org/opensearch/alerting/core/model/DataSources.kt
```diff
@@ -0,0 +1,130 @@
...
+     *  If index is pre-existing, mapping is updated*/
+    val findingsIndex: String? = null,
+
+    /** Configures a custom index to store alerts for a monitor. Creates a new index if index with given name not present.
+     *  If index is pre-existing, mapping is updated. */
+    val alertsIndex: String? = null,
+
+    /** Configures custom mappings by field type for query index.
+     * Custom query index mappings are configurable, only if a custom query index is configured too. */
+    val queryIndexMappingsByType: Map<String, Map<String, String>> = mapOf()
```
lezzago, line 32: If analyzer is the only type right now, we might want an enum set of what we support for query index mappings for when its open to other users and plugins. We might want a follow up item where with percolate queries, what other fields would make sense and restrict it to just those fields. When we open it up, to everyone, it might be a bad user experience to keep it open ended since it makes the usability harder for more novice users.

getsaurabh02: Agree. Plan to open this up as part of Composite Monitors should cover the specific validations around which fields should have what analyzer .

alerting/src/main/kotlin/org/opensearch/alerting/alerts/AlertIndices.kt
```diff
@@ -130,6 +131,23 @@ class AlertIndices(
         fun findingMapping() =
             AlertIndices::class.java.getResource("finding_mapping.json").readText()
 
+        @JvmStatic
+        fun getOrDefaultAlertIndex(monitor: Monitor): String {
+            var alertIndex = ALERT_INDEX
+            if (monitor.dataSources != null && monitor.dataSources.alertsIndex != null) {
```
getsaurabh02, line 137: Do we need to check ``isEmpty`` as well?

alerting/src/main/kotlin/org/opensearch/alerting/alerts/AlertIndices.kt
```diff
@@ -273,6 +313,21 @@ class AlertIndices(
...
+        }
+        val findingsIndex = monitor.dataSources.findingsIndex!!
+        if (!clusterService.state().routingTable().hasIndex(findingsIndex)) {
+            createIndex(
+                findingsIndex,
+                findingMapping()
+            )
+        } else {
+            updateIndexMapping(findingsIndex, findingMapping(), false)
+        }
```
getsaurabh02, line 328: Can we modify the original ``createOrUpdateInitialFindingHistoryIndex `` method to take the optional parameter ? To keep the create index logic at one place.

getsaurabh02 (approved):

getsaurabh02:
Thanks @eirsep Overall looks good. Few comments/suggestions to consider.

core/src/main/kotlin/org/opensearch/alerting/core/model/DataSources.kt
```diff
@@ -0,0 +1,130 @@
...
+import org.opensearch.common.io.stream.StreamOutput
+import org.opensearch.common.io.stream.Writeable
+import org.opensearch.common.xcontent.ToXContent
+import org.opensearch.common.xcontent.ToXContentObject
+import org.opensearch.common.xcontent.XContentBuilder
+import org.opensearch.common.xcontent.XContentParser
+import org.opensearch.common.xcontent.XContentParserUtils
+import java.io.IOException
+
+data class DataSources(
```
sbcd90, line 18: can this class move from `core` module to `alerting` module under the package `org.opensearch.alerting.model`.

there are certain advantages when we do this. e.g. in `current constructor` all the members `need to be set as null` like below:

```
data class DataSources(
    /** Configures a custom query index name for the monitor. Creates a new index if index with given name not present.*/
    val queryIndex: String? = null,

    /** Configures a custom index to store findings for a monitor. Creates a new index if index with given name not present.
     *  If index is pre-existing, mapping is updated*/
    val findingsIndex: String? = null,

    /** Configures a custom index to store alerts for a monitor. Creates a new index if index with given name not present.
     *  If index is pre-existing, mapping is updated. */
    val alertsIndex: String? = null,

    /** Configures custom mappings by field type for query index.
     * Custom query index mappings are configurable, only if a custom query index is configured too. */
    val queryIndexMappingsByType: Map<String, Map<String, String>> = mapOf()
```
but once we move it we can initialize them with their default values

```
data class DataSources(
    /** Configures a custom query index name for the monitor. Creates a new index if index with given name not present.*/
    val queryIndex: String = ScheduledJob.DOC_LEVEL_QUERIES_INDEX,

    /** Configures a custom index to store findings for a monitor. Creates a new index if index with given name not present.
     *  If index is pre-existing, mapping is updated*/
    val findingsIndex: String = FINDING_HISTORY_WRITE_INDEX,

    /** Configures a custom index to store alerts for a monitor. Creates a new index if index with given name not present.
     *  If index is pre-existing, mapping is updated. */
    val alertsIndex: String = ALERT_INDEX,

    /** Configures custom mappings by field type for query index.
     * Custom query index mappings are configurable, only if a custom query index is configured too. */
    val queryIndexMappingsByType: Map<String, Map<String, String>> = mapOf()
```

alerting/src/main/kotlin/org/opensearch/alerting/model/Monitor.kt
```diff
@@ -55,7 +56,8 @@ data class Monitor(
     val schemaVersion: Int = NO_SCHEMA_VERSION,
     val inputs: List<Input>,
     val triggers: List<Trigger>,
-    val uiMetadata: Map<String, Any>
+    val uiMetadata: Map<String, Any>,
+    val dataSources: DataSources? = null
```
sbcd90, line 60: the changes i suggested in `DataSources` will allow us to remove the null check here e.g. from

```
val dataSources: DataSources? = null
```
to
```
val dataSources: DataSources = DataSources()
```

alerting/src/main/kotlin/org/opensearch/alerting/alerts/AlertIndices.kt
sbcd90: with changes i suggested, now `monitor.dataSources` can never be null. so the `if condition` simplifies

alerting/src/main/kotlin/org/opensearch/alerting/alerts/AlertIndices.kt
```diff
@@ -130,6 +131,23 @@ class AlertIndices(
...
+
+        @JvmStatic
+        fun getOrDefaultFindingsHistoryIndex(monitor: Monitor): String {
+            if (monitor.dataSources?.findingsIndex == null || monitor.dataSources.findingsIndex!!.isEmpty()) {
+                return FINDING_HISTORY_WRITE_INDEX
+            }
+            return monitor.dataSources.findingsIndex!!
+        }
+
```
sbcd90, line 150: the `getOrDefault` methods will not be needed anymore as `monitor.dataSources` will either have `custom values` or have `default values`.

sbcd90 (commented):
hi @eirsep, thanks a lot for the pr. it looks good. i have a few comments which may reduce a lot of null checks & may reduce the need of passing monitor to several methods.

kindly have a look.

alerting/src/main/kotlin/org/opensearch/alerting/alerts/AlertIndices.kt
eirsep: in the default `createOrUpdateInitialFindingHistoryIndex` implementation they cache the state of index(i.e. whether it has been initialized or not) which differs from the implementation when a custom index name is passed. hence we can't have the logic for index creation in one place for findings and alerts indices.

core/src/main/kotlin/org/opensearch/alerting/core/model/DataSources.kt
eirsep: I had also thought of it but didn't explore further because `Monitor` model is in `core` module and would not be able to have `DataSources` model if the latter is present in `alerting` module due to cyclic dependency issues.

sbcd90 (approved):

PR #561, opened 2022-09-20
Adds support for fetching alerts from custom alert index in Get Alerts transport layer

eirsep:
Adds support for fetching alerts from custom alert index in Get Alerts transport layer

Related Issues:
#562 #522

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/action/GetAlertsRequest.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/action/GetAlertsRequest.kt
@@ -17,17 +17,20 @@ class GetAlertsRequest : ActionRequest {
     val severityLevel: String
     val alertState: String
     val monitorId: String?
+    val alertIndex: String?
 
     constructor(
         table: Table,
         severityLevel: String,
         alertState: String,
-        monitorId: String?
+        monitorId: String?,
+        alertIndex: String?,
     ) : super() {
         this.table = table
         this.severityLevel = severityLevel
         this.alertState = alertState
         this.monitorId = monitorId
+        this.alertIndex = alertIndex
     }
 
     @Throws(IOException::class)
@@ -35,7 +38,8 @@ class GetAlertsRequest : ActionRequest {
         table = Table.readFrom(sin),
         severityLevel = sin.readString(),
         alertState = sin.readString(),
-        monitorId = sin.readOptionalString()
+        monitorId = sin.readOptionalString(),
+        alertIndex = sin.readOptionalString()
     )
 
     override fun validate(): ActionRequestValidationException? {
@@ -48,5 +52,6 @@ class GetAlertsRequest : ActionRequest {
         out.writeString(severityLevel)
         out.writeString(alertState)
         out.writeOptionalString(monitorId)
+        out.writeOptionalString(alertIndex)
     }
 }
--- a/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
@@ -5,6 +5,10 @@
 
 package org.opensearch.alerting.transport
 
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import org.apache.logging.log4j.LogManager
 import org.opensearch.action.ActionListener
 import org.opensearch.action.search.SearchRequest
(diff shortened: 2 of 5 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/action/GetAlertsRequest.kt
```diff
@@ -17,25 +17,29 @@ class GetAlertsRequest : ActionRequest {
...
     val monitorId: String?
+    val alertIndex: String?
 
     constructor(
         table: Table,
         severityLevel: String,
         alertState: String,
-        monitorId: String?
+        monitorId: String?,
+        alertIndex: String?,
```
getsaurabh02, line 27: Do we need to add the version check for backward compatibility?

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
```diff
@@ -103,39 +107,75 @@ class TransportGetAlertsAction @Inject constructor(
...
+            alertIndex = getAlertsRequest.alertIndex
+            getAlerts(searchSourceBuilder, alertIndex, actionListener, user)
+        } else if (getAlertsRequest.monitorId != null) {
+            transportGetMonitorAction.execute(
+                GetMonitorRequest(
+                    getAlertsRequest.monitorId,
+                    -3L,
+                    RestRequest.Method.GET,
+                    FetchSourceContext.FETCH_SOURCE
+                ),
```
getsaurabh02, line 121: should we move this to a common util and try to re-use across places wherever we need to fetch the monitor config?

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
```diff
@@ -103,39 +107,75 @@ class TransportGetAlertsAction @Inject constructor(
...
+                        }
+                    }
+
+                    override fun onFailure(t: Exception) {
+                        actionListener.onFailure(t)
+                    }
+                }
+            )
+        } else {
+            getAlerts(searchSourceBuilder, alertIndex, actionListener, user)
```
getsaurabh02, line 136: ``getAlerts`` is being called 3 times here in this block. We should first resolve the index name and then make the final call.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
```diff
@@ -103,39 +107,75 @@ class TransportGetAlertsAction @Inject constructor(
...
+                object : ActionListener<GetMonitorResponse> {
+                    override fun onResponse(getMonitorResponse: GetMonitorResponse) {
+                        if (getMonitorResponse.monitor != null) {
+                            alertIndex = getMonitorResponse.monitor!!.dataSources.alertsIndex
+                            getAlerts(searchSourceBuilder, alertIndex, actionListener, user)
+                        }
+                    }
+
+                    override fun onFailure(t: Exception) {
+                        actionListener.onFailure(t)
```
getsaurabh02, line 131: During failure, should we rather log the message and fallback to default index name?

alerting/src/test/kotlin/org/opensearch/alerting/MonitorDataSourcesIT.kt
```diff
@@ -44,6 +47,24 @@ class MonitorDataSourcesIT : AlertingSingleNodeTestCase() {
...
+        try {
+            client()
+                .execute(GetAlertsAction.INSTANCE, GetAlertsRequest(table, "ALL", "ALL", null, "wrong_alert_index"))
+                .get()
+            fail()
+        } catch (e: Exception) {
+        }
     }
 
     fun `test execute monitor with custom alerts index`() {
```
getsaurabh02, line 70: Lets add test for failure scenarios, such as index not present, monitor config-resolution failures?

getsaurabh02 (changes requested):

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
eirsep: +1 on logging message. Will do that. This was existing code which was moved to a method hence overlooked its correctness.

But if the user passes an incorrect index name I don't think we should fallback to default as we might give them misleading info. They might assume we are surfacing alerts from the custom index name that they passed. It's better to throw an error which might make them evaluate what went wrong. Silent failure might be counterproductive.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
eirsep: let's get this PR in along with #563 and I will do that as a follow up.

alerting/src/main/kotlin/org/opensearch/alerting/action/GetAlertsRequest.kt
eirsep: created a task to track this.

eirsep: this question applies to changes that we have been making in the overall Datasources purview. WIll revisit during refactoring models to common utils

alerting/src/test/kotlin/org/opensearch/alerting/MonitorDataSourcesIT.kt
eirsep: have added check to verify index not found scenario.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
getsaurabh02: I am thinking about the stale config information being treated same everywhere. But yeah we can revisit this on the whole.

getsaurabh02 (approved):

sbcd90 (approved):

Issue #581, opened 2022-09-29
[BUG] Support deletion of queries from custom query index


eirsep:
Related to #522

When deleting a monitor, the corresponding query index should be cleaned up. With introduction of custom query index, we need to support deletion of queries from the appropriate query index by checking datasource config of monitor

getsaurabh02:
This is basically a two step cleanup process:

1. **Clean Query Documents from Query Index for Detector being deleted** : All the queries from different monitor which are  for the same log-type end up being part of the same Rule Index. Each query is persisted as a single document with ID as the key identifier of owner. Once the monitor is deleted, it should clean up the queries it added to the query index


2. **Cleanup the Query Index if there are no active monitors** : The last monitor which is cleaning up its queries should clean up the rule index, since it will be an empty index.

PR #584, opened 2022-10-03
support query deletion from custom query index on delete monitor

eirsep:

*Issue #, if available:*
#581

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteMonitorAction.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteMonitorAction.kt
@@ -78,24 +78,7 @@ class TransportDeleteMonitorAction @Inject constructor(
         private val user: User?,
         private val monitorId: String
     ) {
-
         fun resolveUserAndStart() {
-            if (user == null) {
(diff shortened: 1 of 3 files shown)


sbcd90:
@eirsep @getsaurabh02 one issue with the `alerting delete` api is it is asynchronous. when it returns, it doesnt guarantee that data & index is cleaned up or not.
do we want to fix it as part of this pr otherwise `security-analytics` may not be able to delete the `rule index`?

eirsep:
> @eirsep @getsaurabh02 one issue with the `alerting delete` api is it is asynchronous. when it returns, it doesnt guarantee that data & index is cleaned up or not. do we want to fix it as part of this pr otherwise `security-analytics` may not be able to delete the `rule index`?

We should not do asynchronous deletes IMO. We can leave around a lot of dangling resources if we don't do synchronous chaining.

getsaurabh02:
Lets open an issue to discuss the delete behavior for Alerting monitors and reason why was it made async in the first place. Can be out of the scope for this PR

getsaurabh02 (approved):

sbcd90 (approved):

eirsep:
https://github.com/opensearch-project/alerting/issues/586

Issue #588, opened 2022-10-05
[FEATURE]Support fetching bulk findings for multiple monitors in get findings API


eirsep:
Get findings API in Alerting plugin supports fetching findings for a given monitor id or fetches all findings present in an index.

Some use cases would require fetching findings for a given list of monitors rather than one.

In the current implementation this implies `n `calls to alerting plugin for `n` monitors which translates to `n` search requests.

We need to enhance the API to accept list of monitors and that way can boil it down to one search request.

eirsep:
@getsaurabh02 @lezzago @sbcd90

Issue #587, opened 2022-10-05
[FEATURE] Support fetching bulk alerts for multiple monitors in get alerts API 


eirsep:
Get alerts API in Alerting plugin supports fetching alerts for a given monitor id or fetches all alerts present in an index.

Some use cases would require fetching alerts for a given list of monitors rather than one.

In the current implementation this implies n calls to alerting plugin for n monitors which translates to n search requests.

We need to enhance the API to accept list of monitors and that way can boil it down to one search request.

eirsep:
The only caveat to keep in mind is currently we can only fetch from one index.

PR #590, opened 2022-10-06
Enhance Get Alerts and Get Findings for list of monitors in bulk

eirsep:
#587 #588


alerting/src/main/kotlin/org/opensearch/alerting/action/GetAlertsRequest.kt
```diff
@@ -11,26 +11,30 @@ import org.opensearch.alerting.model.Table
...
 import java.io.IOException
+import java.util.Collections
 
 class GetAlertsRequest : ActionRequest {
     val table: Table
     val severityLevel: String
     val alertState: String
     val monitorId: String?
     val alertIndex: String?
+    val monitorIds: List<String>?
```
lezzago, line 22: Can we not add this and make `monitorId` a list? At the rest layer, we can detect if its a string or a list and pass it in the request as such. This allows there to not be any confusion of the `monitorId` and `monitorIds` field

qreshi: Alternatively, the current way may be better if we want a deprecation path to change the parameter for the API. For example, in the beginning we'd support both `monitorId` and `monitorIds` just so there is no breaking changes but in a major version we can remove `monitorId` so that even a single monitor is passed through `monitorIds` (since making `monitorId` a list is confusing as it implies there is only one).

qreshi (commented):
Just some general comments:

1. What is the intended use case of this API?
2. Have we thought about the search limits of querying against multiple Monitors now (the regular search can return at most `10,000` search hits, will we support pagination to scale past this limit?)
3. It looks like custom Alert indices can be specified now, if these are created when they don't exist, are they made as system indices on the users behalf?
4. How would a GetAlerts search look like if a user wanted to search across alert indices (or pattern of indices) or is this even supported?

alerting/src/main/kotlin/org/opensearch/alerting/action/GetAlertsRequest.kt
eirsep: agreed will make the change.

eirsep:
> Just some general comments:
> 
> 1. **What is the intended use case of this API?**
We want to fetch alerts/findings for list of `n` monitors without `n` API calls or transport layer calls as each request to alerting will do one search request each i.e we are optimizing from 'n' search requests to one search request when we pass list of monitors in bulk.
We dont want ALL alerts that are present as for ex. security analytics knows the list of monitors it wants.

> 2. **Have we thought about the search limits of querying against multiple Monitors now (the regular search can return at most `10,000` search hits, will we support pagination to scale past this limit?)**
There already is a `size` param and `from` param. This API's performance doesn't have relevance to this change of support a list of monitor. we should tackle for that problem independently because the size and from params, irrespective of having a list of monitors, or not control how many alerts or findings are shown as this API effectively caters to GET ALL Alerts operation.

> 3. **It looks like custom Alert indices can be specified now, if these are created when they don't exist, are they made as system indices on the users behalf?**
User  cannot specify custom alerts or finding index in rest api request. That facility is only available for transport layer calls to be accessed by other plugins as of now.
> 4. **If the custom Alert indices are **not** system indices, what does the access control look like here? The default Alerting indices are system indices by design so users have to query them via the API. This means we don't apply user context when querying. If the custom indices are not system indices, could this API being used to escalate privileges?**
Same as above. User  cannot specify custom alerts or finding index in rest api request. That facility is only available for transport layer calls to be accessed by other plugins as of now.
> 5. How would a GetAlerts search look like if a user wanted to search across alert indices (or pattern of indices) or is this even supported?
Not supported as of now. we can iteratively enhance this API

qreshi:
@eirsep, given your answers for questions 3 and 4, I still think the access control question is relevant. If internal transport calls are being made to Get Alerts, we'd still want to inject user context for RBAC enabled cases so we filter the Alerts by backend roles. If this internal API can specify arbitrary indices, are we expecting the call to fail with 403s if it's non-system index that the injected user does not have access to? What if it's a system index that was specified, it won't be readable if user context is injected since no user can read them?

Also, I think today the search/get Alerts API can search across both active alerts and history indices (in case we want to get a full picture of current and completed alerts). How does this work when specifying custom indices? Can the history indices be custom as well?

eirsep:
**> @eirsep, given your answers for questions 3 and 4, I still think the access control question is relevant. If internal transport calls are being made to Get Alerts, we'd still want to inject user context for RBAC enabled cases so we filter the Alerts by backend roles. If this internal API can specify arbitrary indices, are we expecting the call to fail with 403s if it's non-system index that the injected user does not have access to? What if it's a system index that was specified, it won't be readable if user context is injected since no user can read them?**

If we take the use case of Security Analytics plugin(SAP) as an example, the plugin would do it's own RBAC on rule detectors  and once validated it would make a transport call to alerting as an admin user to perform alerting operations. User is oblivious to interactions between SAP and Alerting and hence doesn't need 403s thrown from alerting plugin for his SAP use case. my point being, plugins must handle their own security logic for RBAC enabled clusters and make transport calls as admin for get alerts/findings APIs. 

User cannot pass custom index name as of now. So we wouldn't have to worry about permissions on that index for RBAC enabled clusters

> Also, I think today the search/get Alerts API can search across both active alerts and history indices (in case we want to get a full picture of current and completed alerts). How does this work when specifying custom indices? Can the history indices be custom as well?

If custom indices are used, the owner of the custom index would have to manage history of alerts/findings. And if and when the plugin manages the findings/alerts history by itself, it can also pass index pattern in custom index field and achieve the desired results. Alerting plugin will not operate ISM policies for history on custom alerts/findings indices.

qreshi:
@eirsep

> my point being, plugins must handle their own security logic for RBAC enabled clusters and make transport calls as admin for get alerts/findings APIs.

This can't always happen. If the caller is injecting user context to run as user and the consuming action/REST API was run by some non-admin user, would you be escalating to an admin user when calling the Get Alerts API?

> If custom indices are used, the owner of the custom index would have to manage history of alerts/findings

I'd be careful about this from a user experience perspective. Also, Alerting is assuming today that the alerts and alert history indices are system indices. If we remove that assumption, we need to improve our error handling states because that means user can shoot themselves in the foot by corrupting that data. If we **do** maintain the system index requirement, then how can custom indices be validated to be system indices (and in this case, ISM can't managed them)?

sbcd90 (approved):

lezzago (approved):

PR #577, opened 2022-09-28
Recreate request object from writeable for Get alerts and get findings

eirsep:
Recreate request object from writeable for Get alerts and get findings
Signed-off-by: Surya Sashank Nistala <snistala@amazon.com>

PR #569, opened 2022-09-27
Use findings and alerts models, dtos from common utils

eirsep:
Signed-off-by: Surya Sashank Nistala <snistala@amazon.com>

PR #556, opened 2022-09-14
moving over data models from alerting to common-utils

sbcd90:
Signed-off-by: Subhobrata Dey <sbcd90@gmail.com>

*Description of changes:*
moving over data models from alerting to common-utils
Follow-up for this PR: https://github.com/opensearch-project/common-utils/issues/236

*CheckList:*
[X] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/alerting/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #592, opened 2022-10-07
Use alerting models and dtos from common-utils dependencies.

eirsep:
2.x PRs, whose changes are being applied to main.
https://github.com/opensearch-project/alerting/pull/577
https://github.com/opensearch-project/alerting/pull/569
https://github.com/opensearch-project/alerting/pull/568
https://github.com/opensearch-project/alerting/pull/558
https://github.com/opensearch-project/alerting/pull/556
 
**This PR also bumps Alerting plugin to 2.4**


sbcd90 (approved):

lezzago:
Can we cherry-pick the 2.x commits and make the changes accordingly

lezzago:
Can we reference the 2.x PRs that this is coming from?

eirsep:
> Can we reference the 2.x PRs that this is coming from?

@lezzago have updated PR description with the same

build.gradle
```diff
@@ -7,10 +7,10 @@ buildscript {
     apply from: 'build-tools/repositories.gradle'
 
     ext {
-        opensearch_version = System.getProperty("opensearch.version", "2.2.0-SNAPSHOT")
+        opensearch_version = System.getProperty("opensearch.version", "2.4.0-SNAPSHOT")
```
qreshi, line 10: Let's also note in the PR description that we're bumping `main` to `2.4.0` here if it's necessary for it to be in this PR. Ideally, we should separate it into another PR so it's easier to note when it happened in the commit history at a glance.

eirsep: It's necessary in this PR. Have updated the description as suggested. thanks for pointing it out.

getsaurabh02 (approved):

build.gradle
qreshi: Cool, spoke with @eirsep and we'll merge in https://github.com/opensearch-project/alerting/pull/596 and then this PR.

PR #622, opened 2022-10-25
fix alias exists check in findings index creation

eirsep:


getsaurabh02 (approved):

alerting/src/main/kotlin/org/opensearch/alerting/alerts/AlertIndices.kt
```diff
@@ -326,7 +326,7 @@ class AlertIndices(
         }
         val findingsIndex = dataSources.findingsIndex
         val findingsIndexPattern = dataSources.findingsIndexPattern ?: FINDING_HISTORY_INDEX_PATTERN
-        if (!clusterService.state().routingTable().hasIndex(findingsIndex)) {
+        if (!clusterService.state().metadata().hasAlias(findingsIndexPattern)) {
```
getsaurabh02, line 329: Should it be either of the index pattern or index name present?

sbcd90 (approved):

Issue #138, opened 2021-08-09
Provide ability to associate `backend role` and `security role` during monitor create


skkosuri-amzn:
**Is your feature request related to a problem? Please describe.**
Provide ability to associate `backend role` and `security role` during monitor create.

Currently, during monitor creation the user `backend roles` and `security roles` are resolved and associated with monitor. There is no option for the User to select which `backend role` and `security role` to get associated instead of all. 

**Describe the solution you'd like**
Provide optional field to add `backend role` and `security role` in the `POST _plugins/_alerting/monitors` API. When these optional fields are provided, and these are present in resolved roles - use provided roles to associate with monitor.

ryn9:
@skkosuri-amzn is this expected to make its way into the 1.3 release?

brijos:
@ryn9, this isn't gong to make it into 1.3.  We will update this issue when we know more about which release this will fall into.

ryn9:
@brijos . thank you for letting me know. Please let me know if there is any additional information / context I can aide with.

sevenval-admins:
Hi, any news here?

ryn9:
@sevenval-admins - if you and folks you know are looking for this feature - please make sure they give is a thumbs up at the top.

qreshi:
@lezzago Can we close this based on the PR linked above being merged in or is there any additional work to be done?

lezzago:
We have updated the [documentation](https://opensearch.org/docs/latest/monitoring-plugins/alerting/api/#create-a-query-level-monitor) and starting from 2.4, we support allowing to specify backend roles.

PR #635, opened 2022-11-02
Support to specify backend roles for monitors

lezzago:

*Issue #, if available:*
#138 

*Description of changes:*
Support specify backend roles for monitors
New Scenarios:
- Create monitor scenario:
  - User specified backend roles
    - Admin user: use all the specified backend roles to associate to the monitor
    - Normal user: use all the specified backend roles (that the user already has in their list of backend roles) to associate to the monitor
  - User did not specify backend roles:
    - Copy user’s backend roles and associate them to the monitor
 

- Update monitor scenario:
  - User specified backend roles
    - Admin user: Remove all the backend roles associate to the monitor and then use all the specified backend roles to associate to the monitor
    - Normal user: Remove backend roles associated to the monitor that the user has access to, but didn’t specify. Then add all the other specified backend roles (that the user already has in their list of backend roles) to the monitor
  - User did not specify backend roles:
    - Don’t update the backend roles on the monitor

- Empty list is considered as removing all permissions the user has access to or create a monitor with no backend roles
- Throw exceptions if a user tries to associate roles that they do not have access to
- Only admin users can pass in an empty list. Non-admin users need to pass in at least 1 backend role so they dont remove their access from the monitor for mistakes

Example of new create/update monitor request (**`rbac_roles`** is the new optional parameter):
```
POST _plugins/_alerting/monitors
{
  "type": "monitor",
  "name": "test-monitor",
  "monitor_type": "query_level_monitor",
  "enabled": true,
  "schedule": {
    "period": {
      "interval": 1,
      "unit": "MINUTES"
    }
  },
  "inputs": [{
    "search": {
      "indices": ["movies"],
      "query": {
        "size": 0,
        "aggregations": {},
        "query": {
          "bool": {
            "filter": {
              "range": {
                "@timestamp": {
                  "gte": "{{period_end}}||-1h",
                  "lte": "{{period_end}}",
                  "format": "epoch_millis"
                }
              }
            }
          }
        }
      }
    }
  }],
  "triggers": [{
    "name": "test-trigger",
    "severity": "1",
    "condition": {
      "script": {
        "source": "ctx.results[0].hits.total.value > 0",
        "lang": "painless"
      }
    },
    "actions": [{
      "name": "test-action",
      "destination_id": "ld7912sBlQ5JUWWFThoW",
      "message_template": {
        "source": "This is my message body."
      },
      "throttle_enabled": true,
      "throttle": {
        "value": 27,
        "unit": "MINUTES"
      },
      "subject_template": {
        "source": "TheSubject"
      }
    }]
  }],
  "rbac_roles": ["role1", "role2"]
}
```

*CheckList:*

### DIFF
--- a/.github/workflows/security-test-workflow.yml
+++ b/.github/workflows/security-test-workflow.yml
@@ -43,7 +43,12 @@ jobs:
           plugin_version=`echo $plugin|awk -F- '{print $3}'| cut -d. -f 1-4`
           qualifier=`echo $plugin|awk -F- '{print $4}'| cut -d. -f 1-1`
           candidate_version=`echo $plugin|awk -F- '{print $5}'| cut -d. -f 1-1`
-          docker_version=$version
+          if qualifier
+          then
+            docker_version=$version-$qualifier
+          else
+            docker_version=$version
+          fi
 
           [[ -z $candidate_version ]] && candidate_version=$qualifier && qualifier=""
 
--- a/.github/workflows/security-test-workflow.yml
+++ b/.github/workflows/security-test-workflow.yml
@@ -43,7 +43,7 @@ jobs:
           plugin_version=`echo $plugin|awk -F- '{print $3}'| cut -d. -f 1-4`
           qualifier=`echo $plugin|awk -F- '{print $4}'| cut -d. -f 1-1`
           candidate_version=`echo $plugin|awk -F- '{print $5}'| cut -d. -f 1-1`
-          docker_version=$version-$qualifier
+          docker_version=$version
 
           [[ -z $candidate_version ]] && candidate_version=$qualifier && qualifier=""
 
--- a/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
@@ -403,11 +403,15 @@ class TransportIndexMonitorAction @Inject constructor(
             if (user != null) {
                 // Use the backend roles which is an intersection of the requested backend roles and the user's backend roles.
                 // Admins can pass in any backend role. Also if no backend role is passed in, all the user's backend roles are used.
-                val rbacRoles = if (request.rbacRoles.isNullOrEmpty())
+                val rbacRoles = if (request.rbacRoles.isNullOrEmpty()) {
                     user.backendRoles.toSet()
-                else if (!isAdmin(user)) request.rbacRoles!!.intersect(user.backendRoles).toSet()
-                else request.rbacRoles!!
+                } else if (!isAdmin(user)) {
+                    request.rbacRoles!!.intersect(user.backendRoles).toSet()
+                } else {
+                    request.rbacRoles!!
+                }
                 request.monitor = request.monitor.copy(user = User(user.name, rbacRoles.toList(), user.roles, user.customAttNames))
+                log.debug("Created monitor's backend roles: $rbacRoles")
             }
 
             val indexRequest = IndexRequest(SCHEDULED_JOBS_INDEX)
@@ -512,9 +516,11 @@ class TransportIndexMonitorAction @Inject constructor(
                             user = User(user.name, request.rbacRoles, user.roles, user.customAttNames)
                         )
                     } else {
-                        val rolesToRemove = user.backendRoles.filter { !request.rbacRoles!!.contains(it) }
-                        val updatedRbac = currentMonitor.user!!.backendRoles.filter { !rolesToRemove.contains(it) }.toMutableSet()
-                        updatedRbac.addAll(request.monitor.user!!.backendRoles)
+                        val rolesToRemove = user.backendRoles
+                        rolesToRemove.removeAll(request.rbacRoles!!)
+                        val updatedRbac = currentMonitor.user!!.backendRoles
+                        updatedRbac.removeAll(rolesToRemove)
+                        updatedRbac.addAll(request.rbacRoles!!)
                         request.monitor = request.monitor.copy(
                             user = User(user.name, updatedRbac.toList(), user.roles, user.customAttNames)
                         )
@@ -523,6 +529,7 @@ class TransportIndexMonitorAction @Inject constructor(
                     request.monitor = request.monitor
                         .copy(user = User(user.name, currentMonitor.user!!.backendRoles, user.roles, user.customAttNames))
                 }
+                log.debug("Update monitor backend roles to: ${request.monitor.user?.backendRoles}")
             }
 
             request.monitor = request.monitor.copy(schemaVersion = IndexUtils.scheduledJobIndexSchemaVersion)
--- a/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
@@ -541,7 +541,20 @@ class TransportIndexMonitorAction @Inject constructor(
             if (request.monitor.enabled && currentMonitor.enabled)
                 request.monitor = request.monitor.copy(enabledTime = currentMonitor.enabledTime)
 
-            // On update monitor check which backend roles to associate to the monitor
+            /**
+             * On update monitor check which backend roles to associate to the monitor.
+             * Below are 2 examples of how the logic works
+             *
+             * Example 1, say we have a Monitor with backend roles [a, b, c, d] associated with it.
+             * If I'm User A (non-admin user) and I have backend roles [a, b, c] associated with me and I make a request to update
+             * the Monitor's backend roles to [a, b]. This would mean that the roles to remove are [c] and the roles to add are [a, b].
+             * The Monitor's backend roles would then be [a, b, d].
+             *
+             * Example 2, say we have a Monitor with backend roles [a, b, c, d] associated with it.
+             * If I'm User A (admin user) and I have backend roles [a, b, c] associated with me and I make a request to update
+             * the Monitor's backend roles to [a, b]. This would mean that the roles to remove are [c, d] and the roles to add are [a, b].
+             * The Monitor's backend roles would then be [a, b].
+             */
             if (user != null) {
                 if (request.rbacRoles != null) {
                     if (isAdmin(user)) {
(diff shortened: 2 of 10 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -400,6 +400,16 @@ class TransportIndexMonitorAction @Inject constructor(
         private suspend fun indexMonitor() {
             var metadata = createMetadata()
 
+            if (user != null) {
+                // Use the backend roles which is an intersection of the requested backend roles and the user's backend roles.
+                // Admins can pass in any backend role. Also if no backend role is passed in, all the user's backend roles are used.
+                val rbacRoles = if (request.rbacRoles.isNullOrEmpty())
+                    user.backendRoles.toSet()
+                else if (!isAdmin(user)) request.rbacRoles!!.intersect(user.backendRoles).toSet()
+                else request.rbacRoles!!
```
qreshi, line 409: I know we're doing an `isNullOrEmpty()` check for the `if` here but we should try to avoid `!!` in production as much as possible. If empty `request.rbacRoles` are not logically different in how they're evaluated vs `null`, we should just default the argument to an empty list instead and not make it nullable.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -494,6 +504,27 @@ class TransportIndexMonitorAction @Inject constructor(
             if (request.monitor.enabled && currentMonitor.enabled)
                 request.monitor = request.monitor.copy(enabledTime = currentMonitor.enabledTime)
 
+            // On update monitor check which backend roles to associate to the monitor
+            if (user != null) {
```
qreshi, line 508: Could you add tests for these scenarios?

lezzago: yes

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
lezzago: I have fixed it

.github/workflows/security-test-workflow.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           plugin_version=`echo $plugin|awk -F- '{print $3}'| cut -d. -f 1-4`
           qualifier=`echo $plugin|awk -F- '{print $4}'| cut -d. -f 1-1`
           candidate_version=`echo $plugin|awk -F- '{print $5}'| cut -d. -f 1-1`
-          docker_version=$version-$qualifier
+          docker_version=$version
```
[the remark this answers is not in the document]
lezzago: Added an if statement to keep it

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -400,6 +400,19 @@ class TransportIndexMonitorAction @Inject constructor(
         private suspend fun indexMonitor() {
             var metadata = createMetadata()
 
+            if (user != null) {
+                // Use the backend roles which is an intersection of the requested backend roles and the user's backend roles.
+                // Admins can pass in any backend role. Also if no backend role is passed in, all the user's backend roles are used.
+                val rbacRoles = if (request.rbacRoles.isNullOrEmpty()) user.backendRoles.toSet()
+                else if (!isAdmin(user)) request.rbacRoles?.intersect(user.backendRoles)?.toSet()
+                else request.rbacRoles
+
```
rishabhmaurya, line 409: we should throw an error if `rbacRoles.length != request.rbacRoles` if `request.rbacRoles` is not null. This is valid for second case.

If  `request.rbacRoles` is not null but empty then for non-admins we should keep the `rbacRoles` empty.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -494,6 +507,30 @@ class TransportIndexMonitorAction @Inject constructor(
             if (request.monitor.enabled && currentMonitor.enabled)
                 request.monitor = request.monitor.copy(enabledTime = currentMonitor.enabledTime)
 
+            // On update monitor check which backend roles to associate to the monitor
+            if (user != null) {
+                if (!request.rbacRoles.isNullOrEmpty()) {
```
rishabhmaurya, line 512: Lets treat null and empty cases separately - 
Empty case is an intentional behavior and should be treated as if they want to remove all roles from a monitor which they can assume. 
Lets also add unit test for it.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
qreshi: If a non-admin can expunge all roles from their own Monitors, that means only the admin can fix it. Should we allow for that? It might cause noise for the admin due to user error. What would be the use case to lose access to something for a non-admin user?

rishabhmaurya (approved):

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
lezzago: separated the cases

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
qreshi: I see where you're coming from but non-admin removing all users they have access to will essentially lock away the object from them in RBAC cases unless an admin gives them access again. What if a user uses the empty list and causes everyone to lose access to it since they removed backend roles that were shared by everyone? It seems like a mechanic only admin should have. I'm just concerned about the operational implications it can have to users who don't understand its implications.

rishabhmaurya (approved):

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -494,6 +541,31 @@ class TransportIndexMonitorAction @Inject constructor(
...
+                        request.monitor = request.monitor.copy(
+                            user = User(user.name, request.rbacRoles, user.roles, user.customAttNames)
+                        )
+                    } else {
+                        val rolesToRemove = user.backendRoles
+                        request.rbacRoles.orEmpty().let { rolesToRemove.removeAll(it) }
+                        val rolesToAdd = request.rbacRoles?.intersect(user.backendRoles)
+                        val updatedRbac = currentMonitor.user?.backendRoles
+                        updatedRbac?.removeAll(rolesToRemove)
+                        rolesToAdd.orEmpty().let { updatedRbac?.addAll(it) }
```
qreshi, line 557: This is a little hard to read. Am I understand right that we're essentially:

1. Removing the backend roles (currently on the Monitor) which are associated with the calling user **but not** mentioned in the request
2. We're adding the backend roles that are both in the request **and** the calling user has associated as well

So, for example, say we have a Monitor with backend roles `[a, b, c, d]` associated with it.

If I'm User A and I have backend roles `[a, b, c]` associated with me and I make a request to update the Monitor's backend roles to `[a, b]`. This would mean that:

1. The roles to remove are `[c]`
2. The roles to add are `[a, b]`

I'm assuming this is so that the role `[d]` which I never had associated to me, doesn't get removed when I update the Monitor? Is this the behavior that we want? The `admin` wouldn't necessarily have all backend roles associated with them, wouldn't that mean they can't definitely reduce the associated roles in that case?

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -494,6 +541,31 @@ class TransportIndexMonitorAction @Inject constructor(
...
 
+            // On update monitor check which backend roles to associate to the monitor
+            if (user != null) {
+                if (request.rbacRoles != null) {
+                    if (isAdmin(user)) {
+                        request.monitor = request.monitor.copy(
+                            user = User(user.name, request.rbacRoles, user.roles, user.customAttNames)
+                        )
+                    } else {
+                        val rolesToRemove = user.backendRoles
```
qreshi, line 552: I think we could benefit from using set subtracting operations and combining calls to make this a little easier to read:

```
val rolesToRemove = user.backendRoles - request.rbacRoles.orEmpty()
...
val updatedRbac = currentMonitor.user?.backendRoles
    ?.removeAll(rolesToRemove)
    ?.addAll(rolesToAdd.orEmpty())
```

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -494,6 +541,31 @@ class TransportIndexMonitorAction @Inject constructor(
...
+                        request.rbacRoles.orEmpty().let { rolesToRemove.removeAll(it) }
+                        val rolesToAdd = request.rbacRoles?.intersect(user.backendRoles)
```
qreshi, line 554: Is it possible for there to be roles that are in `request.rbacRoles` that aren't in `user.backendRoles`? We do a validation early on that throws an error if the request has backend roles the user doesn't have permissions to. In that case, could we just use `request.rbacRoles`?

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
lezzago: We want the behavior such that role `[d]` is still there.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
lezzago: cleaned up

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
lezzago: added comments in code for this

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
lezzago: cleaned it up

qreshi (approved):

PR #636, opened 2022-11-03
Adds findings in bucket level monitor 

eirsep:
This change surfaces out the list of documents that become a part of bucket level monitor aggregation buckets as findings for the monitor.

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/BucketLevelMonitorRunner.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/BucketLevelMonitorRunner.kt
@@ -6,8 +6,9 @@
 package org.opensearch.alerting
 
 import org.apache.logging.log4j.LogManager
+import org.opensearch.action.bulk.BulkRequest
+import org.opensearch.action.bulk.BulkResponse
 import org.opensearch.action.index.IndexRequest
-import org.opensearch.action.index.IndexResponse
 import org.opensearch.action.search.SearchRequest
 import org.opensearch.action.search.SearchResponse
 import org.opensearch.action.support.WriteRequest
@@ -16,6 +17,7 @@ import org.opensearch.alerting.model.BucketLevelTriggerRunResult
 import org.opensearch.alerting.model.InputRunResults
 import org.opensearch.alerting.model.MonitorRunResult
 import org.opensearch.alerting.opensearchapi.InjectorContextElement
+import org.opensearch.alerting.opensearchapi.retry
 import org.opensearch.alerting.opensearchapi.suspendUntil
 import org.opensearch.alerting.opensearchapi.withClosableContext
 import org.opensearch.alerting.script.BucketLevelTriggerExecutionContext
@@ -23,7 +25,6 @@ import org.opensearch.alerting.util.defaultToPerExecutionAction
 import org.opensearch.alerting.util.getActionExecutionPolicy
 import org.opensearch.alerting.util.getBucketKeysHash
 import org.opensearch.alerting.util.getCombinedTriggerRunResult
-import org.opensearch.client.Client
 import org.opensearch.common.xcontent.LoggingDeprecationHandler
 import org.opensearch.common.xcontent.ToXContent
 import org.opensearch.common.xcontent.XContentBuilder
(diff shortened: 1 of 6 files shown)


sbcd90 (approved):

alerting/src/main/kotlin/org/opensearch/alerting/BucketLevelMonitorRunner.kt
```diff
@@ -287,6 +319,105 @@ object BucketLevelMonitorRunner : MonitorRunner() {
...
+                val bucketValues: Set<String> = triggerResult.aggregationResultBuckets.keys
+                val query = input.query
+                var fieldName = ""
+                var grouByFields = 0 // if number of fields used to group by > 1 we won't calculate findings
+                for (aggFactory in (query.aggregations() as AggregatorFactories.Builder).aggregatorFactories) {
+                    val sources = (aggFactory as CompositeAggregationBuilder).sources()
+                    for (source in sources) {
+                        if (grouByFields > 0) {
+                            return listOf()
+                        }
```
getsaurabh02, line 341: Why do we need to return from here?

alerting/src/main/kotlin/org/opensearch/alerting/BucketLevelMonitorRunner.kt
```diff
@@ -287,6 +319,105 @@ object BucketLevelMonitorRunner : MonitorRunner() {
...
+                logger.debug("Findings: $findingStr")
+                if (shouldCreateFinding) {
+                    val indexRequest = IndexRequest(monitor.dataSources.findingsIndex)
+                        .setRefreshPolicy(WriteRequest.RefreshPolicy.IMMEDIATE)
+                        .source(findingStr, XContentType.JSON)
+                        .id(finding.id)
+                        .routing(finding.id)
+
+                    monitorCtx.client!!.suspendUntil<Client, IndexResponse> {
+                        monitorCtx.client!!.index(indexRequest, it)
```
getsaurabh02, line 411: Are we indexing each finding document, one at a time? Wondering if we should use bulk request instead?

alerting/src/main/kotlin/org/opensearch/alerting/BucketLevelMonitorRunner.kt
```diff
@@ -287,6 +319,105 @@ object BucketLevelMonitorRunner : MonitorRunner() {
...
+                    }
+                }
+
+                findings.add(finding.id)
+            }
+        }
+        return findings
```
getsaurabh02, line 418: Do we need to return the findings from here, as indexing is already done?

alerting/src/test/kotlin/org/opensearch/alerting/MonitorRunnerServiceIT.kt
```diff
@@ -1321,6 +1323,71 @@ class MonitorRunnerServiceIT : AlertingRestTestCase() {
         )
     }
 
+    fun `test bucket-level monitor with findings enabled`() {
```
getsaurabh02, line 1326: Lets validate other scenarios as well:
1. Where findings will not be created - such as compound aggregation.
2. Findings index unavailable, or exception in index.

getsaurabh02 (commented):
Thanks @eirsep . Overall approach looks good to me. Few comments/clarifications to consider.

alerting/src/main/kotlin/org/opensearch/alerting/BucketLevelMonitorRunner.kt
eirsep: If there are more groupby fields then aggregation buckets would become tuples. We can develop iteratively and first begin by supporting only for a single field(buckets with single field aggregated on) to filter on.

alerting/src/main/kotlin/org/opensearch/alerting/BucketLevelMonitorRunner.kt
eirsep: we are returning finding ids. finding ids list is populated in the alert.

sbcd90 (approved):

getsaurabh02 (approved):

PR #670, opened 2022-11-09
Set message when building LegacyCustomWebhookMessage

partlov:
Signed-off-by: Petar Partlov <partlov@gmail.com>

*Issue #, if available:*

*Description of changes:*
While testing legacy destinations for custom webhooks I discovered issue which is 100% reproducible. Scope should not be huge as this can influence only users with some issue during migration of old legacy notification system to new which uses notification plugin.

Issue is that one of required field (message) was not set during building instance of LegacyCustomWebhookMessage object.

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/model/destination/Destination.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/model/destination/Destination.kt
@@ -260,7 +260,7 @@ data class Destination(
             }
             DestinationType.CUSTOM_WEBHOOK -> {
                 destinationMessage = LegacyCustomWebhookMessage.Builder(name)
-                    .withUrl(getLegacyCustomWebhookMessageURL(customWebhook))
+                    .withUrl(getLegacyCustomWebhookMessageURL(customWebhook, compiledMessage))
                     .withHeaderParams(customWebhook?.headerParams)
                     .withMessage(compiledMessage).build()
             }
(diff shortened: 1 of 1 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/model/destination/Destination.kt
```diff
@@ -260,7 +260,7 @@ data class Destination(
             }
             DestinationType.CUSTOM_WEBHOOK -> {
                 destinationMessage = LegacyCustomWebhookMessage.Builder(name)
-                    .withUrl(getLegacyCustomWebhookMessageURL(customWebhook))
+                    .withUrl(getLegacyCustomWebhookMessageURL(customWebhook, compiledMessage))
                     .withHeaderParams(customWebhook?.headerParams)
                     .withMessage(compiledMessage).build()
```
qreshi, line 265: The message is passed in here. It's not required in `getLegacyCustomWebhookMessageURL` since that method is only combining the individual URL parameters to a String to pass to `withUrl`.

partlov: Yes that is correct. It is true that it is not required for actual construction of URI, but unfortunately it is required field of `LegacyCustomWebhookMessage` and it always throws `IllegalArgumentException("Message content is missing")`. I went with this solution as it is simple and this legacy destination is only used in limited number of cases (and it will probably be removed in some of future releases).

It would probably be best to completely separate URI construction code of constructing `LegacyCustomWebhookMessage` object itself, but as message is available here I went with this easy fix.

lezzago (approved):

qreshi (approved):

Issue #731, opened 2022-12-13
[Bug] Support overriding defaultTitle for Notifications send via Alerting


AWSHurneyt:
**Is your feature request related to a problem?**
When the migration of destinations to the Notifications plugin that happened in OS 2.0, the `subject` line of migrated Alerting trigger action messages have been replaced with the `defaultTitle` "Alerting-Notification Action" when using SNS to send the emails. The old `subject` line then appears in the message `body`.

**What solution would you like?**
Ideally, if a `subject` line is populated, Notifications should use that for the email subject instead of the `defaultTitle`.

**Workarounds**
1. Users can setup an `SES Notification Channel` which takes in the `title` as the `subject`.

qreshi:
Transferred this issue to the Alerting plugin repo as that's where the defaultTitle override needed to occur.

qreshi:
Closing issue as the change has been merged in.

Issue #529, opened 2022-08-24
[BUG] Slack alerting notifications regression


sas1024:
**What is the bug?**
In OpenSearch 1.2.x all Slack notifications was formatted as `$subject \n\n $message`. And all worked very fine.
In OpenSearch 2.2.0 all Slack notifications formatted as `Alerting-Notification Action \n\n $subject \n\n $message`. And for now all slack notifications with alerts bad readable because of prefix "Alerting-Notification Action".

**How can one reproduce the bug?**
Just create any alert with slack notification.

**What is the expected behavior?**
Please, return old formatting for slack notifications. `$subject \n\n $message`, without hardcoded prefix "Alerting-Notification Action".

**Do you have any additional context?**
As i see, hardcoded prefix used [here](https://github.com/opensearch-project/alerting/blob/main/alerting/src/main/kotlin/org/opensearch/alerting/util/destinationmigration/NotificationApiUtils.kt#L144).

praveensameneni:
Thank you for finding the issue, looking into it.

uroskarama:
@praveensameneni I have created a pull request with a "hacky" solution to this issue by passing a blank space, since from my investigation the notification plugin does't send a notification if null or empty string to be passed (see [here](https://github.com/opensearch-project/notifications/blob/main/notifications/core-spi/src/main/kotlin/org/opensearch/notifications/spi/model/MessageContent.kt#L25) )
I confirmed by manual testing that it works properly and sends slack messages without this "Alerting-Notification Action".
About integration tests, we don't have any in alerting testing content notifications from what I can see? It is possible to create a socket server to pass notifications to it and then assert them as part of integration test, would that be an acceptable solution?

aviramaz:
+1,
Seems like a wider issue.
@praveensameneni - This issue is also impacting SNS, when triggering to SNS, the email subject is no longer the correct one but the "Alerting-Notification Action" is being received, started after upgrading to 2.x from 1.x.
We can no longer distinguish between the alerts due to it by the Subject.

lezzago:
This issue has been fixed with the PR

PR #750, opened 2023-01-12
Fix setting default title only when no subject has been set

lezzago:

*Issue #, if available:*
#529, #731
*Description of changes:*
Set default title only when there is no subject passed.

action message configuration example:
action message configuration

Current message look in chime:
old notification look

Before Notification plugin introduction in chime:
before 2 0 behavior

With the fix look in chime:
updated message look

With the fix for SNS:
Screen Shot 2023-02-09 at 11 14 16 AM

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/util/destinationmigration/NotificationApiUtils.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/util/destinationmigration/NotificationApiUtils.kt
@@ -14,7 +14,6 @@ import org.opensearch.alerting.opensearchapi.retryForNotification
 import org.opensearch.alerting.opensearchapi.suspendUntil
 import org.opensearch.client.Client
 import org.opensearch.client.node.NodeClient
-import org.opensearch.common.Strings
 import org.opensearch.common.unit.TimeValue
 import org.opensearch.commons.ConfigConstants
 import org.opensearch.commons.destination.message.LegacyBaseMessage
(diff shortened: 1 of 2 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/util/destinationmigration/NotificationApiUtils.kt
```diff
@@ -138,33 +136,11 @@ suspend fun NotificationConfigInfo.sendNotification(client: Client, title: Strin
...
-    return constructMessageContent(subject, message)
-}
-
-/**
- * Similar to Destinations, this is a generic utility method for constructing message content from
- * a subject and message body when sending through Notifications since the Action definition in Monitors can have both.
- */
-private fun constructMessageContent(subject: String?, message: String): String {
-    return if (Strings.isNullOrEmpty(subject)) message else "$subject \n\n $message"
+    return if (subject.isNullOrEmpty()) defaultTitle else subject
```
qreshi, line 143: Title and subject are not necessarily interchangeable right? Also it seems we've now removed the `constructMessageContent` method, is this being replaced by a similar call somewhere else?

lezzago: I was looking into the behavior on Notification and subject is like the title being used here and I think it was a bug for alerting to think that subject wouldn't equal title. This goes back to how Alerting behaved before 2.0. The Notification plugin tried to be the same as Alerting's notification sub project, so the title there behaves similarly to subject.

eirsep (approved):

rishabhmaurya (approved):

PR #768, opened 2023-01-27
fix for ERROR alert state generation in doc-level monitors

sbcd90:
Signed-off-by: Subhobrata Dey <sbcd90@gmail.com>

*Issue #, if available:*

*Description of changes:*
fix for ERROR alert state generation in doc-level monitors

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
@@ -74,25 +74,17 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
         } catch (e: Exception) {
             val id = if (monitor.id.trim().isEmpty()) "_na_" else monitor.id
             logger.error("Error setting up alerts and findings indices for monitor: $id", e)
-            return monitorResult.copy(error = AlertingException.wrap(e))
+            monitorResult = monitorResult.copy(error = AlertingException.wrap(e))
         }
 
         try {
             validate(monitor)
         } catch (e: Exception) {
             logger.error("Failed to start Document-level-monitor. Error: ${e.message}")
-            return monitorResult.copy(error = AlertingException.wrap(e))
+            monitorResult = monitorResult.copy(error = AlertingException.wrap(e))
         }
 
         var (monitorMetadata, _) = MonitorMetadataService.getOrCreateMetadata(monitor, createWithRunContext = false)
-
-        monitorCtx.docLevelMonitorQueries!!.initDocLevelQueryIndex(monitor.dataSources)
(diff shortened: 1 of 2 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -265,6 +264,16 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
             alerts.add(alert)
         }
 
+        if (findingDocPairs.isEmpty() && monitorResult.error != null) {
```
eirsep, line 267: why are we checking if `findingDocPairs.isEmpty()` ?
What happens is findings is not empty and there is monitorResult error?

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -109,6 +101,14 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
         val docsToQueries = mutableMapOf<String, MutableList<String>>()
 
         try {
+            monitorCtx.docLevelMonitorQueries!!.initDocLevelQueryIndex(monitor.dataSources)
```
eirsep, line 104: why did we shift this snippet?

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
sbcd90: if `findingDocPairs` is not empty, an `alert` will still be generated with `ERROR` state due to the logic here https://github.com/opensearch-project/alerting/blob/2ecde357581a968195d1a804c601361743084688/alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt#L257-#L265.

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
sbcd90: the method `monitorCtx.docLevelMonitorQueries!!.indexDocLevelQueries`  assumes that the `customer provided log index` always exist otherwise it throws an `Unhandled error`.

https://github.com/opensearch-project/alerting/blob/2ecde357581a968195d1a804c601361743084688/alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt#L105

So, moved it into the `try` block, so that we always get an `alert` with `ERROR` state for any `unhandled errors`.

eirsep (approved):

rishabhmaurya (approved):

PR #801, opened 2023-02-17
Added test for changes for bug #78

dankyalo599:
*Issue #, if available:*
bug alert trigger #78

*Description of changes:*
 The previouslyTriggered set is being used to keep track of which query triggers have already been fired during the current execution of the program.

the 'contains' method is used to check whether a trigger with a given name has already been fired. If it has been fired previously, the method returns a QueryLevelTriggerRunResult object indicating that the trigger was not run again (false), and the null value indicates that no data was returned.

By keeping track of which triggers have already been fired during the current execution, the previouslyTriggered set ensures that each trigger is only fired once during a single execution of the program. This helps to prevent unnecessary or duplicate operations that could potentially impact the performance and efficiency of the program.

*CheckList:*
[X ] Commits are signed per the DCO using --signoff

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/TriggerService.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/TriggerService.kt
@@ -35,6 +35,7 @@ class TriggerService(val scriptService: ScriptService) {
     private val logger = LogManager.getLogger(TriggerService::class.java)
     private val ALWAYS_RUN = Script("return true")
     private val NEVER_RUN = Script("return false")
+    private val previouslyTriggered = mutableSetOf<String>()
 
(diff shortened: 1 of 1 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/TriggerService.kt
```diff
@@ -35,6 +35,7 @@ class TriggerService(val scriptService: ScriptService) {
     private val logger = LogManager.getLogger(TriggerService::class.java)
     private val ALWAYS_RUN = Script("return true")
     private val NEVER_RUN = Script("return false")
+    private val previouslyTriggered = mutableSetOf<String>()
```
petardz, line 38: you don't insert values in this set ever?

dblock:
@dankyalo599 Want to finish this?

[Catch All Triage - [1](https://github.com/dblock), [2](https://github.com/reta), [3](https://github.com/krisfreedain), [4](https://github.com/OVI3D0)]

Issue #841, opened 2023-03-30
[BUG] When upgrading a 2.0-2.4 domain to 2.5+ domain, there is a monitor creation/update error


lezzago:
**What is the bug?**
The [source_to_query_index_mapping](https://github.com/opensearch-project/alerting/blame/2.5/core/src/main/resources/mappings/scheduled-jobs.json#L516) attribute was introduced in 2.5. If the cluster has been upgraded with existing document level monitors, those monitors would execute and would end up updating the alerting config index mapping in the `source_to_query_index_mapping` field, but would not include the `enabled` attribute. When the user tries to update or create a monitor, the plugin will try to update the alerting config index mapping, but would fail due to the `source_to_query_index_mapping` field already existing and not having the `enabled` field. 

**How can one reproduce the bug?**
1. Create a OpenSearch cluster in 2.0-2.4 and create document level monitors that are running. 
2. Upgrade cluster to 2.5+
3. Attempt to create/update a monitor.

Issue #848, opened 2023-04-03
[FEATURE] Add upgrade tests fopr multiple versions


lezzago:
**Is your feature request related to a problem?**
The plugin needs to have upgrade tests from the last previous versions to the current versions and ensure after monitors are running, the APIs and new monitor executions work without problems.

This would have caught this issue: #841 

**What solution would you like?**
Have the upgrade tests create all monitor types and have them executing. Then perform upgrades to the current version to see the behavior of the plugin is still working correctly. This should be upgrading from all the previous minor versions.

PR #849, opened 2023-04-04
Update config index schema if needed at the start of each monitor execution

lezzago:
*Issue #, if available:*
#841 
*Description of changes:*
Update config index schema if needed at the start of each monitor execution. This fixes issues when upgrading the cluster where the index mapping changed and the user has not indexed or updated a new monitor yet.
*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunnerService.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunnerService.kt
@@ -228,23 +228,6 @@ object MonitorRunnerService : JobRunner, CoroutineScope, AbstractLifecycleCompon
     }
 
     suspend fun runJob(job: ScheduledJob, periodStart: Instant, periodEnd: Instant, dryrun: Boolean): MonitorRunResult<*> {
-
-        // Updating the scheduled job index at the start of monitor execution runs for when there is an upgrade the the schema mapping
-        // has not been updated.
-        if (!IndexUtils.scheduledJobIndexUpdated && monitorCtx.clusterService != null && monitorCtx.client != null) {
-            IndexUtils.updateIndexMapping(
-                ScheduledJob.SCHEDULED_JOBS_INDEX,
-                ScheduledJobIndices.scheduledJobMappings(), monitorCtx.clusterService!!.state(), monitorCtx.client!!.admin().indices(),
-                object : ActionListener<AcknowledgedResponse> {
-                    override fun onResponse(response: AcknowledgedResponse) {
-                    }
-                    override fun onFailure(t: Exception) {
-                        logger.error("Failed to update config index schema", AlertingException.wrap(t))
-                    }
-                }
-            )
-        }
-
         val monitor = job as Monitor
         val runResult = if (monitor.isBucketLevelMonitor()) {
             BucketLevelMonitorRunner.runMonitor(monitor, monitorCtx, periodStart, periodEnd, dryrun)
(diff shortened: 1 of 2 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunnerService.kt
```diff
@@ -223,6 +228,23 @@ object MonitorRunnerService : JobRunner, CoroutineScope, AbstractLifecycleCompon
     }
 
     suspend fun runJob(job: ScheduledJob, periodStart: Instant, periodEnd: Instant, dryrun: Boolean): MonitorRunResult<*> {
+
+        // Updating the scheduled job index at the start of monitor execution runs for when there is an upgrade the the schema mapping
+        // has not been updated.
+        if (!IndexUtils.scheduledJobIndexUpdated && monitorCtx.clusterService != null && monitorCtx.client != null) {
```
getsaurabh02, line 234: Do we need to add tests to validate this execution flow?

lezzago: It would have to be a part of this issue: https://github.com/opensearch-project/alerting/issues/848.

We have no framework for this setup

eirsep, line 234: Why is this not being done at node start up?

alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunnerService.kt
```diff
@@ -223,6 +228,23 @@ object MonitorRunnerService : JobRunner, CoroutineScope, AbstractLifecycleCompon
...
+            IndexUtils.updateIndexMapping(
+                ScheduledJob.SCHEDULED_JOBS_INDEX,
+                ScheduledJobIndices.scheduledJobMappings(), monitorCtx.clusterService!!.state(), monitorCtx.client!!.admin().indices(),
+                object : ActionListener<AcknowledgedResponse> {
+                    override fun onResponse(response: AcknowledgedResponse) {
+                    }
+                    override fun onFailure(t: Exception) {
+                        logger.error("Failed to update config index schema", AlertingException.wrap(t))
```
eirsep, line 242: plz log exception as is. Dont wrap it.
Exception should be wrapped and thrown to user after logging original exception

eirsep (changes requested):
Minor comment

alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunnerService.kt
eirsep: scheduled jobs index schema shouldnt be handled inside a scheduled job. Tomorrow when we have workflows this might become a miss. Let's handle it at a holistic level

alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunnerService.kt
lezzago: done

alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunnerService.kt
lezzago: The only place if we do it at node startup would be in the Alerting plugin class, but it would not work there since to make some of the calls to check the index would not work as the node is still setting up services. I felt it was too much to startup another service from the Alerting plugin class to update the config schema if needed. I believe we can just have the same logic in the workflow runner service to handle this. Additionally the workflow runner service and monitor runner service have very similar logic so maybe parts of it can be refactored to reduce redundant code.

getsaurabh02:
Looks like bwc test is failing

getsaurabh02 (approved):

AWSHurneyt (approved):

PR #882, opened 2023-04-19
Fixed deleting monitor workflow metadata

stevanbz:
*Issue #, if available:*
Deleting the workflow and potential delete of it's metadata.

*Description of changes:*
This PR handles workflow removal, removal of it's delegate monitors - deleting the monitors, theirs metadata, potentially queries from percolate index (if the delegate is doc level monitor), and removal of workflow metadata.

Workflow execution will cause underlying monitor metadata to be generated. Beside deleting the workflow, user can delete the delegate monitors and their metadata - if the delegate monitor is being deleted, all it's metadata should be deleted: metadata related with the given workflow execution and metadata related with the monitor execution.

If user has cluster privilege for deleting the workflows he should be able to delete the delegate monitors that are part of the given workflow. 
Once the workflow deletion starts, the context is being stashed meaning that the program in the section when deleting the workflows, monitors and metadata won't have any thread context related data. Namely, when starting alerting transport actions 
next check is being called: if (!validateUserBackendRoles(user, actionListener)) {
return
}
This check first checks if the the filterBy is enabled and then check if the user is null -- if both of them are true - the error is raised.
The check prevents to call one transport action from another with stashed context - since it will raise an error (because when the context is stashed the user will be null). In order to avoid this situation, the common parts of the code related with checking if the monitor is part of some workflow, deleting the monitors, their metadata and queries is extracted to a helper/utility class named: [MonitorDeleteService](https://github.com/opensearch-project/alerting/pull/882/files#diff-76485686809f4b34c3520d27d44ad9f57ae66f2bb3875577906f0906820cd1dcR44).

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/DeleteMonitorService.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/DeleteMonitorService.kt
@@ -1,0 +1,170 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.alerting
+
+import kotlinx.coroutines.CoroutineName
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import org.apache.logging.log4j.LogManager
+import org.apache.lucene.search.join.ScoreMode
+import org.opensearch.action.ActionListener
+import org.opensearch.action.admin.indices.delete.DeleteIndexRequest
+import org.opensearch.action.admin.indices.exists.indices.IndicesExistsRequest
+import org.opensearch.action.admin.indices.exists.indices.IndicesExistsResponse
+import org.opensearch.action.delete.DeleteRequest
+import org.opensearch.action.delete.DeleteResponse
+import org.opensearch.action.search.SearchRequest
+import org.opensearch.action.search.SearchResponse
+import org.opensearch.action.support.IndicesOptions
+import org.opensearch.action.support.WriteRequest.RefreshPolicy
+import org.opensearch.action.support.master.AcknowledgedResponse
+import org.opensearch.alerting.opensearchapi.suspendUntil
+import org.opensearch.alerting.util.AlertingException
+import org.opensearch.client.Client
+import org.opensearch.commons.alerting.action.DeleteMonitorResponse
+import org.opensearch.commons.alerting.model.Monitor
+import org.opensearch.commons.alerting.model.ScheduledJob
+import org.opensearch.commons.alerting.model.Workflow
+import org.opensearch.index.query.QueryBuilders
+import org.opensearch.index.reindex.BulkByScrollResponse
+import org.opensearch.index.reindex.DeleteByQueryAction
+import org.opensearch.index.reindex.DeleteByQueryRequestBuilder
+import org.opensearch.search.builder.SearchSourceBuilder
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+import kotlin.coroutines.suspendCoroutine
+
+/**
+ * Component used when deleting the monitors
+ */
+object DeleteMonitorService :
+    CoroutineScope by CoroutineScope(SupervisorJob() + Dispatchers.Default + CoroutineName("WorkflowMetadataService")) {
+    private val log = LogManager.getLogger(this.javaClass)
+
+    private lateinit var client: Client
+
+    fun initialize(
+        client: Client
+    ) {
+        this.client = client
+    }
+
+    /**
+     * Deletes the monitor, docLevelQueries and monitor metadata
+     * @param monitor monitor to be deleted
+     * @param refreshPolicy
+     */
+    suspend fun deleteMonitor(monitor: Monitor, refreshPolicy: RefreshPolicy): DeleteMonitorResponse {
+        val deleteResponse = deleteMonitor(monitor.id, refreshPolicy)
+        deleteDocLevelMonitorQueriesAndIndices(monitor)
+        deleteMetadata(monitor)
+        return DeleteMonitorResponse(deleteResponse.id, deleteResponse.version)
+    }
+
+    private suspend fun deleteMonitor(monitorId: String, refreshPolicy: RefreshPolicy): DeleteResponse {
+        val deleteMonitorRequest = DeleteRequest(ScheduledJob.SCHEDULED_JOBS_INDEX, monitorId)
+            .setRefreshPolicy(refreshPolicy)
+        return client.suspendUntil { delete(deleteMonitorRequest, it) }
+    }
+
+    private suspend fun deleteMetadata(monitor: Monitor) {
+        val deleteRequest = DeleteRequest(ScheduledJob.SCHEDULED_JOBS_INDEX, "${monitor.id}-metadata")
+        val deleteResponse: DeleteResponse = client.suspendUntil { delete(deleteRequest, it) }
+    }
+
+    private suspend fun deleteDocLevelMonitorQueriesAndIndices(monitor: Monitor) {
+        val metadata = MonitorMetadataService.getMetadata(monitor)
+        metadata?.sourceToQueryIndexMapping?.forEach { (_, queryIndex) ->
+
+            val indicesExistsResponse: IndicesExistsResponse =
+                client.suspendUntil {
+                    client.admin().indices().exists(IndicesExistsRequest(queryIndex), it)
+                }
+            if (indicesExistsResponse.isExists == false) {
+                return
+            }
+            // Check if there's any queries from other monitors in this queryIndex,
+            // to avoid unnecessary doc deletion, if we could just delete index completely
+            val searchResponse: SearchResponse = client.suspendUntil {
+                search(
+                    SearchRequest(queryIndex).source(
+                        SearchSourceBuilder()
+                            .size(0)
+                            .query(
+                                QueryBuilders.boolQuery().mustNot(
+                                    QueryBuilders.matchQuery("monitor_id", monitor.id)
+                                )
+                            )
+                    ).indicesOptions(IndicesOptions.LENIENT_EXPAND_OPEN_HIDDEN),
+                    it
+                )
+            }
+            if (searchResponse.hits.totalHits.value == 0L) {
+                val ack: AcknowledgedResponse = client.suspendUntil {
+                    client.admin().indices().delete(
+                        DeleteIndexRequest(queryIndex).indicesOptions(IndicesOptions.LENIENT_EXPAND_OPEN_HIDDEN),
+                        it
+                    )
+                }
+                if (ack.isAcknowledged == false) {
+                    log.error("Deletion of concrete queryIndex:$queryIndex is not ack'd!")
+                }
+            } else {
+                // Delete all queries added by this monitor
+                val response: BulkByScrollResponse = suspendCoroutine { cont ->
+                    DeleteByQueryRequestBuilder(client, DeleteByQueryAction.INSTANCE)
+                        .source(queryIndex)
+                        .filter(QueryBuilders.matchQuery("monitor_id", monitor.id))
+                        .refresh(true)
+                        .execute(
+                            object : ActionListener<BulkByScrollResponse> {
+                                override fun onResponse(response: BulkByScrollResponse) = cont.resume(response)
+                                override fun onFailure(t: Exception) = cont.resumeWithException(t)
+                            }
+                        )
+                }
+            }
+        }
+    }
+
+    /**
+     * Checks if the monitor is part of the workflow
+     *
+     * @param monitorId id of monitor that is checked if it is a workflow delegate
+     */
+    suspend fun monitorIsWorkflowDelegate(monitorId: String): Boolean {
+        val queryBuilder = QueryBuilders.nestedQuery(
+            Workflow.WORKFLOW_DELEGATE_PATH,
+            QueryBuilders.boolQuery().must(
+                QueryBuilders.matchQuery(
+                    Workflow.WORKFLOW_MONITOR_PATH,
+                    monitorId
+                )
+            ),
+            ScoreMode.None
+        )
+        try {
+            val searchRequest = SearchRequest()
+                .indices(ScheduledJob.SCHEDULED_JOBS_INDEX)
+                .source(SearchSourceBuilder().query(queryBuilder))
+
+            client.threadPool().threadContext.stashContext().use {
+                val searchResponse: SearchResponse = client.suspendUntil { search(searchRequest, it) }
+                if (searchResponse.hits.totalHits?.value == 0L) {
+                    return false
+                }
+
+                val workflowIds = searchResponse.hits.hits.map { it.id }.joinToString()
+                log.info("Monitor $monitorId can't be deleted since it belongs to $workflowIds")
+                return true
+            }
+        } catch (ex: Exception) {
+            log.error("Error getting the monitor workflows", ex)
+            throw AlertingException.wrap(ex)
+        }
+    }
+}
--- a/alerting/src/main/kotlin/org/opensearch/alerting/model/MonitorMetadata.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/model/MonitorMetadata.kt
@@ -120,9 +120,10 @@ data class MonitorMetadata(
             return MonitorMetadata(sin)
         }
 
-        fun getId(monitor: Monitor, workflowId: String? = null): String {
-            return if (workflowId.isNullOrEmpty()) "${monitor.id}-metadata"
-            else "${monitor.id}-$workflowId-metadata"
+        fun getId(monitor: Monitor, workflowMetadataId: String? = null): String {
+            return if (workflowMetadataId.isNullOrEmpty()) "${monitor.id}-metadata"
+            // WorkflowMetadataId already contains -metadata suffix
+            else "${monitor.id}-$workflowMetadataId"
         }
     }
 }
--- a/alerting/src/main/kotlin/org/opensearch/alerting/model/MonitorMetadata.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/model/MonitorMetadata.kt
@@ -123,14 +123,14 @@ data class MonitorMetadata(
         /** workflowMetadataId is used as key for monitor metadata in the case when the workflow execution happens
          so the monitor lastRunContext (in the case of doc level monitor) is not interfering with the monitor execution
          WorkflowMetadataId will be either workflowId-metadata (when executing the workflow as it is scheduled)
-         or timestampWithUUID-metadata (when the workflow is temp - ie. executed in a dry-run mode)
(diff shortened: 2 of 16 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/model/MonitorMetadata.kt
```diff
@@ -120,9 +120,10 @@ data class MonitorMetadata(
             return MonitorMetadata(sin)
         }
 
-        fun getId(monitor: Monitor, workflowId: String? = null): String {
-            return if (workflowId.isNullOrEmpty()) "${monitor.id}-metadata"
-            else "${monitor.id}-$workflowId-metadata"
+        fun getId(monitor: Monitor, workflowMetadataId: String? = null): String {
+            return if (workflowMetadataId.isNullOrEmpty()) "${monitor.id}-metadata"
+            // WorkflowMetadataId already contains -metadata suffix
+            else "${monitor.id}-$workflowMetadataId"
```
eirsep, line 126: what is the motivation behind this change

alerting/src/main/kotlin/org/opensearch/alerting/transport/SecureTransportAction.kt
```diff
@@ -76,18 +76,12 @@ interface SecureTransportAction {
     }
 
     fun <T : Any> validateUserBackendRoles(user: User?, actionListener: ActionListener<T>): Boolean {
+        if (!filterByEnabled || user == null) {
+            return true
+        }
+
         if (filterByEnabled) {
```
stevanbz, line 83: I think we maybe have an issue on alerting when starting the transport action when security is enabled and filterBy also.
Just wanted to check with you what's your opinion.

Namely, when starting alerting transport actions (like for example TransportIndexMonitorAction: https://github.com/opensearch-project/alerting/blob/main/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt#L126-L128)
we are calling this check: if (!validateUserBackendRoles(user, actionListener)) {
return
}

This check first checks if the the filterBy is enabled and then check if the user is null -- if both of them are true - the error is raised.
Check it out here: https://github.com/opensearch-project/alerting/blob/main/alerting/src/main/kotlin/org/opensearch/alerting/transport/SecureTransportAction.kt#L79-L82

On other plugins, for example on notification, this check is slightly different:
https://github.com/opensearch-project/notifications/blob/main/notifications/notifications/src/main/kotlin/org/opensearch/notifications/security/UserAccessManager.kt#L55-L57
or for example on index-management:
https://github.com/opensearch-project/index-management/blob/main/src/main/kotlin/org/opensearch/indexmanagement/util/SecurityUtils.kt#L137-L139

ie. it checks if the user is null OR if the filterBy is enabled and returns true.
The check we have on alerting will prevent us, to call one transport action from another with stashed context - since it will raise an error (because when the context is stashed the user will be null)

Should we adjust the check on alerting to reflect the same logic or? Tnx in advance!
@eirsep @lezzago

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteWorkflowAction.kt
```diff
@@ -150,26 +154,38 @@ class TransportDeleteWorkflowAction @Inject constructor(
...
+                    var deleteWorkflowResponse = DeleteWorkflowResponse(deleteResponse.id, deleteResponse.version)
+
+                    val workflowMetadataId = WorkflowMetadata.getId(workflow.id)
+
+                    val metadataIdsToDelete = mutableListOf(workflowMetadataId)
+
                     if (deleteDelegateMonitors == true) {
-                        if (user != null && filterByEnabled) {
-                            // Un-stash the context
-                            withClosableContext(
```
stevanbz, line 157: @eirsep @lezzago - if we conclude that we need to change the part I described in the comment above, we should definitely remove the context unstash so the program in next couple of lines, once we are deleting the monitor, can behave as an admin.

If we are going to keep the check in the comment above as it is, then the user must have monitor_delete cluster privilege - meaning that even before starting workflow deletion we would need to check it's cluster permission (which I think we shouldn't do and I don't know how to do since the security plugin does this I guess).

alerting/src/main/kotlin/org/opensearch/alerting/transport/SecureTransportAction.kt
```diff
@@ -76,18 +76,12 @@ interface SecureTransportAction {
     }
 
     fun <T : Any> validateUserBackendRoles(user: User?, actionListener: ActionListener<T>): Boolean {
+        if (!filterByEnabled || user == null) {
```
stevanbz, line 79: I think we maybe have an issue on alerting when starting the transport action when security is enabled and filterBy also.
Just wanted to check with you what's your opinion.

Namely, when starting alerting transport actions (like for example TransportIndexMonitorAction: https://github.com/opensearch-project/alerting/blob/main/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt#L126-L128)
we are calling this check: if (!validateUserBackendRoles(user, actionListener)) {
return
}

This check first checks if the the filterBy is enabled and then check if the user is null -- if both of them are true - the error is raised.
Check it out here: https://github.com/opensearch-project/alerting/blob/main/alerting/src/main/kotlin/org/opensearch/alerting/transport/SecureTransportAction.kt#L79-L82

On other plugins, for example on notification, this check is slightly different:
https://github.com/opensearch-project/notifications/blob/main/notifications/notifications/src/main/kotlin/org/opensearch/notifications/security/UserAccessManager.kt#L55-L57
or for example on index-management:
https://github.com/opensearch-project/index-management/blob/main/src/main/kotlin/org/opensearch/indexmanagement/util/SecurityUtils.kt#L137-L139

ie. it checks if the user is null OR if the filterBy is enabled and returns true.
The check we have on alerting will prevent us, to call one transport action from another with stashed context - since it will raise an error (because when the context is stashed the user will be null)

Should we adjust the check on alerting to reflect the same logic or? Tnx in advance!
@eirsep @lezzago

stevanbz: @getsaurabh02 Adding Saurabh so he can also participate

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteWorkflowAction.kt
stevanbz: @getsaurabh02 also, here

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteWorkflowAction.kt
```diff
@@ -150,26 +154,38 @@ class TransportDeleteWorkflowAction @Inject constructor(
...
-                            deleteMonitors(delegateMonitorIds, RefreshPolicy.IMMEDIATE)
-                        }
+                        val failedMonitorIds = tryDeletingMonitors(delegateMonitorIds, RefreshPolicy.IMMEDIATE)
+                        // Update delete workflow response
+                        deleteWorkflowResponse.nonDeletedMonitors = failedMonitorIds
+                        // Delete monitors workflow metadata
+                        // Monitor metadata will be in monitorId-workflowId-metadata format
+                        metadataIdsToDelete.addAll(deletableMonitors.map { MonitorMetadata.getId(it, workflowMetadataId) })
+                    }
+                    // Delete the monitors and it's metadata
```
lezzago, line 172: Update comment to say `// Delete the monitors' workflow metadata` as that is what it is actually doing.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteWorkflowAction.kt
```diff
@@ -150,26 +154,38 @@ class TransportDeleteWorkflowAction @Inject constructor(
...
+                            .source(ScheduledJob.SCHEDULED_JOBS_INDEX)
+                            .filter(QueryBuilders.idsQuery().addIds(*metadataIdsToDelete.toTypedArray()))
+                            .execute(it)
                     }
-                    actionListener.onResponse(DeleteWorkflowResponse(deleteResponse.id, deleteResponse.version))
+
+                    if (deleteMonitorWorkflowMetadataResponse.isTimedOut) {
+                        actionListener.onFailure(
+                            AlertingException.wrap(
+                                OpenSearchException("Cannot determine that the ${ScheduledJob.SCHEDULED_JOBS_INDEX} index is healthy")
```
lezzago, line 183: Change the exception message. It could be something like: `Cannot delete monitors' workflow metadata due to timeout.`

lezzago (changes requested):

alerting/src/main/kotlin/org/opensearch/alerting/model/MonitorMetadata.kt
```diff
@@ -120,9 +120,16 @@ data class MonitorMetadata(
...
 
-        fun getId(monitor: Monitor, workflowId: String? = null): String {
-            return if (workflowId.isNullOrEmpty()) "${monitor.id}-metadata"
-            else "${monitor.id}-$workflowId-metadata"
+        // workflowMetadataId is used as key for monitor metadata in the case when the workflow execution happens
+        // so the monitor lastRunContext (in the case of doc level monitor) is not interfering with the monitor execution
+        // WorkflowMetadataId will be either workflowId-metadata (when executing the workflow as it is scheduled)
+        // or timestampWithUUID-metadata (when the workflow is temp - ie. executed in a dry-run mode)
+        // In the case of temp workflow, doc level monitors must have lastRunContext created from scratch
+        // That's why we are using workflowMetadataId - in order to ensure that the doc level monitor metadata is created from scratch
```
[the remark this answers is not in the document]
stevanbz: Got it. Tnx. Will do so

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteWorkflowAction.kt
stevanbz: @lezzago @eirsep @getsaurabh02 
Answered in previous response

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteWorkflowAction.kt
```diff
@@ -150,26 +154,38 @@ class TransportDeleteWorkflowAction @Inject constructor(
...
+                        val failedMonitorIds = tryDeletingMonitors(delegateMonitorIds, RefreshPolicy.IMMEDIATE)
+                        // Update delete workflow response
+                        deleteWorkflowResponse.nonDeletedMonitors = failedMonitorIds
+                        // Delete monitors workflow metadata
+                        // Monitor metadata will be in monitorId-workflowId-metadata format
+                        metadataIdsToDelete.addAll(deletableMonitors.map { MonitorMetadata.getId(it, workflowMetadataId) })
+                    }
+                    // Delete the monitors and it's metadata
+                    val deleteMonitorWorkflowMetadataResponse: BulkByScrollResponse = client.suspendUntil {
+                        DeleteByQueryRequestBuilder(client, DeleteByQueryAction.INSTANCE)
```
stevanbz, line 174: Here should say:
 DeleteByQueryRequestBuilder(this, DeleteByQueryAction.INSTANCE)

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteWorkflowAction.kt
stevanbz: Makes sense

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteWorkflowAction.kt
```diff
@@ -193,24 +206,36 @@ class TransportDeleteWorkflowAction @Inject constructor(
...
-                val deleteRequest = DeleteMonitorRequest(monitorId, refreshPolicy)
-                val searchResponse: DeleteMonitorResponse = client.suspendUntil {
-                    AlertingPluginInterface.deleteMonitor(this as NodeClient, deleteRequest, it)
+            for (monitor in monitors) {
+                val monitorId = monitor.id
+                try {
+                    val monitorIsWorkflowDelegate = DeleteMonitorService.monitorIsWorkflowDelegate(monitorId)
+                    if (monitorIsWorkflowDelegate) {
+                        nonDeletedMonitorIds.add(monitorId)
+                    }
```
lezzago, line 224: Why is this needed? We already check if this monitor is not part of any other workflow.

lezzago (changes requested):

alerting/src/main/kotlin/org/opensearch/alerting/DeleteMonitorService.kt
```diff
@@ -0,0 +1,170 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.alerting
```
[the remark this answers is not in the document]
stevanbz: Maybe if we create a new package named service?

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteWorkflowAction.kt
stevanbz: You are right. 
We already filtered the monitors that can be deleted(we can delete only the monitors belonging to a given workflow - if the monitor is shared between multiple workflows - those ones will be skipped)

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteWorkflowAction.kt
```diff
@@ -270,7 +264,7 @@ class TransportDeleteWorkflowAction @Inject constructor(
                 }
                 workflow.copy(id = hit.id, version = hit.version)
             }
-            val workflowMonitors = workflows.filter { it.id != workflowIdToBeDeleted }.flatMap { (it.inputs[0] as CompositeInput).getMonitorIds() }.distinct()
+            val workflowMonitors = workflows.flatMap { (it.inputs[0] as CompositeInput).getMonitorIds() }.distinct()
```
lezzago, line 267: Why are we removing this filter? We need to ignore the workflow we want to delete.

alerting/src/test/kotlin/org/opensearch/alerting/WorkflowMonitorIT.kt
```diff
@@ -595,9 +648,31 @@ class WorkflowMonitorIT : WorkflowSingleNodeTestCase() {
...
+            e.message?.let {
+                assertTrue(
+                    "Exception not returning GetMonitor Action error ",
+                    it.contains("List is empty")
+                )
+            }
+        }
+
+        try {
+            searchMonitorMetadata(monitorMetadataId2)
```
lezzago, line 664: We need to validate the result to make sure it is deleted.

alerting/src/test/kotlin/org/opensearch/alerting/WorkflowMonitorIT.kt
```diff
@@ -595,9 +648,31 @@ class WorkflowMonitorIT : WorkflowSingleNodeTestCase() {
                 )
             }
         }
+        // Verify that the monitors metadata are deleted
+        try {
+            searchMonitorMetadata(monitorMetadataId1)
```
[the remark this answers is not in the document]
lezzago: I think the same issue is there with the usages of `searchWorkflowMetadata` and `getWorkflowById` for checking if it was deleted.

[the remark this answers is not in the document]
lezzago: If it doesnt go through the exception, does it mean it should fail the test? If so, we should add a `fail` right after `searchMonitorMetadata(monitorMetadataId1)`. Same for the similar comment below.

alerting/src/main/kotlin/org/opensearch/alerting/service/DeleteMonitorService.kt
```diff
@@ -0,0 +1,171 @@
...
+/**
+ * Component used when deleting the monitors
+ */
+object DeleteMonitorService :
+    CoroutineScope by CoroutineScope(SupervisorJob() + Dispatchers.Default + CoroutineName("WorkflowMetadataService")) {
+    private val log = LogManager.getLogger(this.javaClass)
+
+    private lateinit var client: Client
+
+    fun initialize(
```
eirsep, line 51: we can just accept client in the method argument and make this a DeleteMonitorUtils class.

eirsep: i will raise a PR making that change.

eirsep (changes requested):

alerting/src/main/kotlin/org/opensearch/alerting/service/DeleteMonitorService.kt
stevanbz: We can - makes sense to do this change. 
I just followed the pattern we used in MetadataService -> great. Tnx for making this change.

alerting/src/test/kotlin/org/opensearch/alerting/WorkflowMonitorIT.kt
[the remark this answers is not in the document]
eirsep: adding fail() check within try() as catch block asserts result list is empty

alerting/src/test/kotlin/org/opensearch/alerting/WorkflowMonitorIT.kt
eirsep: adding fail() check within try() as catch block asserts result list is empty
 addressed https://github.com/opensearch-project/alerting/commit/a3105171b11147bc57241d1bf164477ae13e7fb4

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteWorkflowAction.kt
```diff
@@ -150,26 +151,38 @@ class TransportDeleteWorkflowAction @Inject constructor(
...
-                    actionListener.onResponse(DeleteWorkflowResponse(deleteResponse.id, deleteResponse.version))
+
+                    if (deleteMonitorWorkflowMetadataResponse.isTimedOut) {
+                        actionListener.onFailure(
+                            AlertingException.wrap(
+                                OpenSearchException("Cannot delete monitors workflow metadata due to timeout.")
+                            )
+                        )
+                        return
+                    }
```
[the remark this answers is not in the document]
eirsep: Checking.. will handle holistically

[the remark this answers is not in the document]
eirsep: removed that check [6700a1e](https://github.com/opensearch-project/alerting/pull/882/commits/6700a1e241147afaecf0dd692d16a3d86c6cb054)

Agreed that timeout requires no special attention

lezzago (approved):

Issue #817, opened 2023-02-23
Revisiting Monitor Error Alerts


eirsep:
Doc level monitors create error alerts on each run where there is an error.
This would lead us to filling up the index with error alerts if there are frequently scheduled monitors which are erroring in each run (such as an incorrect index name or an index being deleted etc.)
This is noisy behaviour.
We can check if there is an unacknowledge ERROR state alert and only send out notifications and not create a new alert. If notification is too noisy we can choose to update the last updated time of alert

eirsep:
Will add more details for each type of monitor's error Alert logic.

eirsep:
When there already exists ERROR state alert and end time is null-
1. again monitor errors - upsert current alert error message (if message is different else just update lastNotificationTime)
2. if monitor's next execution is successful -> set an end time on the current error alert

We should roll over alerts with non-null end time into history index for doc level monitor alerts

@petardz

PR #892, opened 2023-04-27
DocLevelMonitor Error Alert - rework

petardz:
*Issue #, if available:*
#817
*Description of changes:*

Specific "Error Alert" is created for a monitor, when error happens during monitor run. State of alert is ERROR and subsequent errors would upsert this "Error Alert". If error message is different then previous one, errorMessage field would be overwritten and previous error would be added to errorHistory rolling array(10 elements).

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
@@ -325,12 +325,9 @@ class AlertService(
             }
             // Deserialize first/latest Alert
             val hit = searchResponse.hits.hits[0]
-            val xcp = XContentHelper.createParser(
-                xContentRegistry, LoggingDeprecationHandler.INSTANCE,
-                hit.sourceRef, XContentType.JSON
-            )
-            XContentParserUtils.ensureExpectedToken(XContentParser.Token.START_OBJECT, xcp.nextToken(), xcp)
+            val xcp = contentParser(hit.sourceRef)
             val existingErrorAlert = Alert.parse(xcp, hit.id, hit.version)
+
             val currentTime = Instant.now()
             alert = if (alert.errorMessage != existingErrorAlert.errorMessage) {
                 var newErrorHistory = existingErrorAlert.errorHistory.update(
--- a/alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
@@ -13,6 +13,7 @@ import org.opensearch.action.bulk.BulkRequest
 import org.opensearch.action.bulk.BulkResponse
 import org.opensearch.action.delete.DeleteRequest
 import org.opensearch.action.index.IndexRequest
+import org.opensearch.action.index.IndexResponse
 import org.opensearch.action.search.SearchRequest
 import org.opensearch.action.search.SearchResponse
 import org.opensearch.action.support.WriteRequest
@@ -39,6 +40,7 @@ import org.opensearch.commons.alerting.model.AggregationResultBucket
 import org.opensearch.commons.alerting.model.Alert
 import org.opensearch.commons.alerting.model.BucketLevelTrigger
 import org.opensearch.commons.alerting.model.DataSources
+import org.opensearch.commons.alerting.model.DocumentLevelTrigger
 import org.opensearch.commons.alerting.model.Monitor
 import org.opensearch.commons.alerting.model.Trigger
 import org.opensearch.commons.alerting.model.action.AlertCategory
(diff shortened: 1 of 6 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
```diff
@@ -187,6 +191,21 @@ class AlertService(
         )
     }
 
+    fun composeMonitorErrorAlert(
```
eirsep, line 194: is this only for doc level monitor? if yes, plz rename method to reflect that

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
```diff
@@ -187,6 +191,21 @@ class AlertService(
...
+        id: String,
+        monitor: Monitor,
+        alertError: AlertError
+    ): Alert {
+        val currentTime = Instant.now()
+        // TODO make Alert constructor without trigger param so that we don't have to create this dummy trigger
+        val trigger = DocumentLevelTrigger(id = "", name = "dummy-trigger", severity = "", listOf(), Script(""))
```
eirsep, line 201: we can't call this dummy trigger. maybe name could be "${monitorId}ExecutionError"

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
```diff
@@ -281,6 +300,60 @@ class AlertService(
...
+        val errorAlertIdPrefix = "error-alert"
+        val newErrorAlertId = "$errorAlertIdPrefix-${monitor.id}-${UUID.randomUUID()}"
+
+        val searchRequest = SearchRequest("${monitor.dataSources.alertsIndex}*")
+            .source(
+                SearchSourceBuilder()
+                    .sort(Alert.START_TIME_FIELD, SortOrder.DESC)
+                    .query(
+                        QueryBuilders.boolQuery()
+                            .must(QueryBuilders.queryStringQuery("${Alert.ALERT_ID_FIELD}:$errorAlertIdPrefix*"))
```
eirsep, line 313: this will not be able to fetch existing errors (i.e. alerts created before this code was deployed)

Let's re-evaluate on how to handle for existing alerts

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
```diff
@@ -281,6 +300,60 @@ class AlertService(
...
+                            .must(QueryBuilders.termQuery(Alert.MONITOR_ID_FIELD, monitor.id))
+                            .must(QueryBuilders.termQuery(Alert.STATE_FIELD, Alert.State.ERROR))
+                    )
+            )
+        val searchResponse: SearchResponse = client.suspendUntil { search(searchRequest, it) }
+
+        var alert = composeMonitorErrorAlert(newErrorAlertId, monitor, AlertError(Instant.now(), errorMessage))
+
+        if (searchResponse.hits.totalHits.value > 0L) {
```
eirsep, line 322: if these hits are 0 it might also indicate there are older alerts.

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
petardz: Not sure if we can isolate existing error alerts.

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
petardz: They could be used for other monitors too I guess

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
petardz: They could be used for other monitors too

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
petardz: done

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
eirsep: the trigger you used is a doc level trigger

petardz: Ah.. it might make sense to allow Alert with null trigger because of this usecase

eirsep (approved):

sbcd90 (approved):

sbcd90 (approved):

lezzago (changes requested):

PR #894, opened 2023-05-01
Fixed a bug that prevented alerts from being generated for doc level monitors that use wildcard characters in index names.

AWSHurneyt:
*Issue #, if available:*

*Description of changes:*
1. Fixed a bug that prevented alerts from being generated for doc level monitors that use wildcard characters in index names.
2. Fixed integration tests that examine this use case.

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
@@ -49,6 +49,7 @@ import org.opensearch.commons.alerting.util.string
 import org.opensearch.core.xcontent.ToXContent
 import org.opensearch.core.xcontent.XContentBuilder
 import org.opensearch.index.query.BoolQueryBuilder
+import org.opensearch.index.query.Operator
 import org.opensearch.index.query.QueryBuilders
 import org.opensearch.percolator.PercolateQueryBuilderExt
 import org.opensearch.rest.RestStatus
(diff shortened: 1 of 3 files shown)


AWSHurneyt:
The BWC and security github checks are expected to be failing on `main` at the moment.

lezzago:
Did we figure out exactly why the previous tests did not catch this bug?

AWSHurneyt:
### Some findings regarding the test that failed to pick up this bug
The `test execute monitor with wildcard index that generates alerts and findings` test used the `*` wildcard character to configure the `index` for the test monitor. However, it still didn't pick up this bug; alerts and findings were still generated.

Some preliminary troubleshooting so far suggests the issue may actually relate to the names used for the test indices; `test1`, `test2`, and `test*` in the monitor configuration itself.

(For reference, the test documents contained the value `"test_field" : "us-west-2"`.)

Even without the fix included in this PR, the test passes when the query is defined as `query = "NOT (test_field:\"us-west-1\")"` or `query = "test_field:\"us-west-2\""`. However, I noticed that the test was using `test1` and `test2` for the index names whereas my manual testing used `test-index1` and `test-index2`. I adjusted the test to add the `-index` to the index names, and it started to fail as expected; but only for `query = "NOT (test_field:\"us-west-1\")"`. After applying the fix, it passed as expected for both query test cases.

lezzago (approved):

lezzago:
Lets create an issue to follow up on investigation

eirsep (approved):

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -543,11 +544,11 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
     ): SearchHits {
-        val boolQueryBuilder = BoolQueryBuilder().filter(QueryBuilders.matchQuery("index", index))
+        val boolQueryBuilder = BoolQueryBuilder().must(QueryBuilders.matchQuery("index", index).operator(Operator.AND))
 
         val percolateQueryBuilder = PercolateQueryBuilderExt("query", docs, XContentType.JSON)
         if (monitor.id.isNotEmpty()) {
-            boolQueryBuilder.filter(QueryBuilders.matchQuery("monitor_id", monitor.id))
+            boolQueryBuilder.must(QueryBuilders.matchQuery("monitor_id", monitor.id).operator(Operator.AND))
         }
         boolQueryBuilder.filter(percolateQueryBuilder)
```
AWSHurneyt, line 553: In follow-up revision, refactor this from `.filter()` to `.must()`.

eirsep:
THE percolateQueryBuilder sub-query is still being added as a "query.filter()" 
we should add that also as a MUSTxAND condition 
Otherwise we are liable to having a change there where we add a new subquery with query.filter() which makes the percolate query optional in an implicit OR with the newly added sub-query

@lezzago @getsaurabh02

PR #925, opened 2023-05-22
Chained alerts

eirsep:
*Issue #, if available:*

*Description of changes:*

*CheckList:*


alerting/src/main/kotlin/org/opensearch/alerting/DeleteMonitorService.kt
```diff
@@ -0,0 +1,139 @@
+/*
```
eirsep, line 1: will remove this.
this is duplicated due to change in path

alerting/src/main/kotlin/org/opensearch/alerting/DeleteMonitorService.kt
```diff
@@ -0,0 +1,139 @@
...
+import org.opensearch.index.reindex.DeleteByQueryRequestBuilder
+import org.opensearch.search.builder.SearchSourceBuilder
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+import kotlin.coroutines.suspendCoroutine
+
+/**
+ * Component used when deleting the monitors
+ */
+object DeleteMonitorService :
```
stevanbz, line 41: I think we can remove this file. It's outdated - we have DeleteMonitorService in another package org.opensearch.alerting.service
Check it out [here](https://github.com/opensearch-project/alerting/blob/feature/composite-monitors/alerting/src/main/kotlin/org/opensearch/alerting/service/DeleteMonitorService.kt)

core/build.gradle
```diff
@@ -15,7 +15,7 @@ dependencies {
     implementation "com.cronutils:cron-utils:9.1.6"
     api "org.opensearch.client:opensearch-rest-client:${opensearch_version}"
     implementation 'com.google.googlejavaformat:google-java-format:1.10.0'
-    api "org.opensearch:common-utils:${common_utils_version}"
+    api files("/Users/snistala/Documents/opensearch/common-utils/build/libs/common-utils-2.7.0.0-SNAPSHOT.jar")
```
lezzago, line 18: change this back

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
```diff
@@ -181,15 +186,36 @@ class AlertService(
         findings: List<String>,
         relatedDocIds: List<String>,
         ctx: DocumentLevelTriggerExecutionContext,
-        alertError: AlertError?
+        alertError: AlertError?,
+        workflowExecutionId: String? = null,
```
lezzago, line 190: should we name this to execution id? Changing the name here makes it more confusing.

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -290,7 +290,8 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
                 listOf(it.first),
                 listOf(it.second),
                 triggerCtx,
-                monitorResult.alertError() ?: triggerResult.alertError()
+                monitorResult.alertError() ?: triggerResult.alertError(),
+                workflowExecutionId // TODO add execution id for other types of monitors also
```
lezzago, line 294: Do we want to use workflowExecutionId or executionId? I thought we might want to extend to make this more generic, so would that be better to do it now?

alerting/src/main/resources/org/opensearch/alerting/alerts/alert_mapping.json
```diff
@@ -74,6 +74,9 @@
         }
       }
     },
+    "workflow_execution_id": {
```
lezzago, line 77: Do we want this to be generic and be `execution_id`? It will be harder to change it to `execution_id` later on and might be best to do that if we want to later correlate an alert to an execution for normal monitors.

eirsep: Agreed that there is value in keep execution_id generic to extend the capability for monitors executing outside workflows

i think the best solution is - for **alerts** and **findings** we should have three fields
- `monitor_id` (empty for chained alerts)
- `workflow_id` (empty for monitors executing independently outside workflows)
- `execution_id` (empty today for monitors executing independently outside workflows but can be populated tomorrow)

This would make it easy to query for alerts page in dashboard where as a user i want to see ONLY chained_alerts and independent monitor alerts (`get all alerts where monitor_id is "" or workflow_id is null`)

And in chained alert detail page for a chained alert I would want to see the delegate monitor alerts - `query by execution id and workflow_id`

alerting/src/main/kotlin/org/opensearch/alerting/workflow/CompositeWorkflowRunner.kt
```diff
@@ -35,7 +48,7 @@ object CompositeWorkflowRunner : WorkflowRunner() {
         monitorCtx: MonitorRunnerExecutionContext,
         periodStart: Instant,
         periodEnd: Instant,
-        dryRun: Boolean
+        dryRun: Boolean,
     ): WorkflowRunResult {
         val workflowExecutionStartTime = Instant.now()
```
eirsep, line 53: log start and finish of worklfow with metadata id, timestamp

eirsep:
Notification throttling
Acknowledge chained alerts
Move delegate monitor alerts to ack
setting to limit number of chained alert triggers
add search workflows API support

PR #931, opened 2023-05-24
revert to deleting monitor metadata after deleting doc level queries to fix delete monitor regression.

eirsep:
*Description of changes:*
Metadata is required for correct query index clean up during monitor deletion. Reverted to deleting monitor metadata after deleting doc level queries to fix regression.


sbcd90 (approved):

lezzago (approved):

eirsep:
#934 contains a change which will fix the mapping for findings in alerting plugin. But this PR also is required to be merged in with CI failing as this has a bug fix for the failing workflow

Once both PRs are merged the test workflow would be fixed fully. `Test Workflow` suite

PR #1020, opened 2023-07-13
Fix get alerts api to also return chained alerts with default params. and skip audit state alerts

eirsep:
Fix get alerts api to also return chained alerts with default params. and skip audit state alerts
If audit state alerts are required, need to pass monitor id, workflow id and state=audit in params.


alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
```diff
@@ -97,7 +97,15 @@ class TransportGetAlertsAction @Inject constructor(
             queryBuilder.filter(QueryBuilders.termQuery("severity", getAlertsRequest.severityLevel))
         }
 
-        if (getAlertsRequest.alertState != "ALL") {
+        if (getAlertsRequest.alertState == "ALL") {
+            // alerting dashboards expects chained alerts and individually executed monitors' alerts to be returned from this api
+            // when invoked with state=ALL. They require that audit alerts are NOT returned in this page
+            // and only be shown in "associated alerts" field under get workflow_alerts API.
+            // But if the API is called with query_params: state=AUDIT,monitor_id=<123>,workflow_id=<abc>, this api
+            // will return audit alerts generated by delegate monitor <123> in workflow <abc>
```
lezzago, line 105: If we are still letting this API call audit alerts, it should be included in ALL. It should not be just for the dashboard use case. 

Otherwise, we should not let users see audit alerts through this API and only from the workflow alerts API

lezzago: If not supporting it in `ALL` is temporary until the front end can fix it, then I am fine with going with this logic as long as we have an open github issue about it

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
```diff
@@ -107,16 +115,23 @@ class TransportGetAlertsAction @Inject constructor(
 
         if (getAlertsRequest.monitorId != null) {
             queryBuilder.filter(QueryBuilders.termQuery("monitor_id", getAlertsRequest.monitorId))
+            if (getAlertsRequest.workflowIds.isNullOrEmpty()) {
+                val noWorklfowIdQuery = QueryBuilders.boolQuery()
```
lezzago, line 119: `noWorklfowIdQuery` is misspelled.

alerting/src/main/kotlin/org/opensearch/alerting/workflow/CompositeWorkflowRunner.kt
```diff
@@ -310,7 +310,7 @@ object CompositeWorkflowRunner : WorkflowRunner() {
                 .should(boolQuery().mustNot(existsQuery(Alert.ERROR_MESSAGE_FIELD)))
                 .should(termsQuery(Alert.ERROR_MESSAGE_FIELD, ""))
             queryBuilder.must(noErrorQuery)
-            searchRequest.source().query(queryBuilder)
+            searchRequest.source().query(queryBuilder).size(9999)
```
lezzago, line 313: why?

eirsep: to make sure we fetch as many alerts as we can
we don't intend to evaluate with just 10 alerts which the default if no size is mentioned

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
eirsep: AUDIT alerts is a state designed to ållow us to create events which we don't want to get users notified about. That's why they rely on chained alerts to be actionable(notifications/acknowledgement etc)

When we return alerts in the api it would clutter the response.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
eirsep: corrected

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
eirsep: Created issue to discuss further and reconcile with front end https://github.com/opensearch-project/alerting/issues/1021

alerting/src/main/kotlin/org/opensearch/alerting/workflow/CompositeWorkflowRunner.kt
lezzago: We should be paginating through them then. Also isnt there a max of 25 delegate monitors to a workflow?

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
lezzago: We would be introducing breaking changes behavior if we change how it is later on

lezzago: I think we should make a call on if we want audit alerts here or not now.

alerting/src/main/kotlin/org/opensearch/alerting/workflow/CompositeWorkflowRunner.kt
eirsep: let's revisit this soon
https://github.com/opensearch-project/alerting/issues/1022 and take a quick follow up.

lezzago:
Approving for now, but we have to follow up on these issues called out. 
1. If audit alerts should be retrieved in the get Alerts API (#1021) (@mnkugler)
2. Pagination on fetching alerts for workflow

lezzago (approved):

sbcd90 (approved):

PR #1054, opened 2023-07-25
add scroll api to fetch associated alerts in  get workflow alerts API

eirsep:
add scroll api to fetch associated alerts in get workflow alerts API 

Each chained alert created by composite monitor may have multiple associated alerts. When get workflow Alerts API is called we need to be able to return high number of associated alerts (due to 1:n ratio of workflow alert to associated alerts)

This PR adds scroll API for the same

*Description of changes:*

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetWorkflowAlertsAction.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetWorkflowAlertsAction.kt
@@ -13,6 +13,7 @@ import org.opensearch.action.ActionListener
 import org.opensearch.action.ActionRequest
 import org.opensearch.action.search.SearchRequest
 import org.opensearch.action.search.SearchResponse
+import org.opensearch.action.search.SearchScrollRequest
 import org.opensearch.action.support.ActionFilters
 import org.opensearch.action.support.HandledTransportAction
 import org.opensearch.alerting.alerts.AlertIndices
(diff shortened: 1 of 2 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetWorkflowAlertsAction.kt
```diff
@@ -221,12 +224,20 @@ class TransportGetWorkflowAlertsAction @Inject constructor(
...
+            var searchRequest = SearchRequest(alertIndex)
+            searchRequest.source().query(queryBuilder).size(scrollSize)
+            searchRequest.scroll(scroll)
+            var response: SearchResponse = client.suspendUntil { search(searchRequest, it) }
             associatedAlerts.addAll(parseAlertsFromSearchResponse(response))
+            val scrollId = response.scrollId
+            while (response.hits.hits.isNotEmpty()) {
+                response = client.suspendUntil { client.searchScroll(SearchScrollRequest(scrollId), it) }
+                associatedAlerts.addAll(parseAlertsFromSearchResponse(response))
+            }
```
lezzago, line 238: Do we want to limit this and have page number responses in the api response so the customer can paginate? This would be dangerous if there were many associated alerts like from doc level monitors.

eirsep: we can paginate only on the primary field - alerts
we cannot paginate on the thee associated alerts field as it provides information associated to primary resource.
Open to ideas on how to handle this
Pagination requires statefulness which we provide through search query out of the box. Associated alerts is a list we compile from the alerts parsed from the primary search query response. we can’t paginate on that

lezzago: I think if the user provides the alert id, then they can paginate through the associated alerts. If they do not provide the alert id, it would paginate through the main alerts only then.

PR #1071, opened 2023-08-02
add alertId parameter in get chained alert API and paginate associated alerts if alertId param is mentioned

eirsep:
Get chained alert API will support filtering based on an alertId. The motivation is to fetch associated alerts with support for pagination. The size and sort params would apply to fetch associated alerts in this api when alert id is mentioned

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetWorkflowAlertsAction.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetWorkflowAlertsAction.kt
@@ -243,7 +243,7 @@ class TransportGetWorkflowAlertsAction @Inject constructor(
                 if (!tableProp.missing.isNullOrBlank()) {
                     sortBuilder.missing(tableProp.missing)
                 }
-                searchRequest.source().sort(sortBuilder).size(tableProp.size)
+                searchRequest.source().sort(sortBuilder).size(tableProp.size).from(tableProp.startIndex)
             }
             queryBuilder.must(QueryBuilders.termsQuery("_id", associatedAlertIds))
             queryBuilder.must(QueryBuilders.termQuery(Alert.STATE_FIELD, Alert.State.AUDIT))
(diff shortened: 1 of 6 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetWorkflowAlertsAction.kt
```diff
@@ -206,22 +211,42 @@ class TransportGetWorkflowAlertsAction @Inject constructor(
...
+                    getWorkflowAlertsRequest
+                )
             actionListener.onResponse(GetWorkflowAlertsResponse(alerts, associatedAlerts, totalAlertCount))
         } catch (e: Exception) {
             actionListener.onFailure(AlertingException("Failed to get alerts", RestStatus.INTERNAL_SERVER_ERROR, e))
         }
     }
 
-    private suspend fun getAssociatedAlerts(associatedAlerts: MutableList<Alert>, alerts: MutableList<Alert>, alertIndex: String) {
+    private suspend fun getAssociatedAlerts(
```
lezzago, line 226: How is pagination done here?

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetWorkflowAlertsAction.kt
```diff
@@ -126,13 +126,18 @@ class TransportGetWorkflowAlertsAction @Inject constructor(
                         .field("trigger_name")
                 )
         }
+        // if alert id is mentioned we cannot set "from" field as it may not return id. we would be using it to paginate associated alerts
+        val from = if (getWorkflowAlertsRequest.alertIds.isNullOrEmpty())
+            tableProp.startIndex
+        else 0
```
lezzago, line 132: The `startIndex` isn't being passed over to `getAssociatedAlerts`

lezzago (changes requested):

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetWorkflowAlertsAction.kt
eirsep: we dont need it to be passed over. we will use `tableProp.startIndex`

we need this check on `from` param if alertId is mentioned because we cannot paginate for single id in search request to fetch the chained alerts

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetWorkflowAlertsAction.kt
eirsep: fixed this and verified in test

lezzago (approved):

AWSHurneyt (approved):

PR #1079, opened 2023-08-04
Chained Alert Behaviour Changes

eirsep:
Context:
---
Workflow/Composite monitor supports triggers whose conditions are configured as Boolean conditions with AND, OR, NOT
Operators and monitors as variables. For ex. Trigger condition M1 && M2 means trigger condition is matched if both monitors M1 and M2 create alerts (in AUDIT state) in the current execution. If workflow trigger condition is matched we create Chained Alert.
 
Current behavior
---
Alerts are created in ACTIVE state. 
When alert are acknowledged they move to ACKNOWLEDGED state.

New behavior introduced with this PR
===
**- If no ACTIVE/ACKNOWLEDGEDstate alert already exists for composite monitor trigger, we create a new ACTIVE state alert.**
**- If ACTIVE/ACKNOWLEDGED state alert already exists, we update that alert’s lastNotificationTime itself. But we do NOT create a new Alert. The pre-existing ACTIVE state alert will remain in ACTIVE state.**
**- When composite monitor trigger condition is NOT matched: 
If ACTIVE/ACKNOWLEDGED  state alert already exists, we would mark it as COMPLETED.**

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
@@ -349,7 +349,7 @@ class AlertService(
                 startTime = Instant.now(),
                 lastNotificationTime = currentTime,
                 state = Alert.State.ACTIVE,
-                errorMessage = null, schemaVersion = -1,
+                errorMessage = null, schemaVersion = IndexUtils.alertIndexSchemaVersion,
                 chainedAlertTrigger = ctx.trigger,
                 executionId = executionId,
                 workflow = workflow,
@@ -846,7 +846,7 @@ class AlertService(
     }
 
     /**
-     * Searches for Alerts in the monitor's alertIndex.
+     * Searches for ACTIVE/ACKNOWLEDGED chained alerts in the workflow's alertIndex.
      *
      * @param monitorId The Monitor to get Alerts for
      * @param size The number of search hits (Alerts) to return
--- a/alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
@@ -19,6 +19,7 @@ import org.opensearch.action.search.SearchResponse
 import org.opensearch.action.support.WriteRequest
 import org.opensearch.alerting.alerts.AlertIndices
 import org.opensearch.alerting.model.ActionRunResult
+import org.opensearch.alerting.model.ChainedAlertTriggerRunResult
 import org.opensearch.alerting.model.QueryLevelTriggerRunResult
 import org.opensearch.alerting.opensearchapi.firstFailureOrNull
 import org.opensearch.alerting.opensearchapi.retry
@@ -82,6 +83,26 @@ class AlertService(
     }
 
     private val logger = LogManager.getLogger(AlertService::class.java)
+
+    suspend fun loadCurrentAlertsForWorkflow(workflow: Workflow, dataSources: DataSources): Map<Trigger, Alert?> {
+        val searchAlertsResponse: SearchResponse = searchAlerts(
+            workflow = workflow,
+            size = workflow.triggers.size * 2, // We expect there to be only a single in-progress alert so fetch 2 to check
+            dataSources = dataSources
+        )
+
+        val foundAlerts = searchAlertsResponse.hits.map { Alert.parse(contentParser(it.sourceRef), it.id, it.version) }
+            .groupBy { it.triggerId }
+        foundAlerts.values.forEach { alerts ->
+            if (alerts.size > 1) {
+                logger.warn("Found multiple alerts for same trigger: $alerts")
+            }
+        }
+
+        return workflow.triggers.associateWith { trigger ->
+            foundAlerts[trigger.id]?.firstOrNull()
+        }
+    }
 
     suspend fun loadCurrentAlertsForQueryLevelMonitor(monitor: Monitor, workflowRunContext: WorkflowRunContext?): Map<Trigger, Alert?> {
         val searchAlertsResponse: SearchResponse = searchAlerts(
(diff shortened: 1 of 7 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
```diff
@@ -257,18 +278,84 @@ class AlertService(
...
+                errorHistory = updatedHistory,
+                actionExecutionResults = updatedActionExecutionResults,
+                schemaVersion = IndexUtils.alertIndexSchemaVersion,
+            )
+        } else {
+            if (alertError == null) Alert.State.ACTIVE
+            else Alert.State.ERROR
+            Alert(
+                startTime = Instant.now(),
+                lastNotificationTime = Instant.now(),
```
sbcd90, line 350: can we use the same variable we defined in line 286?

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
```diff
@@ -758,6 +845,37 @@ class AlertService(
...
+        workflow: Workflow,
+        size: Int,
+        dataSources: DataSources,
+    ): SearchResponse {
+        val workflowId = workflow.id
+        val alertIndex = dataSources.alertsIndex
+
+        val queryBuilder = QueryBuilders.boolQuery()
+            .must(QueryBuilders.termQuery(Alert.WORKFLOW_ID_FIELD, workflowId))
+            .must(QueryBuilders.termQuery(Alert.MONITOR_ID_FIELD, ""))
```
sbcd90, line 864: can there be a doc which has both `workflowId` & `monitorId` field set?

alerting/src/main/kotlin/org/opensearch/alerting/script/ChainedAlertTriggerExecutionContext.kt
```diff
@@ -18,15 +19,17 @@ data class ChainedAlertTriggerExecutionContext(
...
+    val alert: Alert? = null
 ) {
 
     constructor(
         workflow: Workflow,
         workflowRunResult: WorkflowRunResult,
         trigger: ChainedAlertTrigger,
         alertGeneratingMonitors: Set<String>,
-        monitorIdToAlertIdsMap: Map<String, Set<String>>
+        monitorIdToAlertIdsMap: Map<String, Set<String>>,
```
sbcd90, line 31: can we make `periodStart` in the default constructor `nullable` & remove this constructor? Lot of similar fields between this one & the default one.

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
eirsep: Alert created from composite monitor trigger will have only empty `monitorId`.

The underlying alerts created in AUDIT state will have both `monitorId` and `worklfowId`.

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
eirsep: ack

sbcd90 (approved):

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
```diff
@@ -83,6 +84,26 @@ class AlertService(
...
+            workflow = workflow,
+            size = workflow.triggers.size * 2, // We expect there to be only a single in-progress alert so fetch 2 to check
+            dataSources = dataSources
+        )
+
+        val foundAlerts = searchAlertsResponse.hits.map { Alert.parse(contentParser(it.sourceRef), it.id, it.version) }
+            .groupBy { it.triggerId }
+        foundAlerts.values.forEach { alerts ->
+            if (alerts.size > 1) {
+                logger.warn("Found multiple alerts for same trigger: $alerts")
```
lezzago, line 98: should we try to resolve the older alert if this occurs?

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
```diff
@@ -257,18 +278,84 @@ class AlertService(
...
+                schemaVersion = IndexUtils.alertIndexSchemaVersion,
+            )
+        } else {
+            if (alertError == null) Alert.State.ACTIVE
+            else Alert.State.ERROR
+            Alert(
+                startTime = Instant.now(),
+                lastNotificationTime = currentTime,
+                state = Alert.State.ACTIVE,
+                errorMessage = null, schemaVersion = -1,
```
lezzago, line 352: Should schema version be: `IndexUtils.alertIndexSchemaVersion` instead of `-1`?

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
```diff
@@ -758,6 +845,37 @@ class AlertService(
         return searchResponse
     }
 
+    /**
+     * Searches for Alerts in the monitor's alertIndex.
```
lezzago, line 849: Lets call out here that this will not retrieve audit alerts.

alerting/src/main/kotlin/org/opensearch/alerting/script/ChainedAlertTriggerExecutionContext.kt
```diff
@@ -18,27 +19,10 @@ data class ChainedAlertTriggerExecutionContext(
...
-        this(
-            workflow,
-            workflowRunResult,
-            workflowRunResult.executionStartTime,
-            workflowRunResult.executionEndTime,
-            workflowRunResult.error,
-            trigger,
-            alertGeneratingMonitors,
-            monitorIdToAlertIdsMap
-        )
```
lezzago, line 40: we dont need the constructor anymore?

alerting/src/main/kotlin/org/opensearch/alerting/workflow/CompositeWorkflowRunner.kt
```diff
@@ -150,36 +150,69 @@ object CompositeWorkflowRunner : WorkflowRunner() {
...
+            monitorCtx.alertService!!.loadCurrentAlertsForWorkflow(workflow, dataSources)
+        } catch (e: Exception) {
+            logger.error("Failed to fetch current alerts for workflow", e)
+            // We can't save ERROR alerts to the index here as we don't know if there are existing ACTIVE alerts
+            val id = if (workflow.id.trim().isEmpty()) "_na_" else workflow.id
+            logger.error("Error loading alerts for workflow: $id", e)
+            return workflowRunResult.copy(error = e)
+        }
+        try {
+            monitorCtx.alertIndices!!.createOrUpdateAlertIndex(dataSources)
```
lezzago, line 165: Do we need this since we do this on line 154?

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
lezzago: this can be a follow up. But if this is not expected, we can ignore this unless we see issues for this.

alerting/src/main/kotlin/org/opensearch/alerting/workflow/CompositeWorkflowRunner.kt
eirsep: It's not necessary prolly, but I think this would be consistent practice to all other monitors where we ensure alertIndex is present before we fetch current alert.

lezzago (approved):

AWSHurneyt (approved):

PR #1088, opened 2023-08-15
upgrade transitive dependency com.google.guava:guava:32.1.2-jre for CVE-2023-2976

eirsep:
https://nvd.nist.gov/vuln/detail/CVE-2023-2976 recommends using google.guava v32+

we have a transitive dependency from com.google.googlejavaformat:google-java-format:1.10.0. upgrading it won't help as all versions of googlejavaformat are currently using guava versions mentioned in CVE.


lezzago (changes requested):
Needs to be against main and then backported to 2.x

eirsep:
> Needs to be against main and then backported to 2.x

updated to main

PR #1097, opened 2023-08-17
optimize doc-level monitor workflow for index patterns

sbcd90:
*Description of changes:*
This pr addresses following performance problems in `doc-level monitor execution workflow`.
- In the current workflow, if the `doc-level monitor` is monitoring an `index pattern` & a new index is introduced which matches the pattern, the `doc-level monitor` duplicates all the field mappings & queries for that index. This is reproducible using integ test https://github.com/opensearch-project/alerting/pull/1097/files#diff-9b08d53e8cff3d739beb6ba304e4eaf466f08412762c3cd55886a52b722a77f1R503

Now, say, we have `1000` queries & each concrete index behind the index-pattern has 1000 field mappings. Also, lets assume `1 concrete index` is generated everyday. We also know the `default number of field mappings` an index can have is `1000` & today if the no. of field mappings go over `1000` in the query index, we rollover the query index. 

This would mean, we create a new rollover query index everyday & keep on ingesting 1000 queries in it everyday. In 30 days, we create `30 indices`(which means by default `1 primary & 1 replica shards` per index) which contains `30000` duplicate queries.
This causes the `data nodes` to get full resulting in cluster crash.
https://github.com/opensearch-project/security-analytics/issues/509
https://forum.opensearch.org/t/security-analytics-error/14639/11

We do not notice this problem for small no. of queries but duplication of queries piles up over a period of time.
This pr addresses this issue by continously updating `1 set of queries` for all the `concrete indices` belonging to an `index-pattern`. this provides huge storage optimization.

- In the current workflow, for every `concrete index`, we make an `update mapping` api call in no particular order. So, say, index `test1` has fields `f1 & f2` & `test2` has field `f4` & both of them match pattern `test*`, if we make first `update mapping` call for `test1`, then query index gets `f1 & f2` but in the next `update mapping` call for `test2` we completely overwrite it with `f4`. This is reproducible using integ test https://github.com/opensearch-project/alerting/pull/1097/files#diff-9b08d53e8cff3d739beb6ba304e4eaf466f08412762c3cd55886a52b722a77f1R551

This pr addresses this issue by first collecting field mappings of all concrete indices belonging to an index-pattern together, & then making a `single call to update mappings api`.

- In the current workflow, for every `concrete index`, we make an `update mapping` api call in no particular order. So, if there are `100` concrete indices behind an `index-pattern` we make `100 update mapping api` calls.

This pr optimizes the time complexity by making a `single call to update mappings api` for each `index pattern`.

- Here is how the optimization looks like
Screenshot 2023-08-29 at 2 08 52 PM

*CheckList:*
- [X] Commits are signed per the DCO using --signoff

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
@@ -51,6 +51,7 @@ import org.opensearch.core.common.bytes.BytesReference
 import org.opensearch.core.rest.RestStatus
 import org.opensearch.core.xcontent.ToXContent
 import org.opensearch.core.xcontent.XContentBuilder
+import org.opensearch.index.IndexNotFoundException
 import org.opensearch.index.query.BoolQueryBuilder
 import org.opensearch.index.query.Operator
 import org.opensearch.index.query.QueryBuilders
@@ -118,11 +119,15 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
 
         try {
             // Resolve all passed indices to concrete indices
-            val indices = IndexUtils.resolveAllIndices(
+            val concreteIndices = IndexUtils.resolveAllIndices(
                 docLevelMonitorInput.indices,
                 monitorCtx.clusterService!!,
                 monitorCtx.indexNameExpressionResolver!!
             )
+            if (concreteIndices.isEmpty()) {
+                logger.error("indices not found-${docLevelMonitorInput.indices.joinToString(",")}")
+                throw IndexNotFoundException(docLevelMonitorInput.indices.joinToString(","))
+            }
 
             monitorCtx.docLevelMonitorQueries!!.initDocLevelQueryIndex(monitor.dataSources)
             monitorCtx.docLevelMonitorQueries!!.indexDocLevelQueries(
@@ -134,7 +139,7 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
 
             // cleanup old indices that are not monitored anymore from the same monitor
             for (ind in updatedLastRunContext.keys) {
-                if (!indices.contains(ind)) {
+                if (!concreteIndices.contains(ind)) {
                     updatedLastRunContext.remove(ind)
                 }
             }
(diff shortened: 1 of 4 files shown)


lezzago (approved):

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -134,73 +138,84 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
-                    // TODO: If dryrun, we should make it so we limit the search as this could still potentially give us lots of data
-                    if (isTempMonitor) {
-                        indexLastRunContext[shard] = max(-1, (indexUpdatedRunContext[shard] as String).toInt() - 10)
+            docLevelMonitorInput.indices.forEach { indexName ->
+                val concreteIndices = IndexUtils.resolveAllIndices(
+                    listOf(indexName),
+                    monitorCtx.clusterService!!,
+                    monitorCtx.indexNameExpressionResolver!!
+                )
+                val updatedIndexName = indexName.replace("*", "_")
```
eirsep, line 155: why do we do this indexName.replace("*", "_")?

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -118,11 +119,14 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
         try {
             // Resolve all passed indices to concrete indices
-            val indices = IndexUtils.resolveAllIndices(
+            val concreteIndices = IndexUtils.resolveAllIndices(
                 docLevelMonitorInput.indices,
                 monitorCtx.clusterService!!,
                 monitorCtx.indexNameExpressionResolver!!
             )
+            if (concreteIndices.isEmpty()) {
+                throw IndexNotFoundException(docLevelMonitorInput.indices[0])
```
eirsep, line 128: 1. can we add error log with monitor Id and indices list which were not found
2. in exception why are we logging only 0th index element of `indices` variable? can we log all elements in array/list

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -134,73 +138,84 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
-
-                val count: Int = indexLastRunContext["shards_count"] as Int
-                for (i: Int in 0 until count) {
-                    val shard = i.toString()
-
-                    // update lastRunContext if its a temp monitor as we only want to view the last bit of data then
-                    // TODO: If dryrun, we should make it so we limit the search as this could still potentially give us lots of data
-                    if (isTempMonitor) {
-                        indexLastRunContext[shard] = max(-1, (indexUpdatedRunContext[shard] as String).toInt() - 10)
+            docLevelMonitorInput.indices.forEach { indexName ->
```
eirsep, line 149: why did we change this list being looped from computed concrete indices list to the monitor indices list?

eirsep:
To create query index mappings for an index pattern would be a very big optimization but it works only under the assumption that an index pattern test* resolves to indices with same mappings.

this fails in the following scenario:
indices `test1` and `test2` are resolved from `test*` pattern. test1 has a field named `f1` of type `long` and test2 also has a field named `f1` BUT with type `keyword`

when we store the field mapping in query index for `f1`_`test*` this would break the query execution due to incorrect mapping

eirsep:
we can add this optimization for the scenarios where we can identify all indices covered in index pattern are guaranteed to have the same mappings i.e. data streams, rolling indices with index templates.


should we store a field in index mapping called `index_template` if we are creating index from index template? food for throught

eirsep:
> To create query index mappings for an index pattern would be a very big optimization but it works only under the assumption that an index pattern test* resolves to indices that do not have fields with same names but different data types.
> 
> this fails in the following scenario: indices `test1` and `test2` are resolved from `test*` pattern. test1 has a field named `f1` of type `long` and test2 also has a field named `f1` BUT with type `keyword`
> 
> when we store the field mapping in query index for `f1`_`test*` this would break the query execution due to incorrect mapping

Since we are batching all mappings can we check if the above mentioned field mapping mismatch is happening for any fields and then ingest one for the generic test* and one mapping for the specific index. Similarly while fetching mappings look for both index mappings with test* and test1. If latter is found prefer it. if not found, use the former.

sbcd90:
> @sbcd90
> 
> > To create query index mappings for an index pattern would be a very big optimization but it works only under the assumption that an index pattern test* resolves to indices that do not have fields with same names but different data types.
> > this fails in the following scenario: indices `test1` and `test2` are resolved from `test*` pattern. test1 has a field named `f1` of type `long` and test2 also has a field named `f1` BUT with type `keyword`
> > when we store the field mapping in query index for `f1`_`test*` this would break the query execution due to incorrect mapping
> 
> Since we are batching all mappings can we check if the above mentioned field mapping mismatch is happening for any fields and then ingest one for the generic test* and one mapping for the specific index. Similarly while fetching mappings look for both index mappings with test* and test1. If latter is found prefer it. if not found, use the former.

hi @eirsep , that is the exact fix i'm working on.

sbcd90:
hi @eirsep , just addressed the comment with the test case https://github.com/opensearch-project/alerting/pull/1097/files#diff-9b08d53e8cff3d739beb6ba304e4eaf466f08412762c3cd55886a52b722a77f1R503.
Can you please look into the pr again?

alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt
```diff
@@ -378,25 +461,87 @@ class DocLevelMonitorQueries(private val client: Client, private val clusterServ
...
+        }
+        return conflictingFields
+    }
+
     /**
      * Adjusts max field limit index setting for query index if source index has higher limit.
      * This will prevent max field limit exception, when source index has more fields then query index limit
      */
-    private suspend fun checkAndAdjustMaxFieldLimit(sourceIndex: String, concreteQueryIndex: String) {
+    private suspend fun checkMaxFieldLimit(sourceIndex: String): Long {
```
eirsep, line 524: the java doc description for this method seems like it should describe `adjustMaxFieldLimitForQueryIndex`

eirsep: can we update the code comments for this method and below method

alerting/src/test/kotlin/org/opensearch/alerting/DocumentMonitorRunnerIT.kt
```diff
@@ -495,6 +500,375 @@ class DocumentMonitorRunnerIT : AlertingRestTestCase() {
...
+        deleteIndex(ALL_ALERT_INDEX_PATTERN)
+
+        indexDoc(testIndex, "2", testDoc)
+        val testIndex2 = createTestIndex("test2")
+        testDoc = """{
+            "source" : { "device": {"port" : "12345" } },
+            "nested_field": { "test1": "some text" },
+            "test_field": "12345"
+        }"""
+        indexDoc(testIndex2, "1", testDoc)
```
eirsep, line 567: NIT; since you are not explicitly creating index with different mappings 
can you plz fetch second index's mapping and assert that the field type is actually different for `test_field` so as to validate the test scenario. that would make the test more readable. right now it's a bit hard to understand.

alerting/src/test/kotlin/org/opensearch/alerting/DocumentMonitorRunnerIT.kt
```diff
@@ -495,6 +500,375 @@ class DocumentMonitorRunnerIT : AlertingRestTestCase() {
         assertEquals("Findings saved for test monitor expected 14, 51 and 10", 3, foundFindings.size)
     }
 
+    fun `test execute monitor with indices having fields with same name but different data types`() {
```
eirsep, line 503: can we plz add asserts on the expected mappings and content of the query index. that would be the essence of these tests and help us visualize the changes made in this PR

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -717,17 +738,36 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
      */
     private fun transformDocumentFieldNames(
```
eirsep, line 739: this is a very critical piece fo code. can we update java docs description of this method incorporating the new change?

alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt
```diff
@@ -389,6 +461,62 @@ class DocLevelMonitorQueries(private val client: Client, private val clusterServ
...
+            if (!oldField.containsKey(it.key)) {
+                mergedField[it.key] = it.value
+            }
+        }
+        return mergedField
+    }
+
+    fun getAllConflictingFields(clusterState: ClusterState, concreteIndices: List<String>): Set<String> {
+        val conflictingFields = mutableSetOf<String>()
+        val allFlattenPaths = mutableMapOf<String, MutableMap<String, Any>>()
```
eirsep, line 489: does the below flatten logic handle both the nested and non-nested type objects correctly?

what would be the difference after flattening into string in the following 2 cases 
```
"nested_field": {
                      "type": "nested",
                      "properties": {
                        "test1": {
                          "type": "keyword"
                        }
                      }
                      
```

and 


```
     "nested_field": {
                      
                      "properties": {
                        "test1": {
                          "type": "keyword"
                        }
                      }

```

alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt
```diff
@@ -389,6 +461,62 @@ class DocLevelMonitorQueries(private val client: Client, private val clusterServ
...
+                if (indexMetadata.mapping()?.sourceAsMap?.get("properties") != null) {
+                    val properties = (
+                        (indexMetadata.mapping()?.sourceAsMap?.get("properties"))
+                            as MutableMap<String, Any>
+                        )
+                    // Node processor function is used to process leaves of index mappings tree
+                    //
+                    val leafNodeProcessor =
+                        fun(fieldName: String, _: String, props: MutableMap<String, Any>): Triple<String, String, MutableMap<String, Any>> {
+                            return Triple(fieldName, fieldName, props)
```
eirsep, line 502: why do we need field name twice?
can you add some code comments around what is the Triple type object being returned to make it more readable?

eirsep (changes requested):
thanks for making these changes @sbcd90 

i have some minor comments not related to logic but more around readiblity and testsing.

Kindly also respond to earlier comments.

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
sbcd90: this is because we want to expand `1 index pattern` & process all its `concrete indices` separately. Previous logic was we expand `all index patterns` & process all the `concrete indices` one by one.

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
sbcd90: this is because `query string query fields` with `*` have a different meaning & is also not allowed in some cases.

alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt
sbcd90: this is used intentionally because `leafNodeProcessor` is used at another place where the 2nd param is actually used to pass `modified field name`.

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
sbcd90: changed it now. it was copied from eariler logic.

alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt
sbcd90: updated it.

alerting/src/test/kotlin/org/opensearch/alerting/DocumentMonitorRunnerIT.kt
sbcd90: specifying explicit mappings now.

alerting/src/test/kotlin/org/opensearch/alerting/DocumentMonitorRunnerIT.kt
sbcd90: asserts added on actual queries stored in query index.

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
sbcd90: updated the documentation for the method.

alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt
sbcd90: `nested fields` dont work with `query string queries`. https://stackoverflow.com/questions/69857071/how-can-i-use-query-string-to-match-both-nested-and-non-nested-fields-at-the-sam
but if we pair them with an index containing an object field, it will work. https://github.com/opensearch-project/alerting/pull/1097/files#diff-9b08d53e8cff3d739beb6ba304e4eaf466f08412762c3cd55886a52b722a77f1R636

eirsep (approved):

PR #1103, opened 2023-08-23
Update BWC version to fix BWC tests

lezzago:
*Issue #, if available:*
N/A

*Description of changes:*
Update the BWC version to fix the BWC tests as `mixedClusterTask` and `rollingUpgradeClusterTask` needs the previous major version to be the latest (eg. Needs to be 2.10 vs 2.6)

*CheckList:*

### DIFF
--- a/.github/workflows/test-workflow.yml
+++ b/.github/workflows/test-workflow.yml
@@ -42,7 +42,7 @@ jobs:
           java-version: ${{ matrix.java }}
       - name: Build and run with Gradle
         working-directory: ${{ env.WORKING_DIR }}
-        run: ./gradlew gradlew assemble integTest ${{ env.BUILD_ARGS }}
+        run: ./gradlew assemble integTest ${{ env.BUILD_ARGS }}
         env:
           _JAVA_OPTIONS: ${{ matrix.os_java_options }}
       - name: Create Artifact Path
--- a/.github/workflows/test-workflow.yml
+++ b/.github/workflows/test-workflow.yml
@@ -42,7 +42,7 @@ jobs:
           java-version: ${{ matrix.java }}
       - name: Build and run with Gradle
         working-directory: ${{ env.WORKING_DIR }}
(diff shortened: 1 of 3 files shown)


.github/workflows/test-workflow.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           java-version: ${{ matrix.java }}
       - name: Build and run with Gradle
         working-directory: ${{ env.WORKING_DIR }}
-        run: ./gradlew build ${{ env.BUILD_ARGS }} -x alertingBwcCluster#fullRestartClusterTask -x alertingBwcCluster#mixedClusterTask -x alertingBwcCluster#oldVersionClusterTask0 -x alertingBwcCluster#oldVersionClusterTask1 -x alertingBwcCluster#rollingUpgradeClusterTask -x alertingBwcCluster#twoThirdsUpgradedClusterTask
+        run: ./gradlew assemble integTest ${{ env.BUILD_ARGS }}
```
eirsep, line 45: NIT: why do we need to do assemble before running integTest?

lezzago: to ensure it can build the artifact still. We had this in 2.x, so wanted to keep it consistent

alerting/build.gradle
```diff
@@ -16,7 +16,7 @@ apply plugin: 'jacoco'
 
 def usingRemoteCluster = System.properties.containsKey('tests.rest.cluster') || System.properties.containsKey('tests.cluster')
 def usingMultiNode = project.properties.containsKey('numNodes')
-String bwcVersion = "2.6.0.0"
+String bwcVersion = "2.10.0.0"
```
eirsep, line 19: can we set this dynamically by building a new string from "opensearch_version"

eirsep: api "org.opensearch.plugin:percolator-client:${opensearch_version}" is using opensearch_version in this file.. so we can use that itself i guess by removing -SNAPSHOT if required. plz check

lezzago: No since this is the BWC version, not the current opensearch version. We wont know the latest minor version (eg. 2.10) from the last major version (eg. 2.x).

eirsep (approved):

AWSHurneyt (approved):

Issue #974, opened 2023-06-29
Compatibility with segment replication


dreamer-89:
### Summary
With 2.9.0 release, there are lot of enhancements going in for segment replication[1][2] feature (went GA in 2.7.0), we need to ensure different plugins are compatible with current state of this feature. Previously, we ran [tests](https://github.com/opensearch-project/OpenSearch/issues/6602) on plugin repos to verify this compatibility but want plugin owners to be aware of these changes so that required updates (if any) can be made. With `2.10.0` release, [remote store feature](https://github.com/opensearch-project/OpenSearch/issues/2992) is going GA which internally uses SEGMENT replication strategy only i.e. it enforces all indices to use `SEGMENT` replication strategy. So, it is important to validate plugins are compatible with segment replication feature. 

#### 1. Refresh policy behavior
1. RefreshPolicy.IMMEDIATE will only refresh primary shards but not replica shards immediately. Instead post refresh, primary will start  a round of segment replication to update the replica shard copies leading to eventual consistency.
2. RefreshPolicy.WAIT_UNTIL ensures the indexing operation is searchable in your cluster i.e. RAW (Read after write guarantee). With segment replication, this guarantee is not promised due to delay in replica shared updates from asynchronous background refreshes.

####  2. Refresh lag on replicas
With segment replication, there is inherent delay in documents to be searchable on replica shard copies. This is due to the fact that replica shard copies over data (segment) files from primary. Thus, compared to document replication, there will be on average increase in amount of time the replica shards are consistent with primaries.

#### 3. System/hidden indices support
With https://github.com/opensearch-project/OpenSearch/pull/8200, system and hidden indices are now supported with `SEGMENT` replication strategy. We need to ensure there are no bottlenecks which prevents system/hidden indices with segment replication.

### Next steps
With segment replication strong reads are not guaranteed. Thus, if the plugin needs strong reads guarantees specially as alternative to change in behavior of refresh policy and lag on replicas (point 1 and 2 above), we need to update search requests to target primary shard only. With https://github.com/opensearch-project/OpenSearch/pull/7375, core now supports primary shards only based search. Please follow [documentation](https://opensearch.org/docs/latest/api-reference/search/#the-preference-query-parameter) for examples and details

### Open questions
In case of any questions or issues, please post it in [core](https://github.com/opensearch-project/OpenSearch/issues/8211) issue

### Reference
[1] [Design](https://github.com/opensearch-project/OpenSearch/issues/2229)

[...]

PR #1114, opened 2023-09-05
Changes in workflows for seg rep compatibility

eirsep:
In workflows, we save audit state alerts from delegate monitors.

Within the same workflow execution we execute a search request to fetch the persisted audit alerts from alerts system index.

Since RAW guarantees have changed with seg rep, we would need to search on primaries to ensure we have the latest information.

Relates to [Compatibility with segment replication #974
](https://github.com/opensearch-project/alerting/issues/974)

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/workflow/CompositeWorkflowRunner.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/workflow/CompositeWorkflowRunner.kt
@@ -21,6 +21,7 @@ import org.opensearch.alerting.script.ChainedAlertTriggerExecutionContext
 import org.opensearch.alerting.util.AlertingException
 import org.opensearch.alerting.util.isDocLevelMonitor
 import org.opensearch.alerting.util.isQueryLevelMonitor
+import org.opensearch.cluster.routing.Preference
 import org.opensearch.common.xcontent.LoggingDeprecationHandler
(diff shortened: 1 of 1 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/workflow/CompositeWorkflowRunner.kt
```diff
@@ -226,11 +227,6 @@ object CompositeWorkflowRunner : WorkflowRunner() {
             )
         }
         workflowRunResult.executionEndTime = Instant.now()
-
-        val sr = SearchRequest(dataSources!!.alertsIndex)
-        sr.source().query(QueryBuilders.matchAllQuery()).size(10)
-        val searchResponse: SearchResponse = monitorCtx.client!!.suspendUntil { monitorCtx.client!!.search(sr, it) }
-        searchResponse.hits
```
lezzago, line 233: Was this dev code that was accidentally left over?

lezzago (approved):

alerting/src/main/kotlin/org/opensearch/alerting/workflow/CompositeWorkflowRunner.kt
eirsep: yes

sbcd90 (approved):

PR #1115, opened 2023-09-06
Updates demo certs used in integ tests

DarshitChanpura:
Fixes build breaks when running integTests

*CheckList:*


peternied:
This is a hack of a fix.  These repositories should not have an artifact of a security principal checked in locally, instead getting the cert / jks / pem from the deployed instance of the security plugins demo configuration. While this unblocks the teams today, this is the wrong way to solve this problem, this defers the issue where the certificates are regenerated next time.

Why can't these repositories use the artifacts from security plugin demo configuration?

lezzago (approved):

AWSHurneyt (approved):

PR #1139, opened 2023-09-13
add workflow null or empty check only when empty workflow id passed.

eirsep:
add workflow null or empty check only when empty workflow id passed in transport get alerts request.

In Rest get alert action,  add empty string singleton list for workflow ids when none is passed. This is to support returning non-Audit alerts on Alert Dashboard.

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
@@ -115,7 +115,9 @@ class TransportGetAlertsAction @Inject constructor(
 
         if (getAlertsRequest.monitorId != null) {
             queryBuilder.filter(QueryBuilders.termQuery("monitor_id", getAlertsRequest.monitorId))
-            if (getAlertsRequest.workflowIds.isNullOrEmpty()) {
+            if (
+                getAlertsRequest.workflowIds != null && getAlertsRequest.workflowIds!!.size == 1 && getAlertsRequest.workflowIds!![0] == ""
+            ) {
                 val noWorkflowIdQuery = QueryBuilders.boolQuery()
                     .should(QueryBuilders.boolQuery().mustNot(QueryBuilders.existsQuery(Alert.WORKFLOW_ID_FIELD)))
                     .should(QueryBuilders.termsQuery(Alert.WORKFLOW_ID_FIELD, ""))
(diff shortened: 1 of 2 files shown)


eirsep:
I feel this is a more appropriate fix. There shouldn't be a change necessary in security analytics as we are not making any incorrect calls to alerting


Closing this PR in favor of https://github.com/opensearch-project/alerting/pull/1139

amsiglan (approved):

lezzago (approved):

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
```diff
@@ -115,22 +115,29 @@ class TransportGetAlertsAction @Inject constructor(
 
         if (getAlertsRequest.monitorId != null) {
             queryBuilder.filter(QueryBuilders.termQuery("monitor_id", getAlertsRequest.monitorId))
-            if (getAlertsRequest.workflowIds.isNullOrEmpty()) {
+            if (
+                getAlertsRequest.workflowIds != null && getAlertsRequest.workflowIds!!.size == 1 && getAlertsRequest.workflowIds!![0] == ""
```
AWSHurneyt, line 119: Nitpick: Looks like this check is duplicated in a few spots. It could be helpful to extract it into a helper method that has a doc comment explaining it further.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
```diff
@@ -115,22 +115,29 @@ class TransportGetAlertsAction @Inject constructor(
...
                 val noWorkflowIdQuery = QueryBuilders.boolQuery()
                     .should(QueryBuilders.boolQuery().mustNot(QueryBuilders.existsQuery(Alert.WORKFLOW_ID_FIELD)))
                     .should(QueryBuilders.termsQuery(Alert.WORKFLOW_ID_FIELD, ""))
                 queryBuilder.must(noWorkflowIdQuery)
             }
         }
-        if (getAlertsRequest.workflowIds.isNullOrEmpty() == false) {
+        if (
+            getAlertsRequest.workflowIds.isNullOrEmpty() == false &&
+            !(getAlertsRequest.workflowIds!!.size == 1 && getAlertsRequest.workflowIds!![0] == "")
```
AWSHurneyt, line 139: Nitpick: Would the `!(getAlertsRequest.workflowIds!!.size == 1` portion of this check be redundant given the `isNullOrEmpty` call?

AWSHurneyt (approved):

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
eirsep: it would not be redundant because it is checking that the size should be exactly 1.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetAlertsAction.kt
eirsep: extracted into method and added comment

Issue #1028, opened 2023-07-13
[FEATURE] Use waitUntil in all tests vs thread.sleep


lezzago:
Currently in many of the tests for the Alerting plugin, we have `thread.sleep` to wait for background processes to complete for us to validate the tests. These should be replaced with `OpenSearchTestCase.waitUntil`

JacobCho-i:
Hi, I would like to work on this issue.

PR #1242, opened 2023-10-10
change Thread.sleep to waitUntil function under test files

JacobCho-i:
*Issue #, if available:*
#1028 

*Description of changes:*
replaced Thread.sleep functions with waitUntil function in tests under src/test files 

*CheckList:*


alerting/src/test/kotlin/org/opensearch/alerting/resthandler/MonitorRestApiIT.kt
```diff
@@ -956,10 +966,13 @@ class MonitorRestApiIT : AlertingRestTestCase() {
 
     fun `test monitor stats when disabling and re-enabling scheduled jobs with existing monitor`() {
         // Enable Monitor jobs
-        enableScheduledJob()
+
+        var response = enableScheduledJob()
```
AWSHurneyt, line 970: Could you explain how this is being used in the `waitUntil`s? 

My understanding is that `waitUntil` essentially performs the `booleanSupplier` check periodically. If the check is `true`, or the `maxWaitTime` has been met, it will proceed past the `waitUntil` block; if `false`, it will continue performing the check.

Since we're using the `waitUntil` to force a pause in the test execution in this context, would it make sense to just set the `breakSupplier` to `false` that way the block always waits for the timeout?
E.g.,
```
if (isMultiNode) OpenSearchTestCase.waitUntil({
            return@waitUntil false
        }, 2, TimeUnit.SECONDS)
```
I mostly curious if this is an option because I'm not sure using `enableScheduledJob` in this way will do what we want. The constructor of that function doesn't suggest it returns a nullable `Response`. So I'm not sure that object will ever be `null`. Also, `response` doesn't get set back to `null` after line `975`, so the `response` that's checked by the subsequent `waitUntil`s probably won't be `null` either.

JacobCho-i: I think I had some misunderstanding in enableScheduledJob(). The main purpose of setting a response variable is to know if the PUT request in enableScheduledJob() went through all the process before proceeding to the next line. So, my assumption was that the response will be null until the request is made, which will set it to non-null variable and change `booleanSuppler` so that it can proceed afterwards. I will make a change that you suggested to other tests that I can applied to. Thanks!

AWSHurneyt (approved):

alerting/src/test/kotlin/org/opensearch/alerting/bwc/AlertingBackwardsCompatibilityIT.kt
```diff
@@ -69,7 +71,9 @@ class AlertingBackwardsCompatibilityIT : AlertingRestTestCase() {
                     //  the test execution by a lot (might have to wait for Job Scheduler plugin integration first)
                     // Waiting a minute to ensure the Monitor ran again at least once before checking if the job is running
                     // on time
-                    Thread.sleep(60000)
+                    OpenSearchTestCase.waitUntil({
+                        return@waitUntil false
```
lezzago, line 75: Instead of false, we would want to `verifyMonitorStats`. That way we get the benefit of the waitUtil function

lezzago (approved):

AWSHurneyt:
@JacobCho-i Looks like the automatic backport failed for this PR. Could you checkout a fork of the 2.x branch, cherry-pick your changes to it, and create another PR directly to the 2.x branch? There were likely some merge conflicts that occurred that caused the backport to fail.

AWSHurneyt:
> @JacobCho-i Looks like the automatic backport failed for this PR. Could you checkout a fork of the 2.x branch, cherry-pick your changes to it, and create another PR directly to the 2.x branch? There were likely some merge conflicts that occurred that caused the backport to fail.

Manual backport PR https://github.com/opensearch-project/alerting/pull/1301

Issue #1290, opened 2023-10-27
Optimize doc level monitors -  only query indices with new documents in data streams


eirsep:
Doc level monitors intend to track sequence number of indices shards and query only documents ingested since last monitor execution. The problem is that users configure doc level monitors to run against index patterns with rollover or data streams.
More often than not doc level monitors are querying an index patter which resolves to , say 1000 concrete indices, of which 999 are read-only and 1 write index. 
We really don't need to query all the 999 read-only indices as they might not have any documents ingested since the previous execution.

**we can optimize to query only those indices which are writeable and all indices which have been marked as read-only AFTER the previous monitor execution**
**for first run of monitor, when creating doc level queries it should only consider write index and the read indices which were created after the enabled_time of monitor** 

This will drastically reduce the CPU consumption of the doc level monitor and provide a significant optimization for the plugin's resource consumption

PR #1302, opened 2023-11-11
optimize doc-level monitor execution workflow for datastreams

sbcd90:
*Issue #, if available:*
#1290 

*Description of changes:*
This pr optimizes the doc-level monitor execution workflow for data streams.

The optimization proposed only considers the `current write index, any new read-only indices created after last monitor execution run & the last write index of last monitor execution run` for current monitor execution run.
It ignores `read-only indices` that may have been created prior to the last monitor execution run.
 
The below diagram shows the difference.

**Earlier**
Screenshot 2023-11-10 at 6 00 48 PM

**Now**
Screenshot 2023-11-10 at 5 58 30 PM

*CheckList:*
- [X] Commits are signed per the DCO using --signoff

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt
@@ -207,11 +207,25 @@ class DocLevelMonitorQueries(private val client: Client, private val clusterServ
 
         // Run through each backing index and apply appropriate mappings to query index
         indices.forEach { indexName ->
-            val concreteIndices = IndexUtils.resolveAllIndices(
+            var concreteIndices = IndexUtils.resolveAllIndices(
                 listOf(indexName),
                 monitorCtx.clusterService!!,
                 monitorCtx.indexNameExpressionResolver!!
             )
+            if (IndexUtils.isAlias(indexName, monitorCtx.clusterService!!.state()) ||
+                IndexUtils.isDataStream(indexName, monitorCtx.clusterService!!.state())
+            ) {
+                val lastWriteIndex = concreteIndices.find { monitorMetadata.lastRunContext.containsKey(it) }
+                if (lastWriteIndex != null) {
+                    val lastWriteIndexCreationDate =
+                        IndexUtils.getCreationDateForIndex(lastWriteIndex, monitorCtx.clusterService!!.state())
+                    concreteIndices = IndexUtils.getNewestIndicesByCreationDate(
+                        concreteIndices,
+                        monitorCtx.clusterService!!.state(),
+                        lastWriteIndexCreationDate
+                    )
+                }
+            }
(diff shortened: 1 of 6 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt
```diff
@@ -207,11 +207,25 @@ class DocLevelMonitorQueries(private val client: Client, private val clusterServ
 
         // Run through each backing index and apply appropriate mappings to query index
         indices.forEach { indexName ->
-            val concreteIndices = IndexUtils.resolveAllIndices(
+            var concreteIndices = IndexUtils.resolveAllIndices(
                 listOf(indexName),
                 monitorCtx.clusterService!!,
                 monitorCtx.indexNameExpressionResolver!!
             )
+            if (IndexUtils.isAlias(indexName, monitorCtx.clusterService!!.state()) ||
```
eirsep, line 215: if I use any alias, there can be multiple writeIndices. why are we assuming there can be only one writeIndex. That applies to datastream only

plz add tests for alias which points to multiple write indices and modify handling as required. 

I would explore modifying monitor metadata as required to handle aliases with multiple write indices.

eirsep:
plz make change to handle indices queried if the monitor metadata section is empty. Only write index and indices created after creation_time timestamp should be considered for querying .

alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt
sbcd90: An `index/datastream alias` can have only `1 write index`. https://www.elastic.co/guide/en/elasticsearch/reference/current/aliases.html#write-index
`Common Problems` section - https://opster.com/guides/elasticsearch/glossary/elasticsearch-alias/
An `index pattern` can have multiple write indices though.

eirsep (commented):
if they have 2 data streams a1* and a2* but customer creates index on a* index pattern, there will be multiple write indices for one input in `indices` section of monitor config. the changes in new PR might break the experience for this scenario
it’s ok if PR doesn’t optimize for that scenario but should not cause new bug

sbcd90:
> if they have 2 data streams a1* and a2* but customer creates index on a* index pattern, there will be multiple write indices for one input in `indices` section of monitor config. the changes in new PR might break the experience for this scenario it’s ok if PR doesn’t optimize for that scenario but should not cause new bug

the optimization doesnt add a new bug. https://github.com/opensearch-project/alerting/blob/510e66acfb7fc30248097e77672a4a02f9081e61/alerting/src/test/kotlin/org/opensearch/alerting/DocumentMonitorRunnerIT.kt#L1398 
a follow-up pr need to be created for using this optimization in index patterns.

eirsep (approved):

praveensameneni (approved):

PR #1331, opened 2023-12-20
Percolate query optimization: Perform percolate query once for each shard's documents

eirsep:
*Issue #, if available:*

*Description of changes:*
This feature optimizes scalability of percolate query performed in doc level monitors.
Before this change the behaviour was to query and collect data from all shards of each index and perform a percolate query on the docs.
This was not scaling as the data node executing the doc level would exceed heap memory limits when docs held in memory from all shards in the index exceed memory.

After the change, operation of [fetching latest docs from a shard + percolate queries] are still performed synchronously and NOT in parallel because search threadpool threads may all get used up by the monitor (which is a background job contested by other monitors, user's own search workloads)

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
@@ -57,6 +57,7 @@ import org.opensearch.index.query.BoolQueryBuilder
 import org.opensearch.index.query.Operator
 import org.opensearch.index.query.QueryBuilders
 import org.opensearch.percolator.PercolateQueryBuilderExt
+import org.opensearch.search.SearchHit
 import org.opensearch.search.SearchHits
 import org.opensearch.search.builder.SearchSourceBuilder
 import org.opensearch.search.sort.SortOrder
@@ -219,39 +220,18 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
                     // Prepare DocumentExecutionContext for each index
                     val docExecutionContext = DocumentExecutionContext(queries, indexLastRunContext, indexUpdatedRunContext)
 
-                    val matchingDocs = getMatchingDocs(
+                    fetchDataAndExecutePercolateQueriesPerShard(
                         monitor,
                         monitorCtx,
                         docExecutionContext,
                         updatedIndexName,
                         concreteIndexName,
                         conflictingFields.toList(),
-                        matchingDocIdsPerIndex?.get(concreteIndexName)
+                        matchingDocIdsPerIndex?.get(concreteIndexName),
+                        monitorMetadata,
+                        inputRunResults,
+                        docsToQueries
                     )
-
-                    if (matchingDocs.isNotEmpty()) {
-                        val matchedQueriesForDocs = getMatchedQueries(
-                            monitorCtx,
-                            matchingDocs.map { it.second },
-                            monitor,
-                            monitorMetadata,
-                            updatedIndexName,
-                            concreteIndexName
-                        )
-
-                        matchedQueriesForDocs.forEach { hit ->
-                            val id = hit.id
-                                .replace("_${updatedIndexName}_${monitor.id}", "")
-                                .replace("_${concreteIndexName}_${monitor.id}", "")
-
-                            val docIndices = hit.field("_percolator_document_slot").values.map { it.toString().toInt() }
-                            docIndices.forEach { idx ->
-                                val docIndex = "${matchingDocs[idx].first}|$concreteIndexName"
-                                inputRunResults.getOrPut(id) { mutableSetOf() }.add(docIndex)
-                                docsToQueries.getOrPut(docIndex) { mutableListOf() }.add(id)
-                            }
-                        }
-                    }
                 }
             }
             monitorResult = monitorResult.copy(inputResults = InputRunResults(listOf(inputRunResults)))
@@ -598,17 +578,25 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
         return allShards.filter { it.primary() }.size
     }
 
-    private suspend fun getMatchingDocs(
+    /** 1. Fetch data per shard for given index. (only 10000 docs are fetched. needs to be converted to scroll if performant enough)
+     *  2. Transform documents to submit to percolate query
+     *  3. perform percolate queries
+     *  4. update docToQueries Map with all hits from percolate queries
+     *  */
+    private suspend fun fetchDataAndExecutePercolateQueriesPerShard(
         monitor: Monitor,
         monitorCtx: MonitorRunnerExecutionContext,
         docExecutionCtx: DocumentExecutionContext,
-        index: String,
-        concreteIndex: String,
+        indexName: String,
+        concreteIndexName: String,
         conflictingFields: List<String>,
-        docIds: List<String>? = null
-    ): List<Pair<String, BytesReference>> {
+        docIds: List<String>? = null,
+        monitorMetadata: MonitorMetadata,
+        inputRunResults: MutableMap<String, MutableSet<String>>,
+        docsToQueries: MutableMap<String, MutableList<String>>,
+    ) {
         val count: Int = docExecutionCtx.updatedLastRunContext["shards_count"] as Int
-        val matchingDocs = mutableListOf<Pair<String, BytesReference>>()
+        val transformedDocs = mutableListOf<Pair<String, BytesReference>>()
         for (i: Int in 0 until count) {
             val shard = i.toString()
             try {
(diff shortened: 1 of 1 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -598,17 +578,25 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
-        docIds: List<String>? = null
-    ): List<Pair<String, BytesReference>> {
+        docIds: List<String>? = null,
+        monitorMetadata: MonitorMetadata,
+        inputRunResults: MutableMap<String, MutableSet<String>>,
+        docsToQueries: MutableMap<String, MutableList<String>>,
+    ) {
         val count: Int = docExecutionCtx.updatedLastRunContext["shards_count"] as Int
-        val matchingDocs = mutableListOf<Pair<String, BytesReference>>()
+        val transformedDocs = mutableListOf<Pair<String, BytesReference>>()
```
engechas, line 599: Should this go in the shard iterating for-loop? If I'm reading this right, we are adding the docs from each shard to the list on each loop iteration without resetting the list in between. Then executing the percolator query against the docs from the current shard + all previous shards on each loop

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -722,32 +745,37 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
+                var xContentBuilder = XContentFactory.jsonBuilder().map(sourceMap)
+                val sourceRef = BytesReference.bytes(xContentBuilder)
+                logger.debug(
+                    "Monitor $monitorId: Document [${hit.id}] payload after transform for percolate query: ${sourceRef.utf8ToString()}",
+                )
+                return Pair(hit.id, sourceRef)
+            } catch (e: Exception) {
+                logger.error("Monitor $monitorId: Failed to transform payload $hit for percolate query", e)
+                // skip any document which we fail to transform because we anyway won't be able to run percolate queries on them.
+                return null
```
goyamegh, line 776: This will throw an NPE since it is returning to mapNotNull().

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -617,24 +605,59 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
+                    monitor,
+                    monitorMetadata,
+                    indexName,
+                    concreteIndexName
+                )
+
+                matchedQueriesForDocs.forEach { hit ->
+                    val id = hit.id
+                        .replace("_${indexName}_${monitor.id}", "")
+                        .replace("_${concreteIndexName}_${monitor.id}", "")
```
goyamegh, line 645: Can we abstract the percolate query execution to a separate function? Makes it modular and easy to understand.

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
eirsep: no, that's not how mapNotNull() works.

you can refer https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/map-not-null.html for better understanding of the function

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
eirsep: thanks for catching this. will fix it

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
eirsep: this chunk of code is after percolate query execution.

I am renaming the `getMatchedQueries()` to `runPercolateQueryOnTransformedDocs()` for better understanding

engechas (approved):

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -219,39 +220,18 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
                         monitorCtx,
                         docExecutionContext,
                         updatedIndexName,
                         concreteIndexName,
                         conflictingFields.toList(),
-                        matchingDocIdsPerIndex?.get(concreteIndexName)
+                        matchingDocIdsPerIndex?.get(concreteIndexName),
+                        monitorMetadata,
+                        inputRunResults,
+                        docsToQueries
```
sbcd90, line 233: we keep collecting doc ids to queries here & run 1 finding creation, trigger execution workflow https://github.com/opensearch-project/alerting/pull/1331/files#diff-68866b22ed9703814b4d5db8d3488872bcb972086ecaca10c9b8bfd54db981bcR237.
Do we also execute finding creation, trigger execution workflow at each shard level?

sbcd90 (approved):

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
eirsep: trigger execution workflows needs all results.
Findings we can brainstorm if it needs to be done at shard level
But currently findings are being indexed one by one. A bigger and more important optimization would be to ingest them batch wise using bulk requests: https://github.com/opensearch-project/alerting/issues/1333

eirsep:
will  explore how I can factor in available free memory at that point in time and we can decide on the acceptable fraction of the heap size and if the data accumulated is crossing that threshold we can submit the  percolate query with that many docs.
this way we can do for multiple indices also at once as long as it is lesser than threshold and choose executing another percolate query more deterministically

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -665,23 +688,24 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
                 SearchSourceBuilder()
                     .version(true)
                     .query(boolQueryBuilder)
-                    .size(10000) // fixme: make this configurable.
+                    .size(10000) // fixme: use scroll to ensure all docs are covered, when number of queryable docs are greater than 10k
```
lezzago, line 691: tag a github issue for this

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -704,50 +728,58 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
+                    ""
+                )
+                var xContentBuilder = XContentFactory.jsonBuilder().map(sourceMap)
+                val sourceRef = BytesReference.bytes(xContentBuilder)
+                logger.debug(
+                    "Monitor $monitorId: Document [${hit.id}] payload after transform for percolate query: ${sourceRef.utf8ToString()}",
+                )
+                return Pair(hit.id, sourceRef)
+            } catch (e: Exception) {
+                logger.error("Monitor $monitorId: Failed to transform payload $hit for percolate query", e)
```
lezzago, line 778: What kind of documents fail? We need to understand this since this would mean we are missing out on potential alerts for the user.

lezzago (approved):

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
eirsep: i am not aware if a failure here is possible. I jsut don't wish a one off doc de-ser/serialization to fail the entire execution and cause the seq no. to pause here due to repeated failures.
 I am logging this so that we get that information.
 
  I currently don't see anyway in our architecture to metriculate these kind of failures .

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -170,14 +181,17 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
                         )
                     }
                 }
+                concreteIndicesSeenSoFar.addAll(concreteIndices)
                 val updatedIndexName = indexName.replace("*", "_")
+//                lastUpdatedIndexName = updatedIndexName
```
asuresh8, line 186: uncomment or remove this line

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -170,14 +181,17 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
                 val conflictingFields = monitorCtx.docLevelMonitorQueries!!.getAllConflictingFields(
                     monitorCtx.clusterService!!.state(),
                     concreteIndices
                 )
 
                 concreteIndices.forEach { concreteIndexName ->
                     // Prepare lastRunContext for each index
+//                    lastConcreteIndexName = concreteIndexName
```
asuresh8, line 194: uncomment or remove this line

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -219,41 +233,39 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
                     // Prepare DocumentExecutionContext for each index
                     val docExecutionContext = DocumentExecutionContext(queries, indexLastRunContext, indexUpdatedRunContext)
 
-                    val matchingDocs = getMatchingDocs(
+                    fetchShardDataAndMaybeExecutePercolateQueries(
```
asuresh8, line 236: This method has a lot of parameters. Consider using an object as the parameter. That way a builder could be used so that the order of arguments does not matter. Would also make this more readable.

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -219,41 +233,39 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
-                                val docIndex = "${matchingDocs[idx].first}|$concreteIndexName"
-                                inputRunResults.getOrPut(id) { mutableSetOf() }.add(docIndex)
-                                docsToQueries.getOrPut(docIndex) { mutableListOf() }.add(id)
-                            }
-                        }
-                    }
                 }
             }
+            /* if all indices are covered still in-memory docs size limit is not breached we would need to submit
+         the percolate query at the end*/
```
asuresh8, line 255: Nit indentation looks off.

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -800,4 +906,31 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
         jsonAsMap.putAll(tempMap)
     }
+
+    /**
+     * Returns true, if the docs fetched from shards thus far amount to less than threshold
+     * amount of percentage (default:10. setting is dynamic and configurable) of the total heap size or not.
+     *
+     */
+    private fun isInMemoryDocsSizeExceedingMemoryLimit(docsBytesSize: Long, monitorCtx: MonitorRunnerExecutionContext): Boolean {
+        var thresholdPercentage = PERCOLATE_QUERY_DOCS_SIZE_MEMORY_PERCENTAGE_LIMIT.get(monitorCtx.settings)
```
asuresh8, line 916: It seems inefficient to compute the threshold every time this method is called. Is it possible to compute the threshold only when the monitor is initialized?

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -800,4 +906,31 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
+        if (thresholdPercentage > 100 || thresholdPercentage < 0) {
+            thresholdPercentage = PERCOLATE_QUERY_DOCS_SIZE_MEMORY_PERCENTAGE_LIMIT.getDefault(monitorCtx.settings)
+        }
+        val heapMaxBytes = monitorCtx.jvmStats!!.mem.heapMax.bytes
```
asuresh8, line 920: double ! is not safe. Consider replacing this with the appropriate null check.

asuresh8 (commented):
Why no unit tests?

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
eirsep: threshold is a dynamic setting so we need to fetch the latest value everytime we evaluate. opensearch does soms setting caching in the background so we are not essentially computing the threshold

eirsep:
@sbcd90 
Should we also add another condition to do batching of percolate query based on num docs and limit to 100k per search at max along with the memory limit.

That way both conditions are evaluated to decide if we want to perform percolate query on current doc set

eirsep:
Closing in favor of smaller PRs

Issue #827, opened 2023-03-06
[BUG] MonitorRunners - SearchRequest missing cancelAfterTimeInterval parameter


petardz:
**What is the bug?**
In all monitor runners we don't set cancelAfterTimeInterval parameter on SearchRequest object. Consequence of this is that global cluster setting search.cancel_after_time_interval will override it, which if set too low, could impact monitor run success rate.

**How can one reproduce the bug?**
1. Set cluster setting **_search.cancel_after_time_interval_** to something very low, like 1s
2. Create monitor on index with some number of documents/shards
3. See Cancellation timeout error (TaskCancelledExceptioncancelled exception)

rishabhmaurya:
Are you proposing its resolution to be individual plugin's responsibility to set the search.cancel_after_time_interval with max(cluster setting search.cancel_after_time_interval, plugin's default minimum value say 5s) ?

petardz:
@rishabhmaurya If I understood correctly, cluster setting `search.cancel_after_time_interval` will be used [only if same param in SearchRequest is null](https://github.com/opensearch-project/OpenSearch/blob/main/server/src/main/java/org/opensearch/action/support/TimeoutTaskCancellationUtility.java#L57)?

rishabhmaurya:
I was thinking about the case when `search.cancel_after_time_interval` is set at cluster with some higher value and the plugin overrides it to a lower value. Thus, `max(cluster setting search.cancel_after_time_interval, plugin's default minimum value say 5s)` should be used while overriding before calling _search from plugin. 
I agree with you that plugins, which generates search traffic pattern as a constant function number of index/alerts/other plugins settings, overriding it does makes sense. Feel free to raise a PR. Thanks!

petardz:
> I was thinking about the case when `search.cancel_after_time_interval` is set at cluster with some higher value and the plugin overrides it to a lower value. Thus, `max(cluster setting search.cancel_after_time_interval, plugin's default minimum value say 5s)` should be used while overriding before calling _search from plugin.

Ah got it. That makes sense!

PR #1366, opened 2024-01-04
set the cancelAfterTimeInterval parameter on SearchRequest object in …

riysaxen-amzn:
…all MonitorRunners

*#827*

*Set the cancelAfterTimeInterval parameter on SearchRequest object in all the MonitorRunners*

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
@@ -67,6 +67,7 @@ import java.util.concurrent.TimeUnit
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
 import kotlin.coroutines.suspendCoroutine
+import kotlin.math.max
 
 /** Service that handles CRUD operations for alerts */
 class AlertService(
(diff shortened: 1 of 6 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
```diff
@@ -876,6 +877,16 @@ class AlertService(
         return searchResponse
     }
 
+    fun getCancelAfterTimeInterval(): Long {
```
sbcd90, line 880: can we add this to AlertingUtils instead? https://github.com/opensearch-project/alerting/blob/main/alerting/src/main/kotlin/org/opensearch/alerting/util/AlertingUtils.kt

AWSHurneyt (commented):
Would it be a pain to add some tests for these changes?

alerting/src/main/kotlin/org/opensearch/alerting/util/AlertingUtils.kt
```diff
@@ -163,6 +166,16 @@ inline fun <T : ThreadContext.StoredContext, R> T.use(block: (T) -> R): R {
     }
 }
 
+fun getCancelAfterTimeInterval(): Long {
+    // The default value for the cancelAfterTimeInterval is -1 and so, in this case
+    // we should ignore processing on the value
+    val givenInterval = MonitorRunnerService.monitorCtx.cancelAfterTimeInterval!!.minutes
+    if (givenInterval == -1L) {
```
engechas, line 173: Does -1 get interpreted as no cancelAfterTimeout? I'm wondering if we should have a sensible default instead of defaulting to no cancelAfter. Could you elaborate on what conditions would necessitate a user changing this setting from -1 to a specific value?

riysaxen-amzn: yes -1 is a default for [cancel_after_time_interval](https://opensearch.org/docs/latest/api-reference/search/#url-parameters).  Suppose `search.cancel_after_time_interval ` is set at cluster with some higher value and the plugin overrides it to a lower value. Thus, max(cluster setting search.cancel_after_time_interval, plugin's default minimum value) should be used while overriding before calling _search from plugin.

engechas (approved):

AWSHurneyt (approved):

Issue #825, opened 2023-03-03
[BUG] Revisit AlertingException wrap logic


eirsep:
this is the actual error that needs to be surfaced
```
[2023-03-03T23:08:57,030][ERROR][o.o.a.u.AlertingException] [d52c8d13a5350334e8ebfd39b5805d5b] Alerting error: RemoteTransportException[[784e3b9e8589583cc18deb09efece92a][10.212.49.29:9300][indices:admin/mapping/put]]; nested: IllegalArgumentException[Limit of total fields [1000] has been exceeded];
```
But the error being seen is
```
{
    "error": {
        "caused_by": {
            "reason": "org.opensearch.transport.RemoteTransportException: [784e3b9e8589583cc18deb09efece92a][10.212.49.29:9300][indices:admin/mapping/put]",
            "type": "exception"
        },
        "reason": "[784e3b9e8589583cc18deb09efece92a][10.212.49.29:9300][indices:admin/mapping/put]",
        "root_cause": [
            {
                "reason": "[784e3b9e8589583cc18deb09efece92a][10.212.49.29:9300][indices:admin/mapping/put]",
                "type": "alerting_exception"
            }
        ],
        "type": "alerting_exception"
    },
    "status": 500
}
```

which really doesnt give the relevant stacktrack required to understand what went wrong

petardz:
This is why [we do "unwrapping" of exception](https://github.com/opensearch-project/alerting/blob/main/alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt#L336) when checking for specific errors. Maybe we should do the same in this wrap function?

PR #1368, opened 2024-01-06
fix fieldLimit exception in  docLevelMonitor

riysaxen-amzn:
*#825*

*Resurface actual field limit error in DocLevelMonitorRunner when updating index Mappings*

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
@@ -87,8 +87,16 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
             monitorCtx.alertIndices!!.createOrUpdateInitialFindingHistoryIndex(monitor.dataSources)
         } catch (e: Exception) {
             val id = if (monitor.id.trim().isEmpty()) "_na_" else monitor.id
-            logger.error("Error setting up alerts and findings indices for monitor: $id", e)
-            monitorResult = monitorResult.copy(error = AlertingException.wrap(e))
+            val unwrappedException = ExceptionsHelper.unwrapCause(e) as Exception
+            if (unwrappedException is IllegalArgumentException && unwrappedException.message?.contains("Limit of total fields") == true) {
+                val errorMessage =
+                    "Monitor [$id] can't process index [$monitor.dataSources] due to field mapping limit"
+                logger.error(errorMessage)
+                monitorResult = monitorResult.copy(error = AlertingException(errorMessage, RestStatus.INTERNAL_SERVER_ERROR, e))
+            } else {
+                logger.error("Error setting up alerts and findings indices for monitor: $id", e)
+                monitorResult = monitorResult.copy(error = AlertingException.wrap(e))
+            }
         }
 
         try {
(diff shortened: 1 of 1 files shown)


eirsep (commented):
plz add test to verify the behaviour

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -87,8 +87,16 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
             monitorCtx.alertIndices!!.createOrUpdateInitialFindingHistoryIndex(monitor.dataSources)
         } catch (e: Exception) {
             val id = if (monitor.id.trim().isEmpty()) "_na_" else monitor.id
-            logger.error("Error setting up alerts and findings indices for monitor: $id", e)
-            monitorResult = monitorResult.copy(error = AlertingException.wrap(e))
+            val unwrappedException = ExceptionsHelper.unwrapCause(e) as Exception
+            if (unwrappedException is IllegalArgumentException && unwrappedException.message?.contains("Limit of total fields") == true) {
```
eirsep, line 91: if you unwrap and cast to generic Exception how will this check `unwrappedException is IllegalArgumentException` be true ever?

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -87,8 +87,16 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
+                val errorMessage =
+                    "Monitor [$id] can't process index [$monitor.dataSources] due to field mapping limit"
```
eirsep, line 93: why are we not logging error message from the exception?

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -87,8 +87,16 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
+                logger.error(errorMessage)
```
eirsep, line 94: can we log the original exception instead?

eirsep:
NIT: plz rename the PR title to  `Resurface actual field limit error in DocLevelMonitorRunner when updating index Mappings`

praveensameneni (approved):
Can you also backport the fix to all the supported versions

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -87,8 +87,16 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
         } catch (e: Exception) {
             val id = if (monitor.id.trim().isEmpty()) "_na_" else monitor.id
-            logger.error("Error setting up alerts and findings indices for monitor: $id", e)
-            monitorResult = monitorResult.copy(error = AlertingException.wrap(e))
+            val unwrappedException = ExceptionsHelper.unwrapCause(e)
+            if (unwrappedException is IllegalArgumentException && unwrappedException.message?.contains("Limit of total fields") == true) {
+                val errorMessage =
+                    "Monitor [$id] can't process index [$monitor.dataSources] due to field mapping limit"
+                logger.error("Exception: ${unwrappedException.message}")
+                monitorResult = monitorResult.copy(error = AlertingException(errorMessage, RestStatus.INTERNAL_SERVER_ERROR, e))
```
engechas, line 95: A 4XX (maybe 400) would be more appropriate here than a 500 imo since the field limit getting hit is from user behavior rather than an internal server error

engechas (approved):
Agree with the prior comment about adding a test. I think the field limit setting can be manually updated to a lower value than the default so this code should be exercisable in an integ test

I'm OK to add that as a fast follow if needed

PR #1404, opened 2024-02-05
Implemented cross-cluster monitor support

AWSHurneyt:
*Issue #, if available:*
https://github.com/opensearch-project/alerting-dashboards-plugin/issues/796

*Description of changes:*
1. Implemented support for cross-cluster cluster metrics monitors. 
2. Implemented GetRemoteIndexes API to populate the frontend UI with details regarding the remote clusters, and indexes.

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
@@ -20,6 +20,7 @@ import org.opensearch.action.support.WriteRequest
 import org.opensearch.alerting.alerts.AlertIndices
 import org.opensearch.alerting.model.ActionRunResult
 import org.opensearch.alerting.model.ChainedAlertTriggerRunResult
+import org.opensearch.alerting.model.ClusterMetricsTriggerRunResult
 import org.opensearch.alerting.model.QueryLevelTriggerRunResult
 import org.opensearch.alerting.opensearchapi.firstFailureOrNull
 import org.opensearch.alerting.opensearchapi.retry
@@ -190,6 +191,19 @@ class AlertService(
             )
         }
 
+        // Including a list of triggered clusters for cluster metrics monitors
+        var triggeredClusters: MutableList<String>? = null
+        if (result is ClusterMetricsTriggerRunResult)
+            result.clusterTriggerResults.forEach {
+                if (it.triggered) {
+                    // Add an empty list if one isn't already present
+                    if (triggeredClusters.isNullOrEmpty()) triggeredClusters = mutableListOf()
+
+                    // Add the cluster to the list of triggered clusters
+                    triggeredClusters!!.add(it.cluster)
+                }
+            }
+
         // Merge the alert's error message to the current alert's history
         val updatedHistory = currentAlert?.errorHistory.update(alertError)
         return if (alertError == null && !result.triggered) {
@@ -199,7 +213,8 @@ class AlertService(
                 errorMessage = null,
                 errorHistory = updatedHistory,
                 actionExecutionResults = updatedActionExecutionResults,
-                schemaVersion = IndexUtils.alertIndexSchemaVersion
+                schemaVersion = IndexUtils.alertIndexSchemaVersion,
+                clusters = triggeredClusters
             )
         } else if (alertError == null && currentAlert?.isAcknowledged() == true) {
             null
@@ -212,6 +227,7 @@ class AlertService(
                 errorHistory = updatedHistory,
                 actionExecutionResults = updatedActionExecutionResults,
                 schemaVersion = IndexUtils.alertIndexSchemaVersion,
+                clusters = triggeredClusters
             )
         } else {
             val alertState = if (workflorwRunContext?.auditDelegateMonitorAlerts == true) {
@@ -223,7 +239,8 @@ class AlertService(
                 lastNotificationTime = currentTime, state = alertState, errorMessage = alertError?.message,
                 errorHistory = updatedHistory, actionExecutionResults = updatedActionExecutionResults,
                 schemaVersion = IndexUtils.alertIndexSchemaVersion, executionId = executionId,
-                workflowId = workflorwRunContext?.workflowId ?: ""
+                workflowId = workflorwRunContext?.workflowId ?: "",
+                clusters = triggeredClusters
             )
         }
     }
--- a/alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
@@ -11,6 +11,7 @@ import org.opensearch.alerting.action.ExecuteWorkflowAction
 import org.opensearch.alerting.action.GetDestinationsAction
 import org.opensearch.alerting.action.GetEmailAccountAction
 import org.opensearch.alerting.action.GetEmailGroupAction
+import org.opensearch.alerting.action.GetRemoteIndexesAction
 import org.opensearch.alerting.action.SearchEmailAccountAction
 import org.opensearch.alerting.action.SearchEmailGroupAction
 import org.opensearch.alerting.alerts.AlertIndices
@@ -34,6 +35,7 @@ import org.opensearch.alerting.resthandler.RestGetEmailAccountAction
 import org.opensearch.alerting.resthandler.RestGetEmailGroupAction
 import org.opensearch.alerting.resthandler.RestGetFindingsAction
 import org.opensearch.alerting.resthandler.RestGetMonitorAction
+import org.opensearch.alerting.resthandler.RestGetRemoteIndexesAction
 import org.opensearch.alerting.resthandler.RestGetWorkflowAction
 import org.opensearch.alerting.resthandler.RestGetWorkflowAlertsAction
 import org.opensearch.alerting.resthandler.RestIndexMonitorAction
@@ -59,6 +61,7 @@ import org.opensearch.alerting.transport.TransportGetEmailAccountAction
 import org.opensearch.alerting.transport.TransportGetEmailGroupAction
 import org.opensearch.alerting.transport.TransportGetFindingsSearchAction
 import org.opensearch.alerting.transport.TransportGetMonitorAction
+import org.opensearch.alerting.transport.TransportGetRemoteIndexesAction
 import org.opensearch.alerting.transport.TransportGetWorkflowAction
 import org.opensearch.alerting.transport.TransportGetWorkflowAlertsAction
 import org.opensearch.alerting.transport.TransportIndexMonitorAction
@@ -132,6 +135,7 @@ internal class AlertingPlugin : PainlessExtension, ActionPlugin, ScriptPlugin, R
         @JvmField val UI_METADATA_EXCLUDE = arrayOf("monitor.${Monitor.UI_METADATA_FIELD}")
         @JvmField val MONITOR_BASE_URI = "/_plugins/_alerting/monitors"
         @JvmField val WORKFLOW_BASE_URI = "/_plugins/_alerting/workflows"
+        @JvmField val REMOTE_BASE_URI = "/_plugins/_alerting/remote"
         @JvmField val DESTINATION_BASE_URI = "/_plugins/_alerting/destinations"
         @JvmField val LEGACY_OPENDISTRO_MONITOR_BASE_URI = "/_opendistro/_alerting/monitors"
         @JvmField val LEGACY_OPENDISTRO_DESTINATION_BASE_URI = "/_opendistro/_alerting/destinations"
@@ -183,7 +187,8 @@ internal class AlertingPlugin : PainlessExtension, ActionPlugin, ScriptPlugin, R
             RestGetWorkflowAlertsAction(),
             RestGetFindingsAction(),
             RestGetWorkflowAction(),
-            RestDeleteWorkflowAction()
+            RestDeleteWorkflowAction(),
+            RestGetRemoteIndexesAction(),
         )
     }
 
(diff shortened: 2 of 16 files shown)


AWSHurneyt:
Successfully building this PR will depend on merging the common utils changes, and publishing maven.
https://github.com/opensearch-project/common-utils/pull/584

praveensameneni (approved):

bowenlan-amzn:
@AWSHurneyt can you fix the workflow

For the description, can you add how would user use this feature? and a quick walkthrough of the implementation idea.

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
```diff
@@ -190,6 +191,19 @@ class AlertService(
             )
         }
 
+        // Including a list of triggered clusters for cluster metrics monitors
+        var triggeredClusters: MutableList<String>? = null
+        if (result is ClusterMetricsTriggerRunResult)
```
eirsep, line 196: `composeQueryLevelAlert` method is now checking for ClusterMetricsTriggerRunResult

eirsep: why aren't we adding `it.cluster` to all the alerts irrespecitve of monitor type? even if we dont support the other monitor types it will just show local cluster id right?

AWSHurneyt:
> @AWSHurneyt can you fix the workflow
> 
> For the description, can you add how would user use this feature? and a quick walkthrough of the implementation idea.

@bowenlan-amzn I added some more details to the description.

alerting/src/main/kotlin/org/opensearch/alerting/resthandler/RestGetRemoteIndexesAction.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.opensearch.client.node.NodeClient
+import org.opensearch.core.common.Strings
+import org.opensearch.rest.BaseRestHandler
+import org.opensearch.rest.RestHandler
+import org.opensearch.rest.RestRequest
+import org.opensearch.rest.action.RestToXContentListener
+
+private val log = LogManager.getLogger(RestGetRemoteIndexesAction::class.java)
+
+class RestGetRemoteIndexesAction : BaseRestHandler() {
```
eirsep, line 21: javadocs and code comments

alerting/src/main/kotlin/org/opensearch/alerting/resthandler/RestGetRemoteIndexesAction.kt
```diff
@@ -0,0 +1,50 @@
...
+    val ROUTE = "${AlertingPlugin.REMOTE_BASE_URI}/indexes"
```
eirsep, line 22: indices*

AWSHurneyt: I went with `indexes` because we had this issue a while back in which we were moving away from `indices`.
https://github.com/opensearch-project/index-management-dashboards-plugin/issues/637

alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
```diff
@@ -183,7 +187,8 @@ internal class AlertingPlugin : PainlessExtension, ActionPlugin, ScriptPlugin, R
             RestGetWorkflowAlertsAction(),
             RestGetFindingsAction(),
             RestGetWorkflowAction(),
-            RestDeleteWorkflowAction()
+            RestDeleteWorkflowAction(),
+            RestGetRemoteIndexesAction(),
```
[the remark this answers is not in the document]
AWSHurneyt: I went with `indexes` because we had this issue a while back in which we were moving away from `indices`.
https://github.com/opensearch-project/index-management-dashboards-plugin/issues/637

eirsep:
why are we adding support to fetch remote indices in alerting plugin? isn't that a core feature?

alerting/src/main/kotlin/org/opensearch/alerting/util/CrossClusterMonitorUtils.kt
```diff
@@ -0,0 +1,231 @@
...
+        fun parseIndexesForRemoteSearch(indexes: List<String>, localClusterName: String): List<String> {
+            return indexes.map {
+                var index = it
+                val clusterName = parseClusterName(it)
+                if (clusterName.isNotEmpty() && clusterName == localClusterName) {
+                    index = parseIndexName(it)
+                }
+                index
+            }
+        }
```
bowenlan-amzn, line 125: This method is confusing to me, in the end it seems only useful to user input `<local cluster name>:<index>` and transform it in to `<index>`
Instead of doing such thing, why not just let it fail?

alerting/src/main/kotlin/org/opensearch/alerting/QueryLevelMonitorRunner.kt
```diff
@@ -65,7 +66,19 @@ object QueryLevelMonitorRunner : MonitorRunner() {
...
             val triggerCtx = QueryLevelTriggerExecutionContext(monitor, trigger as QueryLevelTrigger, monitorResult, currentAlert)
-            val triggerResult = monitorCtx.triggerService!!.runQueryLevelTrigger(monitor, trigger, triggerCtx)
+            val triggerResult = when (monitor.monitorType) {
+                Monitor.MonitorType.QUERY_LEVEL_MONITOR ->
+                    monitorCtx.triggerService!!.runQueryLevelTrigger(monitor, trigger, triggerCtx)
+                Monitor.MonitorType.CLUSTER_METRICS_MONITOR -> {
+                    val remoteMonitoringEnabled = monitorCtx.clusterService!!.clusterSettings.get(AlertingSettings.REMOTE_MONITORING_ENABLED)
+                    logger.debug("Remote monitoring enabled: {}", remoteMonitoringEnabled)
+                    if (remoteMonitoringEnabled) monitorCtx.triggerService!!.runClusterMetricsTrigger(monitor, trigger, triggerCtx, monitorCtx.clusterService!!)
+                    else monitorCtx.triggerService!!.runQueryLevelTrigger(monitor, trigger, triggerCtx)
```
bowenlan-amzn, line 76: Just curious, why cluster metric monitor run query level trigger?

AWSHurneyt:
> why are we adding support to fetch remote indices in alerting plugin? isn't that a core feature?

@eirsep The GetRemoteIndexes API is used to gather details about the clusters and their indexes to display in the frontend. The frontend plugins do not have access to remoteClients to make most API calls against the remote clusters.

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
```diff
@@ -190,6 +191,19 @@ class AlertService(
             )
         }
 
+        // Including a list of triggered clusters for cluster metrics monitors
+        var triggeredClusters: MutableList<String>? = null
```
bowenlan-amzn, line 195: Consider this is for experiemental, does here need a check for cluster setting for this feature?

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
AWSHurneyt: Correct. Cluster metrics monitors started out as query level monitors before we added additional monitor types when we introduced bucket level monitors. Their alerts were handled in a similar way up until this point, so I added this check in `composeQueryLevelAlert` rather than duplicate the code elsewhere.

alerting/src/main/kotlin/org/opensearch/alerting/util/IndexUtils.kt
```diff
@@ -139,18 +140,27 @@ class IndexUtils {
...
+                    clusterState,
                     IndicesOptions.lenientExpand(),
                     true,
                     index
                 )
                 result.addAll(concreteIndices)
             }
 
+            if (result.size == 0) {
+                throw IndexNotFoundException(indices[0])
```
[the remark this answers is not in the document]
AWSHurneyt: Are you suggesting refactoring line 157 to something like this? I'll revise the PR with that change.
```
indices.forEach { index ->
                val concreteIndices = resolver.concreteIndexNames(
                    clusterState,
                    IndicesOptions.lenientExpand(),
                    true,
                    index
                )
                if (concreteIndices.isNotEmpty()) result.addAll(concreteIndices)
                else throw IndexNotFoundException(indices[0])
            }

File: alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetRemoteIndexesAction.kt [hunk 9]
```diff
@@ -0,0 +1,192 @@
...
+            return
+        }
+
+        val user = readUserFromThreadContext(client)
+        if (!validateUserBackendRoles(user, actionListener)) return
+
+        client.threadPool().threadContext.stashContext().use {
+            scope.launch {
+                val singleThreadContext = newSingleThreadContext("GetRemoteIndexesActionThread")
+                withContext(singleThreadContext) {
```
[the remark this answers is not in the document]
AWSHurneyt: Ah interesting. It looks like on this line ISM checks whether the user info is in the context, and injects it if not. Craig and I discussed doing something similar last week, but landed on the single thread implementation since we want to ensure the actions performed by the API take place in a specific sequence. Do you have any insight into the pros and cons of ISM's approach?
https://github.com/opensearch-project/index-management/blob/ba24f862a4b59bdac5a4def76033766346e9eea8/src/main/kotlin/org/opensearch/indexmanagement/opensearchapi/OpenSearchExtensions.kt#L300

alerting/src/main/kotlin/org/opensearch/alerting/util/CrossClusterMonitorUtils.kt
AWSHurneyt: If the local cluster is named `cluster-1`, and a monitor input is configured with indexes `["cluster-1:index-1","cluster-2:index-2"]`, the node client of the local cluster will not be able to query the first index because the pattern references the local cluster name. This function will parse that list of indexes to `["index-1","cluster-2:index-2"]` so it can be successfully fed into the node client.

alerting/src/main/kotlin/org/opensearch/alerting/QueryLevelMonitorRunner.kt
AWSHurneyt: Cluster metrics monitors were initially query level monitors before bucket level introduced monitor types. If I remember correctly, we gave cluster metrics monitors a unique type around the implementation of doc level monitors. It could be nice to eventually give cluster metrics monitors their own assets like runner, and trigger; but at the time, their execution was so similar to query level, we didn't change them.

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
AWSHurneyt: I don't think that would be necessary, as the feature disabled, this will always be just the local cluster. But I'm open adding a setting check if we think that's best.

praveensameneni (approved):

amsiglan (approved):

getsaurabh02 (approved):
Lets follow up on test failures via opening separate issues

eirsep:
have we made changes in security plugin?

eirsep:
no integ tests added to verify end to end behaviour?

alerting/src/main/kotlin/org/opensearch/alerting/util/CrossClusterMonitorUtils.kt
bowenlan-amzn: Yeah, I get this function. The confusing part is why we want to support user to use local cluster name in the input, does the other feature from UI also support that?

AWSHurneyt: I wanted to cover cases where a user may do this without knowing.

bowenlan-amzn: If user cannot do it in other place, this should also be forbidden here.
So to decide, try do a search call in a cross cluster environment with `<local cluster name>:<index>` on a local index, and see if that works. If so, this is ok.

AWSHurneyt:
> have we made changes in security plugin?

No, we didn't need to make any changes in the security plugin.

AWSHurneyt:
> no integ tests added to verify end to end behaviour?

This is not ideal, but so far testing has all been manual. I have implementing feature-specific tests as a fast follow-up item. For some additional context, query and bucket monitors already supported remote index querying. The major change in this PR is supporting remote calls for cluster metrics monitors.

AWSHurneyt:
PR https://github.com/opensearch-project/alerting/pull/1405 adjusted the ktlint version, which is causing style failures upon rerunning the tests with the PR https://github.com/opensearch-project/common-utils/pull/584/changes published to maven. Cleaning up those errors.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetRemoteIndexesAction.kt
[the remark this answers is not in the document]
bowenlan-amzn: I think using `ThreadContextElement` is more natural to kotlin coroutine, since it's provided directly by them.
Basically what it does is to attach this element to the coroutine that will execute your code.

While force the user info in one thread, I feel it's possible that this thread may be used to execute other coroutine? then it may cause some problem.

praveensameneni (approved):

AWSHurneyt:
> Lets follow up on test failures via opening separate issues

All tests are passing now. Merging in these changes.

PR #1425, opened 2024-02-13
Add an _exists_ check to document level monitor queries

jowg-amazon:
*Issue #, if available:*
[#854](https://github.com/opensearch-project/security-analytics/issues/854)

*Description of changes:*
Related to https://github.com/opensearch-project/security-analytics/pull/852

Checks if `_exists_` is present in the query. If it is, then replace the value with the field name and the correctly appended index name and monitor id.

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt
@@ -396,7 +396,14 @@ class DocLevelMonitorQueries(private val client: Client, private val clusterServ
             var query = it.query
             conflictingPaths.forEach { conflictingPath ->
                 if (query.contains(conflictingPath)) {
-                    query = query.replace("$conflictingPath:", "${conflictingPath}_<index>_$monitorId:")
+                    if (query.contains("_exists_")) {
+                        // 1. append the key field "field:"
+                        query = query.replace("$conflictingPath:", "${conflictingPath}_<index>_$monitorId:")
+                        // 2. append the val field "_exists_field"
+                        query = query.replace("_exists_$conflictingPath", "${conflictingPath}_<index>_$monitorId")
+                    } else {
+                        query = query.replace("$conflictingPath:", "${conflictingPath}_<index>_$monitorId:")
+                    }
                     filteredConcreteIndices.addAll(conflictingPathToConcreteIndices[conflictingPath]!!)
                 }
             }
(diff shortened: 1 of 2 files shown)


eirsep (changes requested):
> Checks if _exists_ is present in the query. 
**Can't we do a not null check instead**?

jowg-amazon:
@eirsep Opensearch allows `null` to be passed in as a value so a not null check would be different than an exists check. We want to check whether or not the field even exists in the document, not if the field is `null`.

alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt
```diff
@@ -350,7 +350,14 @@ class DocLevelMonitorQueries(private val client: Client, private val clusterServ
             var query = it.query
             conflictingPaths.forEach { conflictingPath ->
                 if (query.contains(conflictingPath)) {
-                    query = query.replace("$conflictingPath:", "${conflictingPath}_<index>_$monitorId:")
+                    if (query.contains("_exists_")) {
+                        // 1. append the key field "field:"
+                        query = query.replace("$conflictingPath:", "${conflictingPath}_<index>_$monitorId:")
+                        // 2. append the val field "_exists_field"
```
eirsep, line 356: shouldn't there be `:` between `_exists_` and `<field>` ?

eirsep (changes requested):
shouldn't there be : between _exists_ and <field> ?

alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt
```diff
@@ -372,7 +379,14 @@ class DocLevelMonitorQueries(private val client: Client, private val clusterServ
             var query = it.query
             flattenPaths.forEach { fieldPath ->
                 if (!conflictingPaths.contains(fieldPath.first)) {
-                    query = query.replace("${fieldPath.first}:", "${fieldPath.first}_${sourceIndex}_$monitorId:")
+                    if (query.contains("_exists_")) {
```
eirsep, line 382: can we extract code into common method?

alerting/src/test/kotlin/org/opensearch/alerting/DocumentMonitorRunnerIT.kt
```diff
@@ -1674,6 +1674,349 @@ class DocumentMonitorRunnerIT : AlertingRestTestCase() {
         assertEquals(1, output.objectMap("trigger_results").values.size)
     }
 
+    fun `test execute monitor generates alerts and findings with NOT query`() {
```
eirsep, line 1677: nit: rename test to specify we are verifying not + exists

alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt
jowg-amazon: extracted the code into a common method

alerting/src/test/kotlin/org/opensearch/alerting/DocumentMonitorRunnerIT.kt
jowg-amazon: renamed the test

jowg-amazon:
> shouldn't there be : between _exists_ and ?

@eirsep Removed the _exists_ flag so it parses to find wherever the query is _exists_: <field name> and transforms the <field name>

eirsep (approved):

PR #1427, opened 2024-02-14
Feature findings enhancemnt

riysaxen-amzn:
*#795*

*Added detectionType severity and searchString as params in Findings API*

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/resthandler/RestGetFindingsAction.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/resthandler/RestGetFindingsAction.kt
@@ -46,6 +46,7 @@ class RestGetFindingsAction : BaseRestHandler() {
         val startIndex = request.paramAsInt("startIndex", 0)
         val searchString = request.param("searchString", "")
         val severity: String? = request.param("severity", "ALL")
+        val detectionType: String? = request.param("detectionType", "rules")
 
         val table = Table(
             sortOrder,
@@ -59,7 +60,8 @@ class RestGetFindingsAction : BaseRestHandler() {
         val getFindingsSearchRequest = GetFindingsRequest(
             findingID,
             table,
-            severity
+            severity,
+            detectionType
         )
         return RestChannelConsumer {
                 channel ->
--- a/alerting/src/main/kotlin/org/opensearch/alerting/resthandler/RestGetFindingsAction.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/resthandler/RestGetFindingsAction.kt
@@ -45,6 +45,7 @@ class RestGetFindingsAction : BaseRestHandler() {
         val size = request.paramAsInt("size", 20)
         val startIndex = request.paramAsInt("startIndex", 0)
         val searchString = request.param("searchString", "")
+        val severity: String? = request.param("severity", "ALL")
 
         val table = Table(
             sortOrder,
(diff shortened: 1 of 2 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetFindingsAction.kt
```diff
@@ -109,6 +111,41 @@ class TransportGetFindingsSearchAction @Inject constructor(
...
+                "queries",
+                when {
+                    detectionType.equals("threat", ignoreCase = true) -> {
+                        QueryBuilders.boolQuery().filter(
+                            QueryBuilders.prefixQuery("queries.id", "threat_intel_")
+                        )
+                    }
+                    else -> {
+                        QueryBuilders.boolQuery().mustNot(
+                            QueryBuilders.prefixQuery("queries.id", "threat_intel_")
```
engechas, line 125: What's the reason to filter out the threat intel queries here? As a user I would expect the filter to be based on the provided `detectionType`

riysaxen-amzn: Yes, I have categorized the detectionType parameter into two types: "threat" and normal rule based.  If the frontend API calls the findings API with detectionType set to "threat," all threat intel-related findings will be generated with this query. I'm open to considering alternative solutions for this categorization.

engechas: I think the enum comment I had on the other CR makes this a bit more clear. Having the arbitrary text input + looking for exactly 2 cases seems like it should be enum-based.

If we don't expect the detectionType options to expand, then having a boolean field seems more user-friendly to me. Something like `getThreatIntelFindingsOnly` (name could use some work) that allows the user to pick which type of findings are returned.

Or maybe a new API, say GetThreatIntelFindings, would also make sense to me.

Another thought on this - this change is arguably a breaking one. We are changing the default behavior of an existing API to now filter out the threat intel findings. If a user has a system that relies on this API already and they are expecting threat intel findings, then it would break their use case.

Is there a way we can explore doing the filtering on the frontend, or at least leaving the default behavior as-is? I'm a bit concerned with merging this at the moment, but open to others' thoughts.

engechas (approved):

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetFindingsAction.kt
engechas: Missed line 128.. not a breaking change or change to default behavior since filtering is only applied if a detectionType is supplied

AWSHurneyt (approved):

alerting/src/main/kotlin/org/opensearch/alerting/resthandler/RestGetFindingsAction.kt
```diff
@@ -57,7 +59,9 @@ class RestGetFindingsAction : BaseRestHandler() {
 
         val getFindingsSearchRequest = GetFindingsRequest(
             findingID,
-            table
+            table,
+            severity,
+            detectionType
```
eirsep, line 64: what is detectionType??

eirsep: we can't use this variable .. this is not a keyword in alerting domain.. if it's for security analytics we need to translate it into a jargon or query searchString which is generalized to Alerting plugin?

Issue #1353, opened 2023-12-22
Optimize doc level monitor performance: Batch docs for percolate query searches based on available memory and cpu


eirsep:
**Is your feature request related to a problem?**
Doc level monitor execution does the following 2 memory, cpu sensitive operations in search path :
- Fetch docs source for docs (> last read sequence number) for every shard per index
- Accumulate docs source list per index and perform percolate queries this causes heap usage exceeded exception when docs in memory are too high

**What solution would you like?**
Collect docs from shards up until a threshold which factors heap usage and available memory. Perform percolate query. If more shards remain to be queried repeat the above process until all shards are queried in current execution of doc level monitor
TODO - factor in cpu usage also
**What alternatives have you considered?**
Perform percolate query per shard - caused CPU regression due to too many requests
Perform percolate query for all docs across all shards, indices being  - aggravates heap usage exceed issue as more docs are queried
(Current) Perform percolate query per concrete index - still not determinisitc. may make too many percolate queries (if very few docs in memory from one index) or too less  (if too many docs from one index in memory)

eirsep:
https://github.com/opensearch-project/alerting/pull/1331

PR #1435, opened 2024-02-19
Add jvm aware setting and max num docs settings for batching docs for percolate queries

eirsep:
With these changes the number of docs submitted in a single percolate query is not naively set per shard or per index. Rather we have 2 settings to decide how many docs to submit for percolate query in doc level monitor

**Solves the followping problems**
```
IllegalStateException[Failed to run percolate search for sourceIndex [cloudtrail_alias] and queryIndex [.opensearch-sap-cloudtrail-detectors-queries-000001] for 180000 document(s)];
 nested: SearchPhaseExecutionException[SearchTask was cancelled]; 
nested: TaskCancelledException[org.opensearch.core.concurrency.OpenSearchRejectedExecutionException: cancelled task with reason: Cancellation timeout of 5m is expired]; 
nested: OpenSearchRejectedExecutionException[cancelled task with reason: Cancellation timeout of 5m is expired];
```
```
"error_message" : "IllegalStateException[Failed to run percolate search for sourceIndex [log-aws-cloudtrail-2023-08] and queryIndex [.opensearch-sap-cloudtrail-detectors-queries-000001] for 10000 document(s)]; 
nested: SearchPhaseExecutionException[all shards failed]; 
nested: [cancelled task with reason: heap usage exceeded [45.9mb >= 9.2mb]]; 
nested: OpenSearchRejectedExecutionException[cancelled task with reason: heap usage exceeded [45.9mb >= 9.2mb]];
```
*Issue #, if available:*
**Optimize doc level monitor performance: Batch docs for percolate query searches based on available memory and cpu #1353**

*Description of changes:*

**Log message from opensearch cluster, when setting is at 40k docs per batch and 10% of heap to break batch and perform percolate query for ingestion rate of 250K docs per minute**

**`Monitor org.opensearch.client.node.NodeClient@1440ce1 PERF_DEBUG: Percolate query time taken millis = 9.4s`**

Old Latency of percolate query : 5+ minutes leading to cancellation. New latency <1 minute
-

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
@@ -100,6 +100,7 @@ import org.opensearch.core.xcontent.XContentParser
 import org.opensearch.env.Environment
 import org.opensearch.env.NodeEnvironment
 import org.opensearch.index.IndexModule
+import org.opensearch.monitor.jvm.JvmStats
 import org.opensearch.painless.spi.Allowlist
 import org.opensearch.painless.spi.AllowlistLoader
 import org.opensearch.painless.spi.PainlessExtension
@@ -263,6 +264,7 @@ internal class AlertingPlugin : PainlessExtension, ActionPlugin, ScriptPlugin, R
             .registerTriggerService(TriggerService(scriptService))
             .registerAlertService(AlertService(client, xContentRegistry, alertIndices))
             .registerDocLevelMonitorQueries(DocLevelMonitorQueries(client, clusterService))
+            .registerJvmStats(JvmStats.jvmStats())
             .registerWorkflowService(WorkflowService(client, xContentRegistry))
             .registerConsumers()
             .registerDestinationSettings()
@@ -320,6 +322,8 @@ internal class AlertingPlugin : PainlessExtension, ActionPlugin, ScriptPlugin, R
             AlertingSettings.ALERT_HISTORY_MAX_DOCS,
             AlertingSettings.ALERT_HISTORY_RETENTION_PERIOD,
             AlertingSettings.ALERTING_MAX_MONITORS,
+            AlertingSettings.PERCOLATE_QUERY_DOCS_SIZE_MEMORY_PERCENTAGE_LIMIT,
+            AlertingSettings.PERCOLATE_QUERY_MAX_NUM_DOCS_IN_MEMORY,
             AlertingSettings.REQUEST_TIMEOUT,
             AlertingSettings.MAX_ACTION_THROTTLE_VALUE,
             AlertingSettings.FILTER_BY_BACKEND_ROLES,
--- a/alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
@@ -25,6 +25,8 @@ import org.opensearch.alerting.model.MonitorRunResult
 import org.opensearch.alerting.model.userErrorMessage
 import org.opensearch.alerting.opensearchapi.suspendUntil
 import org.opensearch.alerting.script.DocumentLevelTriggerExecutionContext
+import org.opensearch.alerting.settings.AlertingSettings.Companion.PERCOLATE_QUERY_DOCS_SIZE_MEMORY_PERCENTAGE_LIMIT
+import org.opensearch.alerting.settings.AlertingSettings.Companion.PERCOLATE_QUERY_MAX_NUM_DOCS_IN_MEMORY
 import org.opensearch.alerting.util.AlertingException
 import org.opensearch.alerting.util.IndexUtils
 import org.opensearch.alerting.util.defaultToPerExecutionAction
@@ -60,12 +62,15 @@ import org.opensearch.index.query.BoolQueryBuilder
 import org.opensearch.index.query.Operator
 import org.opensearch.index.query.QueryBuilders
 import org.opensearch.percolator.PercolateQueryBuilderExt
+import org.opensearch.search.SearchHit
 import org.opensearch.search.SearchHits
 import org.opensearch.search.builder.SearchSourceBuilder
 import org.opensearch.search.sort.SortOrder
 import java.io.IOException
 import java.time.Instant
 import java.util.UUID
+import java.util.concurrent.atomic.AtomicLong
+import java.util.stream.Collectors
 import kotlin.math.max
 
 object DocumentLevelMonitorRunner : MonitorRunner() {
@@ -83,7 +88,10 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
         logger.debug("Document-level-monitor is running ...")
         val isTempMonitor = dryrun || monitor.id == Monitor.NO_ID
         var monitorResult = MonitorRunResult<DocumentLevelTriggerRunResult>(monitor.name, periodStart, periodEnd)
-
+        var nonPercolateSearchesTimeTaken = AtomicLong(0)
+        var percolateQueriesTimeTaken = AtomicLong(0)
+        var totalDocsQueried = AtomicLong(0)
+        var docTransformTimeTaken = AtomicLong(0)
         try {
             monitorCtx.alertIndices!!.createOrUpdateAlertIndex(monitor.dataSources)
             monitorCtx.alertIndices!!.createOrUpdateInitialAlertHistoryIndex(monitor.dataSources)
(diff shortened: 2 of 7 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -704,89 +813,114 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
-        concreteIndex: String
+        concreteIndices: List<String>,
+        monitorInputIndices: List<String>,
     ): SearchHits {
-        val boolQueryBuilder = BoolQueryBuilder().must(QueryBuilders.matchQuery("index", index).operator(Operator.AND))
-
-        val percolateQueryBuilder = PercolateQueryBuilderExt("query", docs, XContentType.JSON)
+        val indices = docs.stream().map { it.second.indexName }.distinct().collect(Collectors.toList())
+        val boolQueryBuilder = BoolQueryBuilder().must(buildShouldClausesOverPerIndexMatchQueries(indices))
+        val percolateQueryBuilder =
```
sbcd90, line 837: we have the concrete index here https://github.com/opensearch-project/alerting/pull/1435/files#diff-68866b22ed9703814b4d5db8d3488872bcb972086ecaca10c9b8bfd54db981bcR669 but it is not passed to `performPercolateQueryAndResetCounters` https://github.com/opensearch-project/alerting/pull/1435/files#diff-68866b22ed9703814b4d5db8d3488872bcb972086ecaca10c9b8bfd54db981bcR745 & ultimately we add list of should queries here https://github.com/opensearch-project/alerting/pull/1435/files#diff-68866b22ed9703814b4d5db8d3488872bcb972086ecaca10c9b8bfd54db981bcR888.

can we just pass concreteIndexName & avoid list of should queries?

alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
```diff
@@ -320,6 +322,8 @@ internal class AlertingPlugin : PainlessExtension, ActionPlugin, ScriptPlugin, R
             AlertingSettings.ALERT_HISTORY_MAX_DOCS,
             AlertingSettings.ALERT_HISTORY_RETENTION_PERIOD,
             AlertingSettings.ALERTING_MAX_MONITORS,
+            AlertingSettings.PERCOLATE_QUERY_DOCS_SIZE_MEMORY_PERCENTAGE_LIMIT,
+            AlertingSettings.PERCOLATE_QUERY_MAX_NUM_DOCS_IN_MEMORY,
```
sbcd90, line 326: do we rename this to `DOC_LEVEL_MONITORS` instead of `PERCOLATE_QUERY` as the concept of percolate query is internal to doc level monitors?

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -83,7 +88,10 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
         logger.debug("Document-level-monitor is running ...")
         val isTempMonitor = dryrun || monitor.id == Monitor.NO_ID
         var monitorResult = MonitorRunResult<DocumentLevelTriggerRunResult>(monitor.name, periodStart, periodEnd)
-
+        var nonPercolateSearchesTimeTaken = AtomicLong(0)
+        var percolateQueriesTimeTaken = AtomicLong(0)
+        var totalDocsQueried = AtomicLong(0)
+        var docTransformTimeTaken = AtomicLong(0)
```
sbcd90, line 94: why do we need these variables to be `atomiclong` because `DocumentLevelMonitorRunner` runs in a single thread?

alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunnerService.kt
```diff
@@ -258,11 +264,19 @@ object MonitorRunnerService : JobRunner, CoroutineScope, AbstractLifecycleCompon
         when (job) {
             is Workflow -> {
                 launch {
+                    logger.debug(
+                        "PERF_DEBUG: executing workflow ${job.id} on node " +
```
engechas, line 268: Minor - could we remove the `PERF_DEBUG` prefix? Same comment for other debug logs

alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
```diff
@@ -25,6 +25,29 @@ class AlertingSettings {
             Setting.Property.NodeScope, Setting.Property.Dynamic
         )
 
+        /** Defines the threshold percentage of heap size in bytes till which we accumulate docs in memory before we query against percolate query
+         * index in document level monitor execution.
+         */
+        val PERCOLATE_QUERY_DOCS_SIZE_MEMORY_PERCENTAGE_LIMIT = Setting.intSetting(
```
engechas, line 31: Minor: MEMORY -> HEAP

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
eirsep: Agreed that this is a single threaded operation but `AtomicLong.getAndAdd()` updates the final variable while Kotlin `Long` is immutable and would require me to use new variables if i do additions or any math operations .
```
val docsSizeInBytes: Long = 0L
val updatedDocsSizeInBytes = docsSizeInBytes + 5
```
I was trying to avoid initializing new variables.

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -704,89 +843,119 @@ object DocumentLevelMonitorRunner : MonitorRunner() {
...
             boolQueryBuilder.must(QueryBuilders.matchQuery("monitor_id", monitor.id).operator(Operator.AND))
         }
         boolQueryBuilder.filter(percolateQueryBuilder)
-
-        val queryIndex = monitorMetadata.sourceToQueryIndexMapping[index + monitor.id]
-        if (queryIndex == null) {
-            val message = "Failed to resolve concrete queryIndex from sourceIndex during monitor execution!" +
-                " sourceIndex:$concreteIndex queryIndex:${monitor.dataSources.queryIndex}"
+        val queryIndices =
+            docs.map { monitorMetadata.sourceToQueryIndexMapping[it.second.indexName + monitor.id] }.distinct()
```
engechas, line 877: Is there a case where there would be more than one query index here? If so, should only docs with query index A be submitted to a percolator query against query index A? Currently it seems like we are sending all docs to a percolate query against both index A and index B

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
sbcd90: we can change `val` to `var` & it becomes mutable

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
eirsep: Great question.. will consult @sbcd90 for advice here

alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
eirsep: renamed to `DOC_LEVEL_MONITOR_DOCS_SIZE_HEAP_PERCENTAGE_LIMIT`

alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
eirsep: renamed to `DOC_LEVEL_MONITOR_DOCS_SIZE_HEAP_PERCENTAGE_LIMIT`

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
eirsep: we are collecting docs across different indices with this change. we don't perform percolate query once for each index now. so we can't just pass one concreteIndex name. we will have to pass all indices whose docs are being read

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
eirsep: using LongAdder instead of var as they're being passed to different methods

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
eirsep: Currently for a single monitor there will be only one query index acc. to @sbcd90

engechas: Got it - I could see this get overlooked if that changes in the future. Would it be too difficult to break up the execution of the percolate query to be per-query-index? It shouldn't have any performance/functionality impacts currently with only one query index, but would guard against that changing down the line.

engechas (approved):

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
eirsep: correctness would be intact so that's ok. we would anyway evaluate performance if we make such a change of having to query multiple indices

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
sbcd90: why can't we do something like this?
```
var docsSizeInBytes: Long = 0L
docSizeinBytes = docSizeInBytes.plus(5)
```

sbcd90 (approved):

Issue #1367, opened 2024-01-05
Optimize doc level monitor performance: Only fetch the fields present in doc level queries for docs to submit to percolate queries


eirsep:
**Is your feature request related to a problem?**
A clear and concise description of what the problem is, e.g. _I'm always frustrated when [...]_

**What solution would you like?**

**What alternatives have you considered?**

**Do you have any additional context?**

PR #1441, opened 2024-02-22
optimize to fetch only fields relevant to doc level queries in doc level monitor instead of entire _source for each doc

eirsep:
*Issue #, if available:*
#1367

*Description of changes:*
optimize to fetch only fields relevant to doc level queries in doc level monitor instead of entire _source for each doc 

**Percolate query latency BEFORE change: ~10 seconds
Percolate query latency AFTER change: ~1 second**

*CheckList:*

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
@@ -236,6 +236,28 @@ class DocumentLevelMonitorRunner : MonitorRunner() {
                         }
                     }
 
+                    val fieldsToBeQueried = mutableSetOf<String>()
+
+                    for (it in queries) {
+                        if (it.queryFieldNames.isEmpty()) {
+                            fieldsToBeQueried.clear()
+                            logger.debug(
+                                "Monitor ${monitor.id} : " +
+                                    "Doc Level query ${it.id} : ${it.query} doesn't have queryFieldNames populated. " +
+                                    "Cannot optimize monitor to fetch only query-relevant fields. " +
+                                    "Querying entire doc source."
+                            )
+                            break
+                        }
+                        fieldsToBeQueried.addAll(it.queryFieldNames)
+                    }
+                    if (monitorCtx.fetchOnlyQueryFieldNames && fieldsToBeQueried.isNotEmpty()) {
+                        logger.debug(
+                            "Monitor ${monitor.id} Querying only fields " +
+                                "${fieldsToBeQueried.joinToString()} instead of entire _source of documents"
+                        )
+                    }
+
                     // Prepare DocumentExecutionContext for each index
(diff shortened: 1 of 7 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -236,6 +236,28 @@ class DocumentLevelMonitorRunner : MonitorRunner() {
...
+                            logger.debug(
+                                "Monitor ${monitor.id} : " +
+                                    "Doc Level query ${it.id} : ${it.query} doesn't have queryFieldNames populated. " +
+                                    "Cannot optimize monitor to fetch only query-relevant fields. " +
+                                    "Querying entire doc source."
+                            )
+                            break
+                        }
+                        fieldsToBeQueried.addAll(it.queryFieldNames)
+                    }
```
engechas, line 253: Minor - could put this an if with `monitorCtx.fetchOnlyQueryFieldNames` as well to avoid the unnecessary work. Work should be negligible though so not a big deal

Unrelated to above - is there a scenario where we would want to turn this filtering off? Adding the setting to toggle on/off does no harm IMO, but not sure that it's necessary

engechas (approved):

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -236,6 +236,28 @@ class DocumentLevelMonitorRunner : MonitorRunner() {
                         }
                     }
 
+                    val fieldsToBeQueried = mutableSetOf<String>()
+
+                    for (it in queries) {
+                        if (it.queryFieldNames.isEmpty()) {
```
engechas, line 242: Will we get an NPE here for migrations from older versions that did not have a queryFieldNames field in the DocLevelQuery objects?

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
eirsep: Retaining the feature flag as it's a super admin permission setting only so it can't be flipped off easily

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
eirsep: The constructor will  initialize to empty list if not present in index.

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
eirsep: moved logic for population of fieldsToBeQueried  into if block

sbcd90 (approved):

Issue #1409, opened 2024-02-06
Remove search calls to fetch max seq no for each shard in doc level monitor. Use first call to fetch data from a shard to determinte max seq no.


eirsep:
Current behaviour
---
- If there are `n` indices with `m` shards each, we first run `n*m` search calls to simply fetch the max sequence number
for each shard and then perform the searches which fetch docs from each shard based on last seen seq no. and max seq no.

Optimized behaviour proposed
---
Change first search request from each shard to fetch all docs in descending order of seq no with filter on `seqno > <last seen seqno>
That way we are fetching docs and also able to note the sequence number correctly within the same call and avoid the first batch of search calls which simply help determine the sequence number

PR #1445, opened 2024-02-27
optimize sequence number calculation and reduce search requests in doc level monitor execution

eirsep:
Optimize sequence number calculation and reduce search requests by n, where n is number of shards being queried in the execution

*Issue #, if available:*
#1409

*Description of changes:*
Use the first shard search request on each shard to compute max sequence number instead of making a search call simply to calculate the value.
Fetch data in descending order of sequence numbers to directly calculate the max seq_no

New heuristic:
take an example of one shard
in current iteration: previous seq_no = 10000, current  max seq_no= 40000

first iteration of search shard(which also determines the max sequence number) : TO=int_max, FROM=prev_seq_no : calculate max_seq number
second iteration : TO=max_seq_no-10000 FROM=prev_seq_no
further iterations : TO=TO-10000 FROM=prev_seq_no till while `to>from`

in first iteration we get first 10k results and also calculate the max sequence number
in second iteration we reduce (TO) variable to maxSeqNo -10000
and query again and we continue this loop until we have seen all docs up untill the previous sequence number


alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -615,7 +620,7 @@ class DocumentLevelMonitorRunner : MonitorRunner() {
         )
     }
 
-    private suspend fun updateLastRunContext(
```
lezzago, line 618: Are we sure we do not need suspend anymore? I recall we needed it before

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -697,51 +674,79 @@ class DocumentLevelMonitorRunner : MonitorRunner() {
...
+        val count: Int = indexExecutionCtx.updatedLastRunContext["shards_count"] as Int
         for (i: Int in 0 until count) {
             val shard = i.toString()
             try {
-                val maxSeqNo: Long = docExecutionCtx.updatedLastRunContext[shard].toString().toLong()
-                val prevSeqNo = docExecutionCtx.lastRunContext[shard].toString().toLongOrNull()
+                val prevSeqNo = indexExecutionCtx.lastRunContext[shard].toString().toLongOrNull()
+                val from = prevSeqNo ?: SequenceNumbers.NO_OPS_PERFORMED
+                var to: Long = Long.MAX_VALUE
+                while (to >= from) {
```
lezzago, line 693: Why do we need this while logic? Also it seems like we reduce the `to` by 10000 from the current fetched seqNo?

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
```diff
@@ -697,51 +674,79 @@ class DocumentLevelMonitorRunner : MonitorRunner() {
...
-                val startTime = System.currentTimeMillis()
-                transformedDocs.addAll(
-                    transformSearchHitsAndReconstructDocs(
-                        hits,
-                        indexName,
-                        concreteIndexName,
-                        monitor.id,
-                        conflictingFields,
+                        updateLastRunContext(shard, hits.hits[0].seqNo.toString())
+                        to = hits.hits[0].seqNo - 10000L
```
lezzago, line 710: Are we only fetching 10,000 docs and is that why we chose 10,000? If so, this is not right. Sequence numbers are incremented for every new write operation on the index and that includes deletion.

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
eirsep: we needed to do suspend earlier because we had a search call to calculate max seq_no being made within the method. 
we have removed the search call as we calculate the max seq_no in the first pass. hence no need of a suspend

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
eirsep: take an example of one shard
in current iteration: `previous seq_no = 10000`, `current  max seq_no= 40000`

**first iteration** of search shard(which also determines the max sequence number) : `TO=int_max`, `FROM=prev_seq_no` : calculate max_seq number
**second iteration** : `TO=max_seq_no-10000` `FROM=prev_seq_no`
**third iteration** : `TO=TO-10000` `FROM=prev_seq_no`

in first iteration we get first 10k results and also calculate the max sequence number
in second iteration we reduce (TO) variable to maxSeqNo -10000
and query again and we continue this loop until we have seen all docs up untill the previous sequence number

alerting/src/main/kotlin/org/opensearch/alerting/DocumentLevelMonitorRunner.kt
eirsep: earlier logic was completely wrong
it calculated the max sequence number and assumed it will always be less than (prev_seq_no+10000) ans hence made only one search call with a hard coded size of from=prev_seq to = max_seq but size =10k

explained the current logic in the above comment: https://github.com/opensearch-project/alerting/pull/1445#discussion_r1505227492

engechas (approved):

sbcd90 (approved):

Issue #1401, opened 2024-02-02
[FEATURE] Enrich alert notification message with context. from individual documents' fields in bucket level monitors


eirsep:
**Is your feature request related to a problem?**
Bucket level monitors contain only the information of the buckets that are created based on aggregation queries. there is no data from individual documents avaialble.

When alerts are generated and user wants to send out notifications, they don't have enough information to enrich the notification message because search response doesn't return the docs and returns only the aggregated data.

For ex. consider a bucket level monitor configured to monitor CPU of 100 hosts crossing 80% in last 1 hr. Say, each doc in index has fields - cpu, hostname, timestamp, eventid, jvm, If CPU>80 for a certain host they would like to fire alert. They can bucket on hostName and condition cpu >80. If condition matches they will get alert for cpu>80 and they will have information about which hosts are causing alert. But they wouldn't have access to other fields of such document

**Possible solution**
If alerts are generated i.e. monitor trigger condition is satisfied AND notifications are configured, Alerting plugin should support a post-trigger search request to collect data to enrich notification message.

That way users have additional context to access  document's other fields based on bucket values

eirsep:
Alternatively explore adding an `exemplar` or a sample document in every bucket level monitor by making size = 10  to get atleast some sample data

AWSHurneyt:
Related issue https://github.com/opensearch-project/alerting/issues/1396

Issue #1396, opened 2024-01-31
[FEATURE] Document Information in Alert Body


brijos:
**Is your feature request related to a problem?**
When I receive an alert, I don't have pertinent information included in the alert notification payload to help my engineers quickly understand what is happening.

**What solution would you like?**
Include information in alerts such as threshold, event time, disk name, free space, etc.

**What alternatives have you considered?**
Subsequently calling the API for additional details.

**Do you have any additional context?**
No

Issue #1300, opened 2023-11-10
[FEATURE] Return alert value information in bucket level ctx for notification messages


AWSHurneyt:
**Is your feature request related to a problem?**
Currently, alerts do not expose the value that triggered the trigger condition in the [asTemplateArgs](https://github.com/opensearch-project/common-utils/blob/2.11/src/main/kotlin/org/opensearch/commons/alerting/model/Alert.kt#L560) function. This prevents notification messages from including that info.

For example, if a bucket monitor is configured to query an index with the following type of document, and generate an alert when the `value` field is `< 50`, we cannot currently include that triggering `value` in the notification body.
```
POST test-index-1/_doc?pipeline=test-timestamp
{
  "name": "Thomas",
  "value": 10
}
```

**What solution would you like?**
Expose the triggering value in the ctx of alerts. This information can be helpful for consumers of notifications for determining what actions to take to resolve an alert.

**Additional context**
Related issue https://github.com/opensearch-project/alerting/issues/1299

AWSHurneyt:
Closing in favor of https://github.com/opensearch-project/alerting/issues/1401

AWSHurneyt:
Reopening as issue #1401 is related to this request, but is different.

PR #1450, opened 2024-03-05
Enhance per bucket, and per document monitor notification message ctx.

AWSHurneyt:
*Issue #, if available:*
https://github.com/opensearch-project/alerting/issues/1300
https://github.com/opensearch-project/alerting/issues/1396
https://github.com/opensearch-project/alerting/issues/1401

*Description of changes:*
1. Added support for returning sample documents for bucket level monitors, and document level monitors.
2. Added support for printing query/rule info in notification messages for document level monitors.

*CheckList:*


eirsep:
typically we create smaller PRs. one for each issue is a practice we should try follow

AWSHurneyt:
The nonsecurity-related test failures seem related to the fixes in PR https://github.com/opensearch-project/alerting/pull/1464

The security-related test failures have been failing for some time, so they're unrelated to these changes.
https://github.com/opensearch-project/alerting/actions/workflows/security-test-workflow.yml?query=branch%3Amain

AWSHurneyt:
> typically we create smaller PRs. one for each issue is a practice we should try follow

@eirsep agreed. Would you say that's blocking for this PR? All of the changes will require the AlertContext class, so the second PR couldn't be raised until the first is merged.

alerting/src/main/kotlin/org/opensearch/alerting/BucketLevelMonitorRunner.kt
```diff
@@ -230,6 +235,33 @@ object BucketLevelMonitorRunner : MonitorRunner() {
...
+                    val searchRequest = monitorCtx.inputService!!.getSearchRequest(
+                        monitor = monitor.copy(triggers = listOf(trigger)),
+                        searchInput = monitor.inputs[0] as SearchInput,
+                        periodStart = periodStart,
+                        periodEnd = periodEnd,
+                        prevResult = monitorResult.inputResults,
+                        matchingDocIdsPerIndex = null,
+                        returnSampleDocs = true
+                    )
+                    monitorCtx.client
```
AWSHurneyt, line 252: This line is dev code that should be removed. I'll be pushing a revision shortly.

alerting/src/main/kotlin/org/opensearch/alerting/BucketLevelMonitorRunner.kt
```diff
@@ -220,6 +224,7 @@ object BucketLevelMonitorRunner : MonitorRunner() {
                     ?.addAll(monitorCtx.alertService!!.convertToCompletedAlerts(keysToAlertsMap))
         }
```
eirsep, line 226: What are the keys of this map variable and the inner map ?
Can we add some code comments to elaborate about how the docs are held in memory

alerting/src/main/kotlin/org/opensearch/alerting/BucketLevelMonitorRunner.kt
```diff
@@ -480,17 +516,88 @@ object BucketLevelMonitorRunner : MonitorRunner() {
...
+    private fun getAlertContext(
+        alert: Alert,
+        alertSampleDocs: Map<String, Map<String, List<Map<String, Any>>>>
+    ): AlertContext {
+        val bucketKey = alert.aggregationResultBucket?.getBucketKeysHash()
+        val sampleDocs = alertSampleDocs[alert.triggerId]?.get(bucketKey)
+        return if (!bucketKey.isNullOrEmpty() && !sampleDocs.isNullOrEmpty()) {
+            AlertContext(alert = alert, sampleDocs = sampleDocs)
+        } else {
+            logger.warn(
```
eirsep, line 542: Error log

alerting/src/main/kotlin/org/opensearch/alerting/BucketLevelMonitorRunner.kt
```diff
@@ -480,17 +516,88 @@ object BucketLevelMonitorRunner : MonitorRunner() {
...
+        client: Client,
+        monitorId: String,
+        triggerId: String,
+        searchRequest: SearchRequest
+    ): Map<String, List<Map<String, Any>>> {
+        val sampleDocumentsByBucket = mutableMapOf<String, List<Map<String, Any>>>()
+
+        val searchResponse: SearchResponse = client.suspendUntil { client.search(searchRequest, it) }
+        val aggs = searchResponse.convertToMap().getOrDefault("aggregations", mapOf<String, Any>()) as Map<String, Any>
+        val compositeAgg = aggs.getOrDefault("composite_agg", mapOf<String, Any>()) as Map<String, Any>
```
eirsep, line 564: Are all aggs of composite type??

alerting/src/main/kotlin/org/opensearch/alerting/BucketLevelMonitorRunner.kt
```diff
@@ -480,17 +516,88 @@ object BucketLevelMonitorRunner : MonitorRunner() {
...
+                alert.triggerId,
+                alert.monitorId,
+                alert.executionId
+            )
+            AlertContext(alert = alert, sampleDocs = listOf())
+        }
+    }
+
+    @Suppress("UNCHECKED_CAST")
+    private suspend fun getSampleDocs(
```
eirsep, line 554: Javadocs

eirsep (changes requested):
Skimmed through the PR. Need to take a deeper look to understand the top hits and low hits logic

alerting/src/main/kotlin/org/opensearch/alerting/BucketLevelMonitorRunner.kt
AWSHurneyt: Good callout. I've added a comment to explain that the `alertSampleDocs` map is `Map<TriggerId, Map<BucketKeysHash, List<Alert>>>`.

alerting/src/main/kotlin/org/opensearch/alerting/BucketLevelMonitorRunner.kt
AWSHurneyt: Revised.

alerting/src/main/kotlin/org/opensearch/alerting/BucketLevelMonitorRunner.kt
AWSHurneyt: Added a comment with details.

alerting/src/main/kotlin/org/opensearch/alerting/BucketLevelMonitorRunner.kt
AWSHurneyt: The frontend visual editor translates the bucket monitor creation form selections into a composite aggregation ([source](https://github.com/opensearch-project/alerting-dashboards-plugin/blob/3b6a9bfc0c8f1b5a1df393714aea695b399ace5c/public/pages/CreateMonitor/containers/CreateMonitor/utils/formikToMonitor.js#L239)); however, different aggregation types could potentially be configured using the extraction query editor. 

As discussed offline, I'll take a follow-up item to support more than composite aggregations.

AWSHurneyt: Created issue https://github.com/opensearch-project/alerting/issues/1473 to track this follow-up item.

alerting/src/main/kotlin/org/opensearch/alerting/script/BucketLevelTriggerExecutionContext.kt
```diff
@@ -5,21 +5,25 @@
...
+
 data class BucketLevelTriggerExecutionContext(
     override val monitor: Monitor,
     val trigger: BucketLevelTrigger,
     override val results: List<Map<String, Any>>,
     override val periodStart: Instant,
     override val periodEnd: Instant,
     val dedupedAlerts: List<Alert> = listOf(),
-    val newAlerts: List<Alert> = listOf(),
+    val newAlerts: List<AlertContext> = listOf(),
```
eirsep, line 26: could have been a separate field to have alert context list and retain new alerts

alerting/src/main/kotlin/org/opensearch/alerting/util/AlertingUtils.kt
```diff
@@ -177,3 +186,85 @@ fun ThreadContext.StoredContext.closeFinally(cause: Throwable?) = when (cause) {
...
+/**
+ * Mustache template supports iterating through a list using a `{{#listVariable}}{{/listVariable}}` block.
+ * https://mustache.github.io/mustache.5.html
+ *
+ * This function looks for `{{#${[AlertContext.SAMPLE_DOCS_FIELD]}}}{{/${[AlertContext.SAMPLE_DOCS_FIELD]}}}` blocks,
+ * and parses the contents for tags, which we interpret as fields within the sample document.
+ *
+ * @return a [Set] of [String]s indicating fields within a document.
+ */
+fun parseSampleDocTags(messageTemplate: Script): Set<String> {
```
eirsep, line 199: we should ideally have a parameter `enrichContextWithSampleDocs` in the create monitor request inside the trigger payload

alerting/src/test/kotlin/org/opensearch/alerting/DocumentMonitorRunnerIT.kt
```diff
@@ -2428,6 +2428,222 @@ class DocumentMonitorRunnerIT : AlertingRestTestCase() {
...
+                    assertTrue(
+                        "The notification message is missing the query name.",
+                        actionOutput["message"]!!.contains("(name=${docQuery.name})")
+                    )
+                }
+            }
+        }
+    }
+
+    fun `test expected document and rules print in notification message`() {
```
eirsep, line 2493: can we add a test for scenario where there is error alert from the trigger evaluation

alerting/src/test/kotlin/org/opensearch/alerting/MonitorRunnerServiceIT.kt
```diff
@@ -1989,6 +1990,88 @@ class MonitorRunnerServiceIT : AlertingRestTestCase() {
         }
     }
 
+    fun `test bucket-level monitor notification message includes sample docs per bucket`() {
```
eirsep, line 1993: plz add more test coverage . scenarios verifying each of the checks to printsSampleDocs

eirsep (approved):

alerting/src/main/kotlin/org/opensearch/alerting/script/BucketLevelTriggerExecutionContext.kt
AWSHurneyt: I'll add this as a follow-up item.

alerting/src/test/kotlin/org/opensearch/alerting/DocumentMonitorRunnerIT.kt
AWSHurneyt: I'll add this as a follow-up item.

alerting/src/test/kotlin/org/opensearch/alerting/MonitorRunnerServiceIT.kt
AWSHurneyt: I'll add this as a follow-up item.

alerting/src/main/kotlin/org/opensearch/alerting/util/AlertingUtils.kt
AWSHurneyt: Agreed. I'll create an issue to track updating the trigger data model to add this boolean-based logic.

AWSHurneyt: Created issue https://github.com/opensearch-project/alerting/issues/1475 to track this action item.

alerting/src/test/kotlin/org/opensearch/alerting/MonitorRunnerServiceIT.kt
AWSHurneyt: Create issue https://github.com/opensearch-project/alerting/issues/1476 to track this follow-up item. For context, the `printsSampleDocs` function does have unit testing ([link](https://github.com/opensearch-project/alerting/pull/1450/files#diff-fb6dad353d75750210919e12bc1019bbfbef87393fee8e54e7642205e75f04a1R55)); but integration testing will increase confidence.

alerting/src/test/kotlin/org/opensearch/alerting/DocumentMonitorRunnerIT.kt
AWSHurneyt: Create issue https://github.com/opensearch-project/alerting/issues/1476 to track this follow-up item.

riysaxen-amzn (approved):

alerting/src/main/kotlin/org/opensearch/alerting/BucketLevelMonitorRunner.kt
```diff
@@ -230,6 +236,32 @@ object BucketLevelMonitorRunner : MonitorRunner() {
                 ?: mutableListOf()
             // Update nextAlerts so the filtered DEDUPED Alerts are reflected for PER_ALERT Action execution
             nextAlerts[trigger.id]?.set(AlertCategory.DEDUPED, dedupedAlerts)
+
+            // Only collect sample docs for triggered triggers, and only when at least 1 action prints sample doc data.
+            val isTriggered = !nextAlerts[trigger.id]?.get(AlertCategory.NEW).isNullOrEmpty()
+            if (isTriggered && printsSampleDocData(trigger)) {
+                try {
+                    val searchRequest = monitorCtx.inputService!!.getSearchRequest(
```
AWSHurneyt, line 244: This current implementation executes the monitor query multiple times; the first time on line 124 to collect the data for trigger evaluation, and then subsequent searches are executed for each triggered trigger in order to collect sample documents.

Ideally, we want to collect the sample documents in the call to `collectInputResults` on line 124 so we can avoid multiple queries as that will improve performance.

AWSHurneyt: Created issue https://github.com/opensearch-project/alerting/issues/1481 to track this follow-up item.

chiarch84:
Dear @AWSHurneyt it is not clear to me weather this should fix or not the fact that {{ctx.results.0.hits.hits.0._source}}  doesn't print anything. I tried OpenSearch version 2.16.0 but it seems not to work as in previous versions. Do I need to use a different syntax to make it work?

AWSHurneyt:
Hi @chiarch84!

This is the documentation, and an example for these enhancements.
https://opensearch.org/docs/latest/observing-your-data/alerting/triggers/#mustache-template-example-1

A concern for us when making these enhancements was that the number of `hits` could easily be several thousand depending on the use case. Printing out a single `_source` from that list of thousands probably wouldn't be an issue, but mustache template supports iterating through lists, and printing values from each entry. Printing data in a notification message for thousands of hits could cause the message to exceed the size restrictions accepted by the destination webhook; causing the message to fail to send. So we added the `sample_documents` variable to the `ctx` to help manage the size of the notification message. We currently return a max of 10 sample docs, but we want to enhance that so it's a configurable cluster setting (no timeline for that just yet though).

chiarch84:
Thank you for your answer @AWSHurneyt 
I tried your suggestion in my OpenSearch version 2.16.0 but I still get an empty message.
Here the details:
Content of the message:
``` 
{{ctx.monitor.name}} just entered alert status. 
Please investigate the issue.

  - Severity: {{ctx.trigger.severity}}
  - Period start: {{ctx.periodStart}}
  - Period end: {{ctx.periodEnd}}
  - Hits:  {{ctx.results.0.hits.total.value}}
  - Alerts Info: {{ctx.results.0.hits}}

Alerts
{{#ctx.alerts}}
    Sample documents:
    {{#sample_documents}}
        Index: {{_index}}
        Document ID: {{_id}}
       
        Exc TYPE: {{_source.bdap_exception_type}}
        -----------------
    {{/sample_documents}}
    {{#associated_queries}}
        Name: {{name}}
        Id: {{id}}
        Tags: {{tags}}
    ------------------------
    {{/associated_queries}}
{{/ctx.alerts}}
```
Previewed message obtained:
```
Monitor BDAP data download - per QUERY just entered alert status. 
Please investigate the issue.

  - Severity: 3
  - Period start: 2024-11-26T12:10:52+01:00
  - Period end: 2024-11-26T12:20:52+01:00
  - Hits:  80
  - Alerts Info: [object Object]

Alerts

```
So even though there are 80 hits, I get nothing in the details part of Alerts.
Thank you for your help!

AWSHurneyt:
@chiarch84 Just to clarify, are you only seeing the issue in the preview of the message, or is the variable missing when an actual message is received after a monitor execution?

I just tested using the 2.16.0 docker image, and was able to successfully print sample docs in the message when my per bucket monitor executed; but it does look like the preview isn't showing the samples correctly (I created issue https://github.com/opensearch-project/alerting/issues/1731 to track that visual bug).

Also, based on the message template you shared, it looks like the monitor type is per query? Unfortunately, the sample docs variable is only available for per bucket, and per document monitors currently.
https://opensearch.org/docs/latest/observing-your-data/alerting/triggers/#:~:text=Only%20available%20with%20bucket%2D%20and%20document%2Dlevel%20monitors.

chiarch84:
@AWSHurneyt in fact my monitor is per query, so I will try with one per bucket and let you know.
I was not seeing the result nor in the preview nor in the email, but probably because it is not the correct type of monitoring.

PR #1663, opened 2024-09-24
add logging for remote monitor execution flows

eirsep:
### Description
add logging for remote monitor execution flows

### DIFF
--- a/spi/src/main/kotlin/org/opensearch/alerting/spi/RemoteMonitorRunner.kt
+++ b/spi/src/main/kotlin/org/opensearch/alerting/spi/RemoteMonitorRunner.kt
@@ -49,7 +49,7 @@ open class RemoteMonitorRunner {
     }
 
     open fun getFanOutAction(): String {
-        return DocLevelMonitorFanOutAction.NAME
+        throw UnsupportedOperationException("Fan out action needs to be implemented by remote monitor.");
     }
 
     open suspend fun doFanOut(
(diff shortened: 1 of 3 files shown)


spi/src/main/kotlin/org/opensearch/alerting/spi/RemoteMonitorRunner.kt
```diff
@@ -49,7 +49,7 @@ open class RemoteMonitorRunner {
     }
 
     open fun getFanOutAction(): String {
-        return DocLevelMonitorFanOutAction.NAME
+        throw UnsupportedOperationException("Fan out action needs to be implemented by remote monitor.");
```
jowg-amazon, line 52: Why is this being thrown as an exception now instead of returning the name?

jowg-amazon (approved):

sbcd90 (approved):

spi/src/main/kotlin/org/opensearch/alerting/spi/RemoteMonitorRunner.kt
eirsep: this class needs to be implemented and registered at
`public Map<String, RemoteMonitorRunner> getMonitorTypesToMonitorRunners()` in Plugin

should not have a default fan out action. needs to be specified by the new monitor created by extending this class

PR #1771, opened 2025-01-16
API to enable/disable a monitor

### LINKED ISSUE #1058: [FEATURE] Need an API to disable/enable monitor
making an update API call to disable or enable a monitor is not the best practice.
Since monitors are jobs and require to be enabled or disabled we should provide one of the following:
1. `PUT _plugins/_alerting/monitors/{monitor_id}/enable` and `PUT _plugins/_alerting/monitors/{monitor_id}/disable`
2. `PUT _plugins/_alerting/monitors/{monitor_id}}/status {"enable" = true/false}`

vikhy-aws:
### Description
This PR adds request and response classes to common-utils to create an API that enables or disables the monitor using the following:

```
PUT _plugins/_alerting/monitors/{monitor_id}/enable
PUT _plugins/_alerting/monitors/{monitor_id}/disable
```
### Related Issues
Resolves #1058 

### Check List

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/resthandler/RestToggleMonitorAction.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/resthandler/RestToggleMonitorAction.kt
@@ -58,8 +58,8 @@ class RestToggleMonitorAction : BaseRestHandler() {
     @Throws(IOException::class)
     override fun prepareRequest(request: RestRequest, client: NodeClient): RestChannelConsumer {
         val monitorId = request.param("monitorID", Monitor.NO_ID)
-        if (request.method() == PUT && Monitor.NO_ID == monitorId) {
-            throw AlertingException.wrap(IllegalArgumentException("Missing monitor ID"))
+        if (Monitor.NO_ID == monitorId) {
+            throw AlertingException.wrap(IllegalArgumentException("Missing monitor ID."))
         }
 
         // Check if the request is being made to enable the monitor
--- a/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportToggleMonitorAction.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportToggleMonitorAction.kt
@@ -62,7 +62,7 @@ class TransportToggleMonitorAction @Inject constructor(
                     try {
                         if (getMonitorResponse.monitor == null) {
                             actionListener.onFailure(
-                                OpenSearchStatusException("Monitor $monitorId not found", RestStatus.NOT_FOUND)
+                                OpenSearchStatusException("Monitor $monitorId not found.", RestStatus.NOT_FOUND)
                             )
                             return
                         }
@@ -70,7 +70,7 @@ class TransportToggleMonitorAction @Inject constructor(
                         if (getMonitorResponse.monitor!!.enabled == enabled) {
                             actionListener.onFailure(
                                 OpenSearchStatusException(
-                                    "Monitor $monitorId is already ${if (enabled) "enabled" else "disabled"}",
+                                    "Monitor $monitorId is already ${if (enabled) "enabled." else "disabled."}",
                                     RestStatus.BAD_REQUEST
                                 )
                             )
--- a/alerting/src/test/kotlin/org/opensearch/alerting/resthandler/MonitorRestApiIT.kt
+++ b/alerting/src/test/kotlin/org/opensearch/alerting/resthandler/MonitorRestApiIT.kt
@@ -1365,6 +1365,38 @@ class MonitorRestApiIT : AlertingRestTestCase() {
         assertNull(enabledTime)
     }
 
+    fun `test enable monitor with query params`() {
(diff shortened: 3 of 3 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/resthandler/RestToggleMonitorAction.kt
```diff
@@ -0,0 +1,103 @@
...
+
+    override fun routes(): List<Route> {
+        return listOf()
+    }
+
+    override fun replacedRoutes(): MutableList<RestHandler.ReplacedRoute> {
+        return mutableListOf(
+            ReplacedRoute(
+                PUT,
+                "${AlertingPlugin.MONITOR_BASE_URI}/{monitorID}/_enable",
```
riysaxen-amzn, line 45: What security level perms are we giving to these APIs, we need to make changes in Security plugin for that, check this -> https://github.com/opensearch-project/security/blob/a1ccbd56621609d1ae0b0f7b4780b968ed06024b/config/roles.yml#L32

alerting/src/main/kotlin/org/opensearch/alerting/resthandler/RestToggleMonitorAction.kt
```diff
@@ -0,0 +1,103 @@
...
+                PUT,
+                "${AlertingPlugin.LEGACY_OPENDISTRO_MONITOR_BASE_URI}/{monitorID}/_disable"
+            )
+        )
+    }
+
+    @Throws(IOException::class)
+    override fun prepareRequest(request: RestRequest, client: NodeClient): RestChannelConsumer {
+        val monitorId = request.param("monitorID", Monitor.NO_ID)
+        if (request.method() == PUT && Monitor.NO_ID == monitorId) {
```
AWSHurneyt, line 61: Is the `request.method() == PUT` check here necessary? The current list of routes only includes `PUT` actions, so I don't think the request will reach your API unless it's using `PUT`.

alerting/src/main/kotlin/org/opensearch/alerting/resthandler/RestToggleMonitorAction.kt
```diff
@@ -0,0 +1,103 @@
...
+            throw AlertingException.wrap(IllegalArgumentException("Missing monitor ID"))
```
AWSHurneyt, line 62: Could you confirm whether this is thrown as a 4xx or a 5xx error? If it's throwing 5xx, let's refactor it to instead throw a 4xx error.

alerting/src/main/kotlin/org/opensearch/alerting/resthandler/RestToggleMonitorAction.kt
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        // Check if the request is being made to enable the monitor
+        val enabled = request.path().endsWith("/_enable")
```
AWSHurneyt, line 66: Most OpenSearch API allow users to append their calls with query params like `?pretty`, which will pretty print the JSON response so it's easier to read. Just to confirm, the `.path()` function excludes those query params, right? I'm pretty sure it does, but just want to make sure. If it includes the query params, you'll need to rethink this check a bit.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportToggleMonitorAction.kt
```diff
@@ -0,0 +1,128 @@
...
+
+        client.execute(
+            GET_MONITOR_ACTION_TYPE,
+            getMonitorRequest,
+            object : ActionListener<GetMonitorResponse> {
+                override fun onResponse(getMonitorResponse: GetMonitorResponse) {
+                    try {
+                        if (getMonitorResponse.monitor == null) {
+                            actionListener.onFailure(
+                                OpenSearchStatusException("Monitor $monitorId not found", RestStatus.NOT_FOUND)
```
AWSHurneyt, line 65: Small observation. As a best practice, we should end error messages with periods (`.`).

alerting/src/main/kotlin/org/opensearch/alerting/resthandler/RestToggleMonitorAction.kt
vikhy-aws: Not required. Picked up this snippet from another file and did not change it. Will update.

alerting/src/main/kotlin/org/opensearch/alerting/resthandler/RestToggleMonitorAction.kt
vikhy-aws: Yes, it excludes all the parameters passed. The current approach works. Will include a test case for this.

alerting/src/main/kotlin/org/opensearch/alerting/resthandler/RestToggleMonitorAction.kt
vikhy-aws: Throws a 400 for all IllegalArgumentExceptions

PR #1815, opened 2025-03-24
change  monitor fan out logic to handle empty request and checking node attributes before fanning out

eirsep:
change  monitor fan out logic to handle empty request and checking node attributes before fanning out

this handles upgrade situation where old version is fanning out to new version and encounters failure in request serialization to silently ignore processing the set of shards assigned

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/MonitorFanOutUtils.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/MonitorFanOutUtils.kt
@@ -1,0 +1,35 @@
+package org.opensearch.alerting
+
+import org.apache.logging.log4j.LogManager
+import org.apache.logging.log4j.Logger
+import org.opensearch.Version
+import org.opensearch.cluster.node.DiscoveryNode
+
+private val logger: Logger = LogManager.getLogger("FanOutEligibility")
+
+fun isNodeEligibleForFanOut(
+    candidateNode: DiscoveryNode,
+    coordinatorNode: DiscoveryNode,
+): Boolean {
+    try {
+        val candidateNodeAttributes: Map<String, Any> = candidateNode.attributes
+        val coordinatorNodeAttributes: Map<String, Any> = coordinatorNode.attributes
+        if (candidateNodeAttributes.containsKey("di_number") && candidateNodeAttributes["di_number"] != null &&
+            coordinatorNodeAttributes.containsKey("di_number") && candidateNodeAttributes["di_number"] != null
+        ) {
+            return candidateNodeAttributes["di_number"] as Int >= coordinatorNodeAttributes["di_number"] as Int
+        }
+    } catch (e: Exception) {
+        logger.error("Error in isNodeEligibleForFanOut criteria evaluation", e)
+        return true
+    }
+    return true
+}
+
+fun getNodes(monitorCtx: MonitorRunnerExecutionContext): Map<String, DiscoveryNode> {
+    val clusterService = monitorCtx.clusterService!!
+    return clusterService.state().nodes.dataNodes.filter {
+        it.value.version >= Version.CURRENT &&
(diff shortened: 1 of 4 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/MonitorFanOutUtils.kt
```diff
@@ -0,0 +1,35 @@
...
+    candidateNode: DiscoveryNode,
+    coordinatorNode: DiscoveryNode,
+): Boolean {
+    try {
+        val candidateNodeAttributes: Map<String, Any> = candidateNode.attributes
+        val coordinatorNodeAttributes: Map<String, Any> = coordinatorNode.attributes
+        if (candidateNodeAttributes.containsKey("di_number") && candidateNodeAttributes["di_number"] != null &&
+            coordinatorNodeAttributes.containsKey("di_number") && candidateNodeAttributes["di_number"] != null
+        ) {
+            return candidateNodeAttributes["di_number"] as Int >= coordinatorNodeAttributes["di_number"] as Int
```
engechas, line 20: Is there a downside to keeping the fanout fully within the same DI (`==` instead of `>=`)? That seems like a safer approach than relying on backwards compatibility in newer DIs

alerting/src/main/kotlin/org/opensearch/alerting/MonitorFanOutUtils.kt
```diff
@@ -0,0 +1,35 @@
...
+
+fun isNodeEligibleForFanOut(
+    candidateNode: DiscoveryNode,
+    coordinatorNode: DiscoveryNode,
+): Boolean {
+    try {
+        val candidateNodeAttributes: Map<String, Any> = candidateNode.attributes
+        val coordinatorNodeAttributes: Map<String, Any> = coordinatorNode.attributes
+        if (candidateNodeAttributes.containsKey("di_number") && candidateNodeAttributes["di_number"] != null &&
+            coordinatorNodeAttributes.containsKey("di_number") && candidateNodeAttributes["di_number"] != null
```
AWSHurneyt, line 18: Not blocking:
I think the use of `containsKey` and `!= null` would be redundant in this context.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDocLevelMonitorFanOutAction.kt
```diff
@@ -207,6 +207,22 @@ class TransportDocLevelMonitorFanOutAction
         listener: ActionListener<DocLevelMonitorFanOutResponse>
     ) {
         try {
+            // handle situations where coordinator node is not on the same opensearch version or
+            // from the same domain instance and return no-op response instead of failing entire fan-out
+            if (request.indexExecutionContext == null) {
```
engechas, line 212: Checking this specific field for null seems brittle. Is there a more definitive way to see if the deserialization failed?

alerting/src/main/kotlin/org/opensearch/alerting/MonitorFanOutUtils.kt
AWSHurneyt, line 20: Would we want this to instead be an equalTo operator, or are we confident that the coodinator node will always be in the old di?

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDocLevelMonitorFanOutAction.kt
riysaxen-amzn: +1

alerting/src/main/kotlin/org/opensearch/alerting/MonitorFanOutUtils.kt
eirsep: yup going with ==

Issue #1880, opened 2025-07-07
[RFC] Alerting V2


toepkerd:
## Overview 
This RFC proposes building V2 APIs for Alerting to support PPL (Piped Processing Language) and simplify the monitor creation experience. 

## Problem Statement 
Current alerting APIs lack PPL support and have a complex user experience for defining multiple monitor types. Users need a more streamlined approach with enhanced query capabilities. 

## Goals
1. Implement PPL language support in alerting 
2. Simplify monitor creation workflow 
3. Provide enhanced trigger condition configuration

### Workflow Architecture
1. Create Monitor  - Define Monitor name  - Configure schedule 
2. Configure Alert Condition  - Query Editor  - Query Input  - Dry Run capability - PPL query support 
3. Configure Notifications  - Define trigger conditions  - Set notification rules

API Shapes

1. Create Monitor v2

* name must be unique.
* Only one schedule definition (period or cron) is allowed.

Endpoint
```
POST /_plugins/_alerting/v2/monitors
```
```
{
  "name": "v2-ppl-monitor-1",                        // (Required) Unique, descriptive name.
  "enabled": true,                               // (Required) Activate monitor on creation.
  "query": "source=_internal (log_level=ERROR OR log_level=FATAL OR log_level=CRITICAL) | stats count by log_level",  // (Required) PPL query string.
  "schedule": {                                  // (Required) Execution schedule (choose one).
    "frequency": "by_interval",                  // (Required) "by_interval" or "cron".
    "period": {                                  // (Required if by_interval) Interval settings.
      "interval": 5,                             // (Required) Interval count.
      "unit": "MINUTES"                          // (Required) MINUTES|HOURS|DAYS.
    }
    /*
    "cron": {                                   // (Required if cron) Cron schedule.
      "expression": "0 * * * *",                // (Required) Cron expression.
      "timezone": "UTC"                         // (Optional) Timezone.
    }
    */
  },
  "labels": {                                    // (Optional) Key/value tags.
    "service": "payment_gateway",                // (Required if used) metadata tag.
    "team": "oreo"                               // (Required if used) metadata tag.
  },
  "triggers": [                                  // (Optional) triggers.
    {
      "name": "high-number-of-fatal-errors",     // (Required) Trigger identifier.
      "condition": "where level='FATAL' | search count > 10", // (Required) PPL (based on language) condition.
      "severity": "critical",                    // (Required) Severity level.
[...]

PR #1968, opened 2025-10-31
PPL Alerting: Delete Monitor, More V1/V2 Separation

toepkerd:
### Description
Delete Monitor functionality and V1/V2 API separation measures

### Related Issues
https://github.com/opensearch-project/alerting/issues/1880

### Check List
- [Y] New functionality includes testing.
- [Y] Commits are signed per the DCO using `--signoff`.

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/AlertingV2Utils.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/AlertingV2Utils.kt
@@ -9,6 +9,8 @@ import org.apache.lucene.search.TotalHits
 import org.apache.lucene.search.TotalHits.Relation
 import org.opensearch.action.search.SearchResponse
 import org.opensearch.action.search.ShardSearchFailure
+import org.opensearch.alerting.AlertingPlugin.Companion.MONITOR_BASE_URI
+import org.opensearch.alerting.AlertingPlugin.Companion.MONITOR_V2_BASE_URI
 import org.opensearch.alerting.modelv2.MonitorV2
 import org.opensearch.commons.alerting.model.Monitor
 import org.opensearch.commons.alerting.model.ScheduledJob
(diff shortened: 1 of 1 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetMonitorAction.kt
```diff
@@ -168,20 +175,6 @@ class TransportGetMonitorAction @Inject constructor(
         }
     }
 
-    // Checks if the exception is caused by an IndexNotFoundException (directly or nested).
-    private fun isIndexNotFoundException(e: Exception): Boolean {
```
lezzago: is this not used?

alerting/src/main/kotlin/org/opensearch/alerting/transportv2/TransportDeleteMonitorV2Action.kt
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        val xcp = XContentHelper.createParser(
+            xContentRegistry, LoggingDeprecationHandler.INSTANCE,
+            getResponse.sourceAsBytesRef, XContentType.JSON
+        )
+        val scheduledJob = ScheduledJob.parse(xcp, getResponse.id, getResponse.version)
+
+        AlertingV2Utils.validateMonitorV2(scheduledJob)?.let {
+            actionListener.onFailure(AlertingException.wrap(it))
```
lezzago: we should inform that the user needs to use the v1 api to delete

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetMonitorAction.kt
toepkerd: The functions are moved from TransportSearchMonitorAction to AlertingV2Utils. This logic isn't actually being deleted. Previous PRs included this util function in AlertingV2Utils, and this later PR catches up to delete it from where it came from.

AWSHurneyt (commented):
Not blocking:
My only comment is that it seems like separating some of these API into versions for V1 and V2 seems to just make managing the plugin assets via API more complex for customers.

Considering the assets these API interact with (V1 and V2 monitors) are all stored in the same index, and each asset has a unique ID, requiring customers to call different API in order to delete/get the asset via that ID seems overly complex. The V1/V2 split SearchMonitor API could probably be accomplished by using a filter either in the query, or a query param in the URI.

AWSHurneyt (approved):

PR #2024, opened 2026-02-19
Adds periodic cleanup job to delete doc level monitor percolate query indices

eirsep:
### Change
: Implement a scheduled cleanup job that: 1. Identifies query indices whose source data indices no longer exist 2. Verifies no active monitors reference the query index 3. Deletes orphaned query indices older than a configurable threshold (e.g., 7 days)
### **Background**
Alerting Doc level monitors leaves clusters experiencing gradual growth of number query indices, leading to: - 100+ query indices per detector accumulating over time - Increased cluster storage costs - Degraded cluster performance and causing high CPU. Manual intervention required to delete old indices

Doc-level monitors (detectors) use a specialized index called a "query index" to efficiently match incoming documents against detection rules. This is called a percolate search query.  Think of it as a reverse index: instead of searching documents for queries, we search queries for matching documents.

Key Characteristics: - Each monitor maintains query indices under an alias (e.g., .opensearch-sap-cloudtrail-detectors-queries-optimized-*) - Query indices mirror field mappings from source data indices to enable query execution - When field limits are exceeded, the system automatically creates a new query index (rollover)

**Problem**: - Timeseries data indices (e.g., log-aws-cloudtrail-000001, 000002, etc.) are regularly rolled over and eventually moved to UltraWarm tier and deleted - Doc-level monitors do not track when source indices are deleted or moved out of hot tier - Query indices created for these indices remain indefinitely - No automated cleanup job exists to remove unused query indices

**Impact**: - Query indices accumulate at the rate of data index rollovers - For daily index rollovers over 3 years: 1000+ query indices - Each query index consumes cluster resources even when no longer needed

**Example Timeline:**

Day 1: log-aws-cloudtrail-000001 created → query index 000001 created
Day 5: log-aws-cloudtrail-000010 created → query index 000002 created
...
**Day 90: log-aws-cloudtrail-000001 deleted (by ISM retention policy)
        → query index 000001 still exists (orphaned)**
Day 1000: 100 query indices exist, 98 are orphaned

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/QueryIndexCleanup.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/QueryIndexCleanup.kt
@@ -28,6 +28,7 @@ import org.opensearch.common.xcontent.XContentHelper
 import org.opensearch.common.xcontent.XContentType
 import org.opensearch.commons.alerting.model.Monitor
 import org.opensearch.commons.alerting.model.MonitorMetadata
+import org.opensearch.commons.alerting.model.ScheduledJob
 import org.opensearch.commons.alerting.model.ScheduledJob.Companion.SCHEDULED_JOBS_INDEX
 import org.opensearch.core.action.ActionListener
 import org.opensearch.core.xcontent.NamedXContentRegistry
@@ -49,6 +50,7 @@ class QueryIndexCleanup(
     private val client: Client,
     private val threadPool: ThreadPool,
     private val clusterService: ClusterService,
+    private val xContentRegistry: NamedXContentRegistry,
 ) : ClusterStateListener {
 
     private val logger = LogManager.getLogger(javaClass)
@@ -121,8 +123,10 @@ class QueryIndexCleanup(
                 val queryIndexUsageMap = buildQueryIndexUsageMap(allMetadata)
                 val indicesToDelete = determineIndicesToDelete(queryIndexUsageMap)
 
+                // Always clean metadata mappings for dead source indices
+                cleanupMetadataMappings(allMetadata, indicesToDelete)
+
                 if (indicesToDelete.isNotEmpty()) {
-                    cleanupMetadataMappings(allMetadata, indicesToDelete)
                     deleteQueryIndices(indicesToDelete)
                 } else {
                     logger.info("No query indices eligible for deletion")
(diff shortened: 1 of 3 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/QueryIndexCleanup.kt
```diff
@@ -0,0 +1,459 @@
...
+    private fun checkIfWriteIndex(aliasName: String, concreteQueryIndex: String): Boolean {
+        val indicesLookup = clusterService.state().metadata().indicesLookup ?: return false
+        val indexAbstraction = indicesLookup.get(aliasName) ?: return false
+        val writeIndexName = indexAbstraction.writeIndex?.index?.name
+        logger.debug("Checking write index for alias $aliasName: writeIndex=$writeIndexName, concrete=$concreteQueryIndex")
+        return writeIndexName == concreteQueryIndex
+    }
+
+    private suspend fun getMonitor(monitorId: String): Monitor? {
+        val getRequest = GetRequest(".opendistro-alerting-config", monitorId)
```
AWSHurneyt: Nitpick:
We probably have a constant somewhere in the package we could reference here.

alerting/src/main/kotlin/org/opensearch/alerting/QueryIndexCleanup.kt
```diff
@@ -0,0 +1,459 @@
...
+        val isWriteIndex: Boolean,
+        val monitorUsages: MutableMap<String, MutableSet<String>>,
+    )
+
+    private suspend fun buildQueryIndexUsageMap(allMetadata: List<MonitorMetadata>): Map<String, QueryIndexUsage> {
+        val queryIndexUsageMap = mutableMapOf<String, QueryIndexUsage>()
+
+        for (metadata in allMetadata) {
+            val monitorId = extractMonitorId(metadata.id)
+            val monitor = getMonitor(monitorId)
```
AWSHurneyt: Nitpick:
Rather than calling `getMonitor` for each metadata, would it make sense to call `searchMonitor` and query using all of the monitorIDs?

alerting/src/main/kotlin/org/opensearch/alerting/QueryIndexCleanup.kt
```diff
@@ -0,0 +1,459 @@
...
+        }
+
+        val retainedCount = queryIndexUsageMap.size - indicesToDelete.size
+        logger.info("Query index cleanup summary: ${indicesToDelete.size} to delete, $retainedCount retained")
+        return indicesToDelete
+    }
+
+    private suspend fun cleanupMetadataMappings(allMetadata: List<MonitorMetadata>, indicesToDelete: List<String>) {
+        for (metadata in allMetadata) {
+            val monitorId = extractMonitorId(metadata.id)
```
AWSHurneyt: Clarification question:
It looks like the `MonitorMetadata` object already has a `monitorId` attribute. Is there a reason to not use that attribute here?
https://github.com/opensearch-project/common-utils/blob/main/src/main/kotlin/org/opensearch/commons/alerting/model/MonitorMetadata.kt#L25

engechas:
A couple high level thoughts on this:
1. Do we need a query index per data index? I recall one single query index was used before and it was moved to an index per to avoid mapping conflicts. Is there a better middle ground here where shared query indices are used so long as there's no conflicts and a new index created only on conflicts?
2. Can we hook into the index rollover, deletion, or transition to warm flows to trigger deletion of the query index rather than relying on a sweeper pattern?
3. If we delete the query index for a warm index, then the index is migrated back to hot - will monitors created against the index continue to function?

alerting/src/main/kotlin/org/opensearch/alerting/util/ValidationHelpers.kt
```diff
@@ -0,0 +1,78 @@
...
+import inet.ipaddr.IPAddressString
+import org.apache.logging.log4j.LogManager
+import java.net.InetAddress
+import java.net.URL
+
+object ValidationHelpers {
+
+    private val logger = LogManager.getLogger()
+
+    const val FQDN_REGEX =
```
engechas: nit: for complex regex like this it's helpful to add a comment explaining what it matches and a few examples

alerting/src/main/kotlin/org/opensearch/alerting/QueryIndexCleanup.kt
```diff
@@ -0,0 +1,459 @@
...
+        return sourceKey.removeSuffix(monitorId)
+    }
+
+    internal fun extractAliasFromConcreteIndex(concreteQueryIndex: String): String {
+        return concreteQueryIndex.substringBeforeLast("-")
+    }
+
+    internal fun extractIndexNumber(concreteQueryIndex: String): Int {
+        return concreteQueryIndex.substringAfterLast("-").toIntOrNull() ?: 0
+    }
```
engechas: This seems brittle. It relies on the rollover to always follows -<number> as a suffix pattern. Things like a daily rollover would not follow this pattern. Is there a more deterministic way to get this information? Maybe the write index of the alias?

alerting/src/main/kotlin/org/opensearch/alerting/QueryIndexCleanup.kt
```diff
@@ -0,0 +1,459 @@
...
+        client.admin().indices().delete(
+            deleteIndexRequest,
+            object : ActionListener<AcknowledgedResponse> {
+                override fun onResponse(response: AcknowledgedResponse) {
+                    logger.info("Successfully deleted query indices: $indicesToDelete")
+                }
+
+                override fun onFailure(e: Exception) {
+                    logger.error("Batch delete failed for: $indicesToDelete. Retrying individually.", e)
+                    deleteQueryIndicesOneByOne(indicesToDelete)
```
engechas: If the bulk delete fails, would we reasonably expect 1-by-1 deletion to succeed?

alerting/src/main/kotlin/org/opensearch/alerting/util/ValidationHelpers.kt
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+package org.opensearch.alerting.util
+
+import inet.ipaddr.IPAddressString
+import org.apache.logging.log4j.LogManager
+import java.net.InetAddress
+import java.net.URL
+
+object ValidationHelpers {
```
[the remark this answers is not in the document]
eirsep: Removing this. 
This is a junk file from my git stash for an older issue. It's not used.

PR #2034, opened 2026-03-06
Add filter by backend roles setting

### LINKED ISSUE #1940: [FEATURE] Add setting to control access to alerting objects by backend roles
**Is your feature request related to a problem?**

We are using [filtering by backend roles to control access to alerting objects](https://docs.opensearch.org/latest/observing-your-data/alerting/security/#advanced-limit-access-by-backend-role). As the documentation describes, the plugin has a hard-coded condition for controlling how user backend roles are evaluated to determine access:

> Now when users view alerting resources in OpenSearch Dashboards (or make REST API calls), they only see monitors and destinations that are created by users who share **at least one backend role**

We can see this condition coded into the plugin here: https://github.com/opensearch-project/alerting/blob/main/alerting/src/main/kotlin/org/opensearch/alerting/transport/SecureTransportAction.kt#L125.

Instead of allowing users who share **one backend role** to see each other's object, we would like to enforce that users can only see monitors and destinations for users who share **all backend roles**.

**What solution would you like?**

I would like a setting to be added that allows users to control how access by backend role is enforced for alerting objects.

The setting could be called `filter_by_backend_roles_access_strategy` (or something) could have the possible values:

- `intersect` - This is the current behavior where users must share at least one backend role to see each other's monitors and destinations
- `all` - This would be the new behavior where users must share all backend roles to see each other's monitors and destinations.

I imagine that this new setting could be added to https://github.com/opensearch-project/alerting/blob/main/alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt.

Based on the value of this setting, the logic in the code around https://github.com/opensearch-project/alerting/blob/main/alerting/src/main/kotlin/org/opensearch/alerting/transport/SecureTransportAction.kt#L125 would change.

**What alternatives have you considered?**

None

**Do you have any additional context?**

markdboyd:
### Description

This PR adds a new plugin setting, `plugins.alerting.filter_by_backend_roles_access_strategy`, which allows users to control how filtering by backend roles works to determine access to alerting objects (monitors).  The options for this setting are:

- `exact` - Users have access to alerting objects if they have **exactly the same (with no additional)** backend roles as the user who created the object
- `intersect` - Users have access to alerting objects if they share **at least one** backend role with the user who created the object
- `all` - Users have access to alerting objects if their backend roles **contain all of the backend roles** of the user who created the object

### Related Issues

Resolves #1940 

### Check List

- [ ] Commits are signed per the DCO using `--signoff`.

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
@@ -432,7 +432,8 @@ internal class AlertingPlugin : PainlessExtension, ActionPlugin, ScriptPlugin, R
             AlertingSettings.COMMENTS_MAX_CONTENT_SIZE,
             AlertingSettings.MAX_COMMENTS_PER_ALERT,
             AlertingSettings.MAX_COMMENTS_PER_NOTIFICATION,
-            AlertingSettings.NOTIFICATION_CONTEXT_RESULTS_ALLOWED_ROLES
+            AlertingSettings.NOTIFICATION_CONTEXT_RESULTS_ALLOWED_ROLES,
+            AlertingSettings.FILTER_BY_BACKEND_ROLES_ACCESS_STRATEGY
         )
     }
 
--- a/alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
@@ -527,7 +527,7 @@ internal class AlertingPlugin : PainlessExtension, ActionPlugin, ScriptPlugin, R
             AlertingSettings.JOB_QUEUE_ACCOUNT_PROVIDER_TYPE,
             AlertingSettings.TARGET_TYPE_TO_SERVICE_NAME,
             AlertingSettings.TENANT_ACCOUNT_ID_HEADER,
-            AlertingSettings.TENANT_RESOURCE_ID_HEADER
+            AlertingSettings.TENANT_RESOURCE_ID_HEADER,
             AlertingSettings.FILTER_BY_BACKEND_ROLES_ACCESS_STRATEGY
         )
     }
--- a/alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
@@ -528,6 +528,7 @@ internal class AlertingPlugin : PainlessExtension, ActionPlugin, ScriptPlugin, R
             AlertingSettings.TARGET_TYPE_TO_SERVICE_NAME,
             AlertingSettings.TENANT_ACCOUNT_ID_HEADER,
             AlertingSettings.TENANT_RESOURCE_ID_HEADER
+            AlertingSettings.FILTER_BY_BACKEND_ROLES_ACCESS_STRATEGY
         )
     }
 
--- a/alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
@@ -462,6 +462,7 @@ internal class AlertingPlugin : PainlessExtension, ActionPlugin, ScriptPlugin, R
             AlertingSettings.REQUEST_TIMEOUT,
             AlertingSettings.MAX_ACTION_THROTTLE_VALUE,
             AlertingSettings.FILTER_BY_BACKEND_ROLES,
+            AlertingSettings.FILTER_BY_BACKEND_ROLES_ACCESS_STRATEGY,
             AlertingSettings.MAX_ACTIONABLE_ALERT_COUNT,
             LegacyOpenDistroAlertingSettings.INPUT_TIMEOUT,
             LegacyOpenDistroAlertingSettings.INDEX_TIMEOUT,
@@ -527,8 +528,7 @@ internal class AlertingPlugin : PainlessExtension, ActionPlugin, ScriptPlugin, R
             AlertingSettings.JOB_QUEUE_ACCOUNT_PROVIDER_TYPE,
             AlertingSettings.TARGET_TYPE_TO_SERVICE_NAME,
             AlertingSettings.TENANT_ACCOUNT_ID_HEADER,
-            AlertingSettings.TENANT_RESOURCE_ID_HEADER,
-            AlertingSettings.FILTER_BY_BACKEND_ROLES_ACCESS_STRATEGY
+            AlertingSettings.TENANT_RESOURCE_ID_HEADER
         )
     }
 
--- a/alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
@@ -302,5 +302,11 @@ class AlertingSettings {
             Setting.Property.NodeScope,
             Setting.Property.Dynamic
         )
+
+        val FILTER_BY_BACKEND_ROLES_ACCESS_STRATEGY = Setting.simpleString(
+            "plugins.alerting.filter_by_backend_roles_access_strategy",
+            "intersect",
+            Setting.Property.NodeScope, Setting.Property.Dynamic
+        )
     }
 }
--- a/alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
@@ -305,8 +305,10 @@ class AlertingSettings {
 
         val FILTER_BY_BACKEND_ROLES_ACCESS_STRATEGY = Setting.simpleString(
             "plugins.alerting.filter_by_backend_roles_access_strategy",
-            "intersect",
-            Setting.Property.NodeScope, Setting.Property.Dynamic
+            FilterByBackendRolesAccessStrategy.INTERSECT.strategy,
+            FilterByBackendRolesAccessStrategyValidator(),
+            Setting.Property.NodeScope,
+            Setting.Property.Dynamic
         )
     }
 }
--- a/alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
@@ -483,5 +483,11 @@ class AlertingSettings {
             "plugins.alerting.tenant.resource_id_header.",
             Setting.Property.NodeScope, Setting.Property.Dynamic
         )
+
+        val FILTER_BY_BACKEND_ROLES_ACCESS_STRATEGY = Setting.simpleString(
+            "plugins.alerting.filter_by_backend_roles_access_strategy",
+            "intersect",
+            Setting.Property.NodeScope, Setting.Property.Dynamic
+        )
     }
 }
--- a/alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
@@ -486,8 +486,10 @@ class AlertingSettings {
 
         val FILTER_BY_BACKEND_ROLES_ACCESS_STRATEGY = Setting.simpleString(
             "plugins.alerting.filter_by_backend_roles_access_strategy",
-            "intersect",
-            Setting.Property.NodeScope, Setting.Property.Dynamic
+            FilterByBackendRolesAccessStrategy.INTERSECT.strategy,
+            FilterByBackendRolesAccessStrategyValidator(),
+            Setting.Property.NodeScope,
+            Setting.Property.Dynamic
         )
     }
 }
--- a/alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
@@ -489,7 +489,8 @@ class AlertingSettings {
             FilterByBackendRolesAccessStrategy.INTERSECT.strategy,
             FilterByBackendRolesAccessStrategyValidator(),
             Setting.Property.NodeScope,
-            Setting.Property.Dynamic
+            Setting.Property.Dynamic,
+            Setting.Property.Sensitive
         )
     }
 }
--- a/alerting/src/main/kotlin/org/opensearch/alerting/settings/FilterByBackendRolesAccessStrategyValidator.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/settings/FilterByBackendRolesAccessStrategyValidator.kt
@@ -12,7 +12,7 @@ class FilterByBackendRolesAccessStrategyValidator : Setting.Validator<String> {
         val allStrategies: List<String> = FilterByBackendRolesAccessStrategy.entries.map { it.strategy }
 
         when (strategy) {
-            FilterByBackendRolesAccessStrategy.ALL.strategy
+            FilterByBackendRolesAccessStrategy.ALL.strategy,
             FilterByBackendRolesAccessStrategy.EXACT.strategy,
             FilterByBackendRolesAccessStrategy.INTERSECT.strategy -> {}
             else -> throw IllegalArgumentException(
--- a/alerting/src/main/kotlin/org/opensearch/alerting/settings/FilterByBackendRolesAccessStrategyValidator.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/settings/FilterByBackendRolesAccessStrategyValidator.kt
@@ -1,0 +1,22 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.alerting.settings
+
+import org.opensearch.common.settings.Setting
+
+class FilterByBackendRolesAccessStrategyValidator : Setting.Validator<String> {
+    override fun validate(strategy: String) {
+        val allStrategies: List<String> = FilterByBackendRolesAccessStrategy.entries.map { it.strategy }
+
+        when (strategy) {
+            FilterByBackendRolesAccessStrategy.INTERSECT.strategy,
+            FilterByBackendRolesAccessStrategy.ALL.strategy -> {}
+            else -> throw IllegalArgumentException(
+                "Setting value must be one of [$allStrategies]"
+            )
+        }
+    }
+}
--- a/alerting/src/main/kotlin/org/opensearch/alerting/settings/FilterByBackendRolesAccessStrategyValidator.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/settings/FilterByBackendRolesAccessStrategyValidator.kt
@@ -12,8 +12,9 @@ class FilterByBackendRolesAccessStrategyValidator : Setting.Validator<String> {
         val allStrategies: List<String> = FilterByBackendRolesAccessStrategy.entries.map { it.strategy }
 
         when (strategy) {
-            FilterByBackendRolesAccessStrategy.INTERSECT.strategy,
-            FilterByBackendRolesAccessStrategy.ALL.strategy -> {}
+            FilterByBackendRolesAccessStrategy.ALL.strategy
+            FilterByBackendRolesAccessStrategy.EXACT.strategy,
+            FilterByBackendRolesAccessStrategy.INTERSECT.strategy -> {}
             else -> throw IllegalArgumentException(
                 "Setting value must be one of [$allStrategies]"
             )
--- a/alerting/src/main/kotlin/org/opensearch/alerting/settings/FilterByBackendRolesAccessStrategyValidator.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/settings/FilterByBackendRolesAccessStrategyValidator.kt
@@ -11,13 +11,8 @@ class FilterByBackendRolesAccessStrategyValidator : Setting.Validator<String> {
     override fun validate(strategy: String) {
         val allStrategies: List<String> = FilterByBackendRolesAccessStrategy.entries.map { it.strategy }
 
-        when (strategy) {
-            FilterByBackendRolesAccessStrategy.ALL.strategy,
-            FilterByBackendRolesAccessStrategy.EXACT.strategy,
-            FilterByBackendRolesAccessStrategy.INTERSECT.strategy -> {}
-            else -> throw IllegalArgumentException(
-                "Setting value must be one of [$allStrategies]"
-            )
+        if (strategy.lowercase() !in allStrategies) {
+            throw IllegalArgumentException("Setting value must be one of [$allStrategies]")
         }
     }
 }
--- a/alerting/src/main/kotlin/org/opensearch/alerting/transport/SecureTransportAction.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/transport/SecureTransportAction.kt
@@ -38,13 +38,13 @@ private val log = LogManager.getLogger(SecureTransportAction::class.java)
 interface SecureTransportAction {
 
     var filterByEnabled: Boolean
-    // var filterByAccessStrategy: String
+    var filterByAccessStrategy: String
 
     fun listenFilterBySettingChange(clusterService: ClusterService) {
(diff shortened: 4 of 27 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/transport/SecureTransportAction.kt
```diff
@@ -100,6 +105,19 @@ interface SecureTransportAction {
         return true
     }
 
+    fun doUserBackendRolesMatchResource(userBackendRoles: List<String>, resourceBackendRoles: List<String>): Boolean {
+        if (filterByAccessStrategy == FilterByBackendRolesAccessStrategy.INTERSECT.strategy) {
+            return resourceBackendRoles.any { it in userBackendRoles }
+        } else if (filterByAccessStrategy == FilterByBackendRolesAccessStrategy.ALL.strategy) {
```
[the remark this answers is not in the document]
markdboyd: @AWSHurneyt - I thought more about this and I agree: `ALL` should provide access if the user roles contain all of the object backend roles, even if the user has additional roles. So for `ALL`, the user would have access in scenarios 1 and 2.

I added new logic to handle the `ALL` strategy. And I added `EXACT` as a separate strategy where the roles must match exactly with no additional roles

alerting/src/main/kotlin/org/opensearch/alerting/transport/SecureTransportAction.kt
```diff
@@ -100,6 +105,21 @@ interface SecureTransportAction {
         return true
     }
 
+    fun doUserBackendRolesMatchResource(userBackendRoles: List<String>, resourceBackendRoles: List<String>): Boolean {
+        if (filterByAccessStrategy == FilterByBackendRolesAccessStrategy.ALL.strategy) {
+            return userBackendRoles.containsAll(resourceBackendRoles)
+        } else if (filterByAccessStrategy == FilterByBackendRolesAccessStrategy.INTERSECT.strategy) {
+            return resourceBackendRoles.any { it in userBackendRoles }
+        } else if (filterByAccessStrategy == FilterByBackendRolesAccessStrategy.EXACT.strategy) {
+            return resourceBackendRoles.sorted().equals(userBackendRoles.sorted())
```
AWSHurneyt: **Nitpick:**
Sorting both arrays before comparing them could probably be replaced by converting them to sets instead.

AWSHurneyt (commented):
Could you also add tests for the following:
1. Editing a monitor with/without the necessary backend roles ([example test](https://github.com/opensearch-project/alerting/blob/main/alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt#L768))
2. Deleting a monitor with/without the necessary backend roles ([example test](https://github.com/opensearch-project/alerting/blob/main/alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt#928))
3. Executing a monitor with/without the necessary backend roles ([example test](https://github.com/opensearch-project/alerting/blob/main/alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt#L1061))
4. Acknowledging an alert with/without the necessary backend roles ([helper function used by various tests to acknowledge alerts](https://github.com/opensearch-project/alerting/blob/main/alerting/src/test/kotlin/org/opensearch/alerting/AlertingRestTestCase.kt#L752))

Properly filtering out assets (e.g., monitors, and alerts) based on backend roles should prevent a user from performing those actions via the UI, but they could still try by using curl commands/devtools if they happen to know the ID for the asset.

markdboyd:
@AWSHurneyt OK, I believe that I have added all of the requested tests

AWSHurneyt:
@markdboyd I've rerun the failing security test a few times, but it still failing unfortunately. Could you take a look?

markdboyd:
@AWSHurneyt OK. I've run 5 of the failing tests from that workflow on my local machine, but none are failing so far.

markdboyd:
@AWSHurneyt OK, I think I have resolved the test failures. The cause of the failures was the changes that I have made in the `TransportAcknowledgeAlertAction.kt` code.

Initially, the problem was that I was having failures on tests acknowledging alerts when filtering by backend roles is enabled. Specifically, the code was returning this error: https://github.com/cloud-gov/opensearch-alerting/blob/dcb6f946bf12da4e9d6124db2f9fa7c059d1bf24/alerting/src/main/kotlin/org/opensearch/alerting/transport/SecureTransportAction.kt#L86-L92

I traced out the error to the fact that `TransportAcknowledgeAlertAction` is doing a request to get a monitor: https://github.com/cloud-gov/opensearch-alerting/blob/dcb6f946bf12da4e9d6124db2f9fa7c059d1bf24/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportAcknowledgeAlertAction.kt#L112. And when filtering by backend role is enabled, this request needs the user information from the parent context, otherwise it fails with the above error.

My initial fix was to pass along just the user information from the parent context manually: 16a21e3b6b7a60cb957b6955ad2d5449f8d63d75. But it seems that solution was not thread safe, so inspired by https://github.com/opensearch-project/alerting/blob/ab154d2ef17f9dbf849433a3e02f703a42168bfa/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetRemoteIndexesAction.kt#L116-L120, I refactored the code to 4c908c01b1f7caef46b024ffcbf0c809bca26eb3. 

However, with that change, we started seeing several integration test failures related to acknowledging alerts. I believe that the problem with 4c908c01b1f7caef46b024ffcbf0c809bca26eb3 is that it was including the user information from the parent context when invoking the `AcknowledgeHandler`. However, the `AcknowledgeHandler` wants to run in the system context without any information about the authenticated user, otherwise it fails to retrieve information about alerts.

So I have now refactored the code again to 298276a13488e8867b31b18f1e467f4330b839ad. The gist of the behavior is now:

- The parent context is stashed here: https://github.com/cloud-gov/opensearch-alerting/blob/dcb6f946bf12da4e9d6124db2f9fa7c059d1bf24/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportAcknowledgeAlertAction.kt#L95C10-L95C57
- We explicitly restore the parent context here: https://github.com/cloud-gov/opensearch-alerting/blob/dcb6f946bf12da4e9d6124db2f9fa7c059d1bf24/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportAcknowledgeAlertAction.kt#L102 so that the get monitor request has access to the user stored in the parent context. the user must be accessible in the context for the get monitor request when filtering by backend role is enabled, otherwise you get this error: https://github.com/cloud-gov/opensearch-alerting/blob/dcb6f946bf12da4e9d6124db2f9fa7c059d1bf24/alerting/src/main/kotlin/org/opensearch/alerting/transport/SecureTransportAction.kt#L86-L92
- Before running the acknowledge handler, we **re-stash the parent context,** because the `AcknowledgeHandler` should be acting as the system, not the authenticated user: https://github.com/cloud-gov/opensearch-alerting/blob/dcb6f946bf12da4e9d6124db2f9fa7c059d1bf24/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportAcknowledgeAlertAction.kt#L129-L131. Without this change, all those security tests fail because they cannot properly acknowledge alerts when acting as the authenticated user

With these changes, I believe that the code should function as expected when filtering by backend role is enabled or not. And that all tests should now pass.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportAcknowledgeAlertAction.kt
```diff
@@ -89,32 +91,49 @@ class TransportAcknowledgeAlertAction @Inject constructor(
...
-                    }
-                if (getMonitorResponse.monitor == null) {
-                    actionListener.onFailure(
-                        AlertingException.wrap(
-                            ResourceNotFoundException(
-                                String.format(
-                                    Locale.ROOT,
-                                    "No monitor found with id [%s]",
-                                    request.monitorId
+                val singleThreadContext = newSingleThreadContext("TransportAcknowledgeAlertAction")
```
AWSHurneyt: I'm not sure this is the correct approach. The `stash -> restore -> re-stash` pattern with `newSingleThreadContext` seems overly complex and differs from how other transport actions in this package handle access checks.

The `DeleteMonitor` API follows a pattern that involves `retrieving the monitor -> determining whether the user has the necessary roles to delete the monitor -> deleting the monitor`. Could this pattern be used here?

https://github.com/opensearch-project/alerting/blob/main/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteMonitorAction.kt#L111

AWSHurneyt: @markdboyd and I discussed the `TransportAcknowledgeAlertAction` changes offline. He called out that the GetMonitor API call was failing within the `stashContext` call.

The better approach looks to follow the `TransportDeleteMonitorAction` pattern of extending the `SecureTransportAction`. I raised this dummy PR in my personal fork with such changes.
https://github.com/AWSHurneyt/OpenSearch-Alerting/pull/14

markdboyd: @AWSHurneyt and I discussed more offline and I implemented the changes from https://github.com/AWSHurneyt/OpenSearch-Alerting/pull/14 in d8fe86e8dd8937bb9b5503fa3ce44eb003f6b909

alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt
```diff
@@ -959,6 +1440,216 @@ class SecureMonitorRestApiIT : AlertingRestTestCase() {
...
+            listOf("role1"),
+            false
+        )
+        val deleteUserClient = SecureRestClientBuilder(clusterHosts.toTypedArray(), isHttps(), deleteUser, password)
+            .setSocketTimeout(60000)
+            .setConnectionRequestTimeout(180000)
+            .build()
+
+        try {
+            deleteMonitor(monitor = createdMonitor)
```
AWSHurneyt: @markdboyd Is this using the admin client instead of the user client? Is that intended?

alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt
```diff
@@ -959,6 +1440,216 @@ class SecureMonitorRestApiIT : AlertingRestTestCase() {
...
```
AWSHurneyt: @markdboyd similar to the above comment, is this using the admin client instead of the user client? Is that intended?

alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt
```diff
@@ -1016,97 +1707,279 @@ class SecureMonitorRestApiIT : AlertingRestTestCase() {
...
+            .setSocketTimeout(60000)
+            .setConnectionRequestTimeout(180000)
+            .build()
+
+        try {
+            executeUserClient?.makeRequest(
+                "POST",
+                "$ALERTING_BASE_URI/${createdMonitor.id}/_execute",
+                mutableMapOf()
+            )
```
AWSHurneyt, line 1842: @markdboyd do we need an `assert` here to confirm the status code and content of the execute API call?

alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt
```diff
@@ -1016,97 +1707,279 @@ class SecureMonitorRestApiIT : AlertingRestTestCase() {
...
 
         try {
-            val executeMonitorResponse = userClient?.makeRequest(
+            executeUserClient?.makeRequest(
                 "POST",
-                "$ALERTING_BASE_URI/${monitor.id}/_execute",
+                "$ALERTING_BASE_URI/${createdMonitor.id}/_execute",
                 mutableMapOf()
             )
-            assertEquals("Get monitor failed", RestStatus.OK, executeMonitorResponse?.restStatus())
```
AWSHurneyt, line 1078: @markdboyd similar to the comment above, do we need an assert here to confirm the status code and content of the execute API call?

alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt
```diff
@@ -1315,6 +2188,194 @@ class SecureMonitorRestApiIT : AlertingRestTestCase() {
...
+        assertNotNull("The monitor was not created", createdMonitor)
+
+        val activeAlert = createAlert(randomAlert(createdMonitor).copy(state = Alert.State.ACTIVE))
+
+        val acknowledgeUser = "acknowledgeUser4"
+        createUserWithRoles(
+            acknowledgeUser,
+            listOf(ALERTING_FULL_ACCESS_ROLE, READALL_AND_MONITOR_ROLE),
+            // does not contain all roles of monitor
+            listOf("role1"),
```
AWSHurneyt: This this instead be `listOf("role1", "role2", "role3")`?

The monitor that's created on line 2350 is configured with roles `listOf("role1", "role2")`. Giving the additional role to the user would better illustrate that the user needs the _**exact**_ same roles as the monitor as opposed to just _**all**_ of the same roles.

alerting/src/test/kotlin/org/opensearch/alerting/settings/FilterByBackendRolesAccessStrategyValidatorTests.kt
```diff
@@ -0,0 +1,26 @@
...
+import org.opensearch.test.OpenSearchTestCase
+import kotlin.test.assertFailsWith
+
+class FilterByBackendRolesAccessStrategyValidatorTests : OpenSearchTestCase() {
+
+    fun `test accepts valid strategies`() {
+        val validator = FilterByBackendRolesAccessStrategyValidator()
+        validator.validate("all")
+        validator.validate("exact")
+        validator.validate("intersect")
```
AWSHurneyt, line 19: Do we expect capitalized and mixed-case inputs to be valid (e.g., `ALL`, and `aLl`)?

alerting/src/test/kotlin/org/opensearch/alerting/settings/FilterByBackendRolesAccessStrategyValidatorTests.kt
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    fun `test rejects invalid strategy`() {
+        val validator = FilterByBackendRolesAccessStrategyValidator()
+        assertFailsWith(IllegalArgumentException::class, "Expected IllegalArgumentException") {
+            validator.validate("invalid")
+        }
```
AWSHurneyt, line 27: Could you add the following cases as well? Unless these are covered elsewhere I missed.
```
validator.validate("")
validator.validate(" ")

File: alerting/src/main/kotlin/org/opensearch/alerting/settings/FilterByBackendRolesAccessStrategyValidator.kt [hunk 11]
```diff
@@ -0,0 +1,23 @@
...
+
+package org.opensearch.alerting.settings
+
+import org.opensearch.common.settings.Setting
+
+class FilterByBackendRolesAccessStrategyValidator : Setting.Validator<String> {
+    override fun validate(strategy: String) {
+        val allStrategies: List<String> = FilterByBackendRolesAccessStrategy.entries.map { it.strategy }
+
+        when (strategy) {
```
AWSHurneyt: **Not blocking:**
Rather than using `when` here, it might be simpler to use `!in`. With the current implementation, cases will need to be added to the `when` block if new strategies are added in the future.

E.g.,
```
if (strategy !in allStrategies) {
   throw IllegalArgumentException("Setting value must be one of [$allStrategies]")
}

File: alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt (hunk 4 above)
Comment (markdboyd) [reply alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt]: @AWSHurneyt You are correct, this test wasn't using the user client correctly. Fixed in 05aa894ae4342eb2616763affcc7f25d03a38a08

File: alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt (hunk 5 above)
Comment (markdboyd) [reply alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt]: @AWSHurneyt You are correct, this test wasn't using the user client correctly. Fixed in 05aa894ae4342eb2616763affcc7f25d03a38a08

File: alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt (hunk 6 above)
Comment (markdboyd) [reply alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt:1842]: @AWSHurneyt Added in 05aa894ae4342eb2616763affcc7f25d03a38a08

File: alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt (hunk 7 above)
Comment (markdboyd) [reply alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt:1078]: @AWSHurneyt Added in 05aa894ae4342eb2616763affcc7f25d03a38a08

File: alerting/src/test/kotlin/org/opensearch/alerting/settings/FilterByBackendRolesAccessStrategyValidatorTests.kt (hunk 9 above)
Comment (markdboyd) [reply alerting/src/test/kotlin/org/opensearch/alerting/settings/FilterByBackendRolesAccessStrategyValidatorTests.kt:19]: I made the validation case insensitive: 2412be4a761f203a7cf1aac2a37321f3e6b7d1eb, however, does this change alone allow a capitalized input to be accepted and used by the system?

File: alerting/src/test/kotlin/org/opensearch/alerting/settings/FilterByBackendRolesAccessStrategyValidatorTests.kt (hunk 10 above)
Comment (markdboyd) [reply alerting/src/test/kotlin/org/opensearch/alerting/settings/FilterByBackendRolesAccessStrategyValidatorTests.kt:27]: Added: 2412be4a761f203a7cf1aac2a37321f3e6b7d1eb

File: alerting/src/main/kotlin/org/opensearch/alerting/settings/FilterByBackendRolesAccessStrategyValidator.kt (hunk 11 above)
Comment (markdboyd) [reply alerting/src/main/kotlin/org/opensearch/alerting/settings/FilterByBackendRolesAccessStrategyValidator.kt]: Good tip! Changed in 2412be4a761f203a7cf1aac2a37321f3e6b7d1eb

File: alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt (hunk 8 above)
Comment (markdboyd) [reply alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt]: Good point. Changed in 05aa894ae4342eb2616763affcc7f25d03a38a08

File: alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt [hunk 12]
```diff
@@ -483,5 +483,13 @@ class AlertingSettings {
             "plugins.alerting.tenant.resource_id_header.",
             Setting.Property.NodeScope, Setting.Property.Dynamic
         )
+
+        val FILTER_BY_BACKEND_ROLES_ACCESS_STRATEGY = Setting.simpleString(
+            "plugins.alerting.filter_by_backend_roles_access_strategy",
+            FilterByBackendRolesAccessStrategy.INTERSECT.strategy,
+            FilterByBackendRolesAccessStrategyValidator(),
+            Setting.Property.NodeScope,
+            Setting.Property.Dynamic
```
cwperks: Same as alerting PR, let's add `Setting.Property.Sensitive` here.

alerting/src/test/kotlin/org/opensearch/alerting/settings/FilterByBackendRolesAccessStrategyValidatorTests.kt
cwperks: @AWSHurneyt since this is a cluster setting, I believe adding mixed case here allows all these permutations for flexibility. Similar to a CLI program accepting y|Y|YES|yes for more flexibility.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexAlertingCommentAction.kt
```diff
@@ -88,6 +88,7 @@ constructor(
     @Volatile private var indexTimeout = INDEX_TIMEOUT.get(settings)
 
     @Volatile override var filterByEnabled = AlertingSettings.FILTER_BY_BACKEND_ROLES.get(settings)
+    @Volatile override var filterByAccessStrategy = AlertingSettings.FILTER_BY_BACKEND_ROLES_ACCESS_STRATEGY.get(settings)
```
cwperks, line 91: I'm not an expert in this repo, but why must this be copied to all of these transport actions?

alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
```diff
@@ -527,7 +527,8 @@ internal class AlertingPlugin : PainlessExtension, ActionPlugin, ScriptPlugin, R
             AlertingSettings.JOB_QUEUE_ACCOUNT_PROVIDER_TYPE,
             AlertingSettings.TARGET_TYPE_TO_SERVICE_NAME,
             AlertingSettings.TENANT_ACCOUNT_ID_HEADER,
-            AlertingSettings.TENANT_RESOURCE_ID_HEADER
+            AlertingSettings.TENANT_RESOURCE_ID_HEADER,
+            AlertingSettings.FILTER_BY_BACKEND_ROLES_ACCESS_STRATEGY
```
cwperks: nit: add this near related FILTER_BY_BACKEND_ROLES setting in this list.

alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt
```diff
@@ -1315,6 +2190,194 @@ class SecureMonitorRestApiIT : AlertingRestTestCase() {
         }
     }
 
+    fun `test acknowledge alerts succeeds when filterByAccessStrategy is all`() {
+        enableFilterBy()
+        putAlertMappings()
+
+        if (!isHttps()) {
```
cwperks, line 2197: btw I think you can add something like https://github.com/opensearch-project/alerting/blob/1b33dbe4242cd3520c078c4fd63c74b0e6593cf1/alerting/src/test/kotlin/org/opensearch/alerting/transport/PluggableDataFormatMonitorBlockIT.kt#L30-L36 at the top of the test suite to skip all tests if running ITs without security enabled.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexAlertingCommentAction.kt
markdboyd: @cwperks I think the reason this variable has to be copied into all of the transport actions is that `filterByAccessStrategy` is defined as a member on the `SecureTransportAction` interface which all of these actions implement: https://github.com/cloud-gov/opensearch-alerting/blob/2412be4a761f203a7cf1aac2a37321f3e6b7d1eb/alerting/src/main/kotlin/org/opensearch/alerting/transport/SecureTransportAction.kt#L42. So in order to satisfy the interface, all of the transport actions must include that member. 

If I remove this line from the code then try to compile the plugin, I get this error:

```
Class 'TransportIndexAlertingCommentAction' is not abstract and does not implement abstract member:
var filterByAccessStrategy: String
```

alerting/src/test/kotlin/org/opensearch/alerting/resthandler/SecureMonitorRestApiIT.kt
markdboyd: @cwperks Not all of the tests in this file seem to require that `https` is enabled. Should they and then I can apply your suggested change?

markdboyd:
Related documentation changes: https://github.com/opensearch-project/documentation-website/pull/12794

AWSHurneyt (approved):

AWSHurneyt:
Confirmed with @cwperks offline that his comments have been addressed. Merging.

PR #2036, opened 2026-03-11
Add configurable max triggers per monitor setting

### LINKED ISSUE #468: Make number of triggers per monitor configurable
Hello everyone,

As mentioned on the [documentation](https://opensearch.org/docs/latest/monitoring-plugins/alerting/monitors/#painless-triggers):
> The cluster metrics monitor supports up to **ten** triggers.

Let me describe my use-case. We have a considerable amount of OpenSearch clusters (~200). To monitor them, we perfmon requests like `GET _cluster/health`, `GET _nodes/stats`, etc. sequentially to all clusters (e.g. every 5 minutes) and store the results into different indices in a dedicated cluster that we use for monitoring. There, we want to create alerts for the different things that can go wrong in our clusters. 

So, targeting the index that holds data of `GET _nodes/stats` calls, and in order to alert us on clusters getting full, we create a monitor with a painless script (`"100.0 * doc['free_in_bytes'].value / doc['total_in_bytes'].value < 15"`) that returns the clusters and the corresponding nodes that are getting full, in the different buckets:

```
"buckets": [
                {
                    "doc_count": 36,
                    "total": {
                        "value": 3702600499200
                    },
                    "nodes": {
                        "doc_count_error_upper_bound": 0,
                        "sum_other_doc_count": 0,
                        "buckets": [
                            {
                                "doc_count": 12,
                                "total": {
                                    "value": 3702600499200
                                },
                                "free": {
                                    "value": 522902073344
                                },
                                "key": "esacluster101_data1"
                            },
                            {
                                "doc_count": 12,
                                "total": {
                                    "value": 3702600499200
                                },
                                "free": {
                                    "value": 513142358016
                                },
                                "key": "esbcluster101_data1"
                            },
                            {
                                "doc_count": 12,
                                "total": {
                                    "value": 3702600499200
                                },
                                "free": {
                                    "value": 308051558400
                                },
                                "key": "esdcluster101_data1"
                            }
                        ]
                    },
                    "free": {
                        "value": 308051558400
                    },
                    "key": "cluster101"
                },
...

```

Now, we want to create 1 trigger for each of the cluster, in order to send the notification on the respective destination.
Here's the trigger condition (the name of the trigger is the cluster name -e.g., 'cluster101'):
```
boolean full= false;
for (int i=0; i < ctx.results[0].aggregations.clusters.buckets.length; i++) {
  if (ctx.results[0].aggregations.clusters.buckets[i].key == ctx.trigger.name) {
    ctx.results[0].full_nodes =ctx.results[0].aggregations.clusters.buckets[i].nodes.buckets;
     full = true;
  }
}
return full;
```

The above process works fine, the only problem is that I can only create 10 triggers for the monitor. Instead of the 200 triggers that I want, I could create 200 monitors, but that sounds even worse to me.

Thus, unless there is an important reason to limit number of triggers allowed in a monitor, I would suggest to make this setting configurable (still having 10 as default value). Let me know what you think.

### LINKED ISSUE #1828: [FEATURE] - Support more than 10 Triggers
**Is your feature request related to a problem?**
Yes. When using Security Analytics, it is common to have a detector with multiple rules. Ideally we should have multiple conditions based on the rule to do multiple determined actions. However, if you need more than 10 triggers it requires a new detector which is not the ideal scenarion.

**What solution would you like?**
Since a detector can have multiple conditions, ideally we shouldn't have a limit on the ammount of triggers. You might have the same rule wanting to do more than 10 different things.

**What alternatives have you considered?**
Using multiple detectors, but it doesn't make sense to scan the same index multiple times just because I need multiple detectors due to the limitation on the amount of triggers.

**Do you have any additional context?**

Image

manaswini1920:
### Description

Add `plugins.alerting.monitor.max_triggers` cluster setting (default: 10, min: 1, dynamic) to allow configuring the maximum number of triggers per monitor. Validation is performed in `TransportIndexMonitorAction` 
before monitor creation.

Companion common-utils PR: https://github.com/opensearch-project/common-utils/pull/913
Resolves #468

### Related Issues
Resolves #1828 #468 
https://github.com/opensearch-project/alerting/issues/1828
https://github.com/opensearch-project/alerting#468

### Testing

End-to-end tested on a local cluster:
- Default limit (10): 11 triggers rejected — "Monitors can only support up to 10 triggers."
- Limit set to 20: 11 triggers accepted and stored
- Limit set to 5: 6 triggers rejected — "Monitors can only support up to 5 triggers."

Testing info here : https://github.com/opensearch-project/common-utils/pull/913#issue-4054949628 

### Check List
- [X] New functionality includes testing.
- [X] Commits are signed per the DCO using `--signoff`.

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
@@ -37,6 +37,7 @@ import org.opensearch.alerting.settings.AlertingSettings
 import org.opensearch.alerting.settings.AlertingSettings.Companion.ALERTING_MAX_MONITORS
 import org.opensearch.alerting.settings.AlertingSettings.Companion.INDEX_TIMEOUT
 import org.opensearch.alerting.settings.AlertingSettings.Companion.MAX_ACTION_THROTTLE_VALUE
+import org.opensearch.alerting.settings.AlertingSettings.Companion.MAX_TRIGGERS_PER_MONITOR
 import org.opensearch.alerting.settings.AlertingSettings.Companion.REQUEST_TIMEOUT
 import org.opensearch.alerting.settings.DestinationSettings.Companion.ALLOW_LIST
 import org.opensearch.alerting.util.DocLevelMonitorQueries
@@ -106,6 +107,7 @@ class TransportIndexMonitorAction @Inject constructor(
     SecureTransportAction {
 
     @Volatile private var maxMonitors = ALERTING_MAX_MONITORS.get(settings)
+    @Volatile private var maxTriggersPerMonitor = MAX_TRIGGERS_PER_MONITOR.get(settings)
     @Volatile private var requestTimeout = REQUEST_TIMEOUT.get(settings)
     @Volatile private var indexTimeout = INDEX_TIMEOUT.get(settings)
     @Volatile private var maxActionThrottle = MAX_ACTION_THROTTLE_VALUE.get(settings)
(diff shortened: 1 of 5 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -407,6 +411,12 @@ class TransportIndexMonitorAction @Inject constructor(
             }
         }
 
+        private fun validateTriggerCount(monitor: Monitor) {
+            require(monitor.triggers.size <= maxTriggersPerMonitor) {
+                "Monitors can only support up to $maxTriggersPerMonitor triggers."
```
AWSHurneyt: Nitpick:
Rather than say monitors can't support X number of triggers, perhaps it would make sense to reword this to something like "The current cluster settings only allow up to $maxTriggersPerMonitor triggers per monitor."

AWSHurneyt (commented):
@manaswini1920 could you add integration tests for this setting?

As mentioned offline, a main test case I have in mind would involve creating a monitor with 10 triggers, confirming it executes properly, then change the cluster setting to a max of 5 triggers, and confirm it still executes as expected.

We want to make sure the changes to break any pre-existing monitors.

PR #2079, opened 2026-04-01
Add new query level trigger evaluation that doesn't depend on ScriptingService or _scripts API 

eirsep:
Adds remote query-level trigger evaluation behind the **plugins.alerting.multi_tenant_trigger_eval_enabled** feature flag (default: false).

When enabled, query-level trigger scripts are evaluated remotely via filter aggregations on the data index instead of locally via ScriptService. This 
rewrites Painless trigger conditions (e.g. ctx.results[0].hits.total.value > 5) into filter aggregations, batches all triggers into a single _search request
with size: 0, and determines trigger state from doc_count. On evaluation error, triggers default to triggered=true (fail-open) so users are notified.

When disabled (default), the existing ScriptService evaluation path is completely unchanged.

Changes:
- AlertingSettings / AlertingPlugin / MonitorRunnerExecutionContext / MonitorRunnerService — register and wire the new boolean cluster setting
- TriggerScriptRewriter — rewrites ctx.results[0] → params.results_0 in trigger scripts
- RemoteQueryLevelTriggerEvaluator — builds filter-agg search request, parses response into QueryLevelTriggerRunResult
- QueryLevelMonitorRunner — branches on feature flag: remote eval path vs existing ScriptService path
- AlertingRestTestCase — adds includeDefaults param to getSettings() helper

- 10 integration tests covering: simple threshold, aggregation value, boolean logic (AND/OR), loop over hits, multiple triggers, script error (fail-open), 
large response, dry run
- 3 regression tests verifying flag=false preserves existing behavior

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/QueryLevelMonitorRunner.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/QueryLevelMonitorRunner.kt
@@ -11,6 +11,7 @@ import org.opensearch.alerting.opensearchapi.InjectorContextElement
 import org.opensearch.alerting.opensearchapi.withClosableContext
 import org.opensearch.alerting.script.QueryLevelTriggerExecutionContext
 import org.opensearch.alerting.settings.AlertingSettings
+import org.opensearch.alerting.trigger.RemoteQueryLevelTriggerEvaluator
 import org.opensearch.alerting.util.CommentsUtils
 import org.opensearch.alerting.util.isADMonitor
 import org.opensearch.commons.alerting.model.Alert
@@ -18,6 +19,7 @@ import org.opensearch.commons.alerting.model.Monitor
 import org.opensearch.commons.alerting.model.MonitorRunResult
 import org.opensearch.commons.alerting.model.QueryLevelTrigger
 import org.opensearch.commons.alerting.model.QueryLevelTriggerRunResult
+import org.opensearch.commons.alerting.model.SearchInput
 import org.opensearch.commons.alerting.model.WorkflowRunContext
 import org.opensearch.transport.TransportService
 import java.time.Instant
@@ -77,6 +79,25 @@ object QueryLevelMonitorRunner : MonitorRunner() {
         val updatedAlerts = mutableListOf<Alert>()
         val triggerResults = mutableMapOf<String, QueryLevelTriggerRunResult>()
 
+        // When multi-tenant trigger eval is enabled, batch-evaluate all query-level triggers
+        // remotely on the customer's cluster instead of running Painless locally
+        val remoteTriggerResults = if (
+            monitorCtx.multiTenantTriggerEvalEnabled &&
+            Monitor.MonitorType.valueOf(monitor.monitorType.uppercase(Locale.ROOT)) == Monitor.MonitorType.QUERY_LEVEL_MONITOR &&
+            monitorResult.inputResults.results.isNotEmpty()
+        ) {
+            val searchInput = monitor.inputs[0] as SearchInput
+            val queryLevelTriggers = monitor.triggers.filterIsInstance<QueryLevelTrigger>()
+            RemoteQueryLevelTriggerEvaluator.evaluate(
+                monitorCtx.client!!,
+                searchInput.indices,
+                queryLevelTriggers,
+                monitorResult.inputResults.results[0]
+            )
+        } else {
+            null
+        }
+
         val maxComments = monitorCtx.clusterService!!.clusterSettings.get(AlertingSettings.MAX_COMMENTS_PER_NOTIFICATION)
         val alertsToExecuteActionsForIds = currentAlerts.mapNotNull { it.value }.map { it.id }
         val allAlertsComments = CommentsUtils.getCommentsForAlertNotification(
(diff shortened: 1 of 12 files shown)


alerting/src/test/kotlin/org/opensearch/alerting/RemoteQueryLevelTriggerIT.kt
```diff
@@ -0,0 +1,392 @@
...
+import java.time.ZonedDateTime
+import java.time.format.DateTimeFormatter
+import java.time.temporal.ChronoUnit
+
+/**
+ * Integration tests for query-level trigger evaluation with the multi-tenant trigger eval flag enabled.
+ * These tests verify that trigger scripts are correctly evaluated remotely via filter aggregations
+ * on the customer's cluster instead of locally via ScriptService.
+ */
+class RemoteQueryLevelTriggerIT : AlertingRestTestCase() {
```
engechas, line 23: Is there a way we can run the full set of QueryLevelTriggerIT with and without the remote flag enabled? Something similar to Junit's @ParametrizedTest annotation. Since the input/output should be the same for both modes, that sort of test would give confidence that the correctness is fully intact in the new mode

alerting/src/main/kotlin/org/opensearch/alerting/QueryLevelMonitorRunner.kt
```diff
@@ -77,6 +79,25 @@ object QueryLevelMonitorRunner : MonitorRunner() {
...
 
+        // When multi-tenant trigger eval is enabled, batch-evaluate all query-level triggers
+        // remotely on the customer's cluster instead of running Painless locally
+        val remoteTriggerResults = if (
+            monitorCtx.multiTenantTriggerEvalEnabled &&
+            Monitor.MonitorType.valueOf(monitor.monitorType.uppercase(Locale.ROOT)) == Monitor.MonitorType.QUERY_LEVEL_MONITOR &&
+            monitorResult.inputResults.results.isNotEmpty()
+        ) {
+            val searchInput = monitor.inputs[0] as SearchInput
+            val queryLevelTriggers = monitor.triggers.filterIsInstance<QueryLevelTrigger>()
```
engechas, line 90: Is it possible for a trigger type to be non-QueryLevelTrigger if the monitor type is QueryLevel?

Wondering why the filterIsInstance is needed/if it excludes any valid triggers

alerting/src/main/kotlin/org/opensearch/alerting/QueryLevelMonitorRunner.kt
```diff
@@ -77,6 +79,25 @@ object QueryLevelMonitorRunner : MonitorRunner() {
...
+            RemoteQueryLevelTriggerEvaluator.evaluate(
+                monitorCtx.client!!,
+                searchInput.indices,
+                queryLevelTriggers,
+                monitorResult.inputResults.results[0]
```
engechas, line 95: There's probably some context from back when they were implemented, but do we know why the results are always a singleton list? It seems like the result set should just be inputResults rather than further nesting as a singleton list

alerting/src/main/kotlin/org/opensearch/alerting/util/TriggerScriptRewriter.kt
```diff
@@ -0,0 +1,26 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.alerting.util
+
+/**
+ * Rewrites Painless trigger scripts for remote evaluation on the customer's cluster.
+ * They are sent to the customer's cluster via a filter aggregation.
+ * The script context on the customer's cluster uses `params` instead of `ctx`,
```
engechas: nit: customer -> user

engechas (approved):
No blocking comments but it'd be great to have the test coverage include all existing cases for the non-remote monitor equivalents. See comment on parametrized test

alerting/src/main/kotlin/org/opensearch/alerting/QueryLevelMonitorRunner.kt
eirsep: well ideologically it can't be anything but query level triggers but just for sanity leaving this here to guardrail against missing this translation fi we add any newer trigger types. 

I added this because there is a possilibity we need to support a DataFusion engine instead of lucene engine and they have some limitations which might force us to introduce new triggers.

alerting/src/main/kotlin/org/opensearch/alerting/QueryLevelMonitorRunner.kt
eirsep: `MONITOR_MAX_INPUTS` is hardcoded to 1 in the alerting plugin, so 
inputResults.results is always a single-element list. It's a list because the data model was designed to support multiple inputs in the future, but the constraint was never relaxed.

alerting/src/test/kotlin/org/opensearch/alerting/RemoteQueryLevelTriggerIT.kt
eirsep: The OpenSearch test framework uses JUnit 4 (via RandomizedRunner), not JUnit 5, so @ParameterizedTest isn't available. I have added regression tests in this same class where I toggle the setting on and off and compare the results and validate that they are the same.

AWSHurneyt (approved):

alerting/src/main/kotlin/org/opensearch/alerting/trigger/RemoteQueryLevelTriggerEvaluator.kt
```diff
@@ -0,0 +1,139 @@
...
+        client: Client,
+        indices: List<String>,
+        triggers: List<QueryLevelTrigger>,
+        searchResponse: Map<String, Any>
+    ): Map<String, QueryLevelTriggerRunResult> {
+        val triggerData = triggers.map { TriggerData(it.id, it.name, it.condition.idOrCode) }
+        val searchSource = buildEvalSearchSource(triggerData, searchResponse)
+
+        return try {
+            val evalRequest = SearchRequest(*indices.toTypedArray()).source(searchSource)
```
AWSHurneyt, line 57: Not blocking:
Do we want to make use of the `plugins.alerting.request_timeout` or one of our other timeout-related cluster settings here?

alerting/src/main/kotlin/org/opensearch/alerting/QueryLevelMonitorRunner.kt
```diff
@@ -77,6 +79,25 @@ object QueryLevelMonitorRunner : MonitorRunner() {
...
         val triggerResults = mutableMapOf<String, QueryLevelTriggerRunResult>()
 
+        // When multi-tenant trigger eval is enabled, batch-evaluate all query-level triggers
+        // remotely on the user's cluster instead of running Painless locally
+        val remoteTriggerResults = if (
+            monitorCtx.multiTenantTriggerEvalEnabled &&
+            Monitor.MonitorType.valueOf(monitor.monitorType.uppercase(Locale.ROOT)) == Monitor.MonitorType.QUERY_LEVEL_MONITOR &&
+            monitorResult.inputResults.results.isNotEmpty()
+        ) {
+            val searchInput = monitor.inputs[0] as SearchInput
```
riysaxen-amzn, line 89: nit: Query-level monitors should always have SearchInput, a ClassCastException here would be unhandled

riysaxen-amzn (approved):

PR #2081, opened 2026-04-02
Use SdkClient for persistence in IndexAlertingComment, IndexMonitor

manaswini1920:
PR title: Use SdkClient for persistence in IndexAlertingComment, IndexMonitor

PR description:
### Description
Complete SDK migration for the final 3 transport actions:
- TransportIndexAlertingCommentAction: put (create/update) + search via SDK
- TransportIndexMonitorAction: get + put (create/update) via SDK

This completes the SDK migration for all 13 transport actions.

### Related Issues
Follows #2053, #2060, #2061
Part of #2094

### Check List
- [x] New functionality includes testing.
- [x] Commits are signed per the DCO using `--signoff`.

PR #2083, opened 2026-04-03
Use SdkClient for persistence in MonitorMetadataService

manaswini1920:
### Description                                                                                                                                                                                                   
Migrates MonitorMetadataService.upsertMetadata() and getMetadata() to use SdkClient instead of direct OpenSearch client calls, as part of the remote metadata SDK migration for background services.

Changes:
- MonitorMetadataService.upsertMetadata(): Replaced IndexRequest with PutDataObjectRequest via sdkClient.putDataObjectAsync()
- MonitorMetadataService.getMetadata(): Replaced GetRequest with GetDataObjectRequest via sdkClient.getDataObjectAsync()
- AlertingPlugin.kt: Set sdkClient on MonitorMetadataService after SDK client creation
- SdkUtils.kt: Added suspend fun CompletionStage<T>.await() helper for coroutine-safe SDK calls (same as #2081)

Notes:
- createFullRunContext() calls (getIndex, indicesStats) are admin/infrastructure operations and remain on direct client
- Routing was previously used on SCHEDULED_JOBS_INDEX but is optional (not _routing.required: true), so SDK migration works without routing support
- Depends on #2081 for SdkUtils.kt

### Related Issues
Part of #2094 

Part of #2094

### Check List
- [x] Commits are signed per the DCO using --signoff.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/alerting/blob/main/CONTRIBUTING.md#developer-certificate
-of-origin).

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
@@ -357,6 +357,7 @@ internal class AlertingPlugin : PainlessExtension, ActionPlugin, ScriptPlugin, R
             xContentRegistry,
             settings
         )
+        MonitorMetadataService.sdkClient = sdkClient
 
         WorkflowMetadataService.initialize(
             client,
--- a/alerting/src/main/kotlin/org/opensearch/alerting/MonitorMetadataService.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/MonitorMetadataService.kt
@@ -13,26 +13,19 @@ import org.apache.logging.log4j.LogManager
 import org.opensearch.ExceptionsHelper
 import org.opensearch.OpenSearchSecurityException
 import org.opensearch.OpenSearchStatusException
-import org.opensearch.action.DocWriteRequest
-import org.opensearch.action.DocWriteResponse
 import org.opensearch.action.admin.indices.get.GetIndexRequest
 import org.opensearch.action.admin.indices.get.GetIndexResponse
 import org.opensearch.action.admin.indices.stats.IndicesStatsAction
 import org.opensearch.action.admin.indices.stats.IndicesStatsRequest
 import org.opensearch.action.admin.indices.stats.IndicesStatsResponse
-import org.opensearch.action.get.GetRequest
-import org.opensearch.action.get.GetResponse
-import org.opensearch.action.index.IndexRequest
-import org.opensearch.action.index.IndexResponse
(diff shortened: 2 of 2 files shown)


engechas (approved):

alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
```diff
@@ -357,6 +357,7 @@ internal class AlertingPlugin : PainlessExtension, ActionPlugin, ScriptPlugin, R
             xContentRegistry,
             settings
         )
+        MonitorMetadataService.sdkClient = sdkClient
```
eirsep, line 360: why aren't we adding it in the constructor?

manaswini1920: Because MonitorMetadataService is a Kotlin object (singleton), not a class — it doesn't have a constructor. It's initialized via MonitorMetadataService.initialize(client, clusterService, xContentRegistry, settings) which sets fields individually. The 
sdkClient follows the same pattern.

You could add it as a parameter to initialize() instead of setting it separately. That would be cleaner:

kotlin
```
fun initialize(
    client: Client,
    clusterService: ClusterService,
    xContentRegistry: NamedXContentRegistry,
    settings: Settings,
    sdkClient: SdkClient
) {
    ...
    this.sdkClient = sdkClient
}


```

i can take it as follow up on the next PR

alerting/src/main/kotlin/org/opensearch/alerting/MonitorMetadataService.kt
```diff
@@ -194,14 +191,23 @@ object MonitorMetadataService :
...
-            if (e.message?.contains("no such index") == true) {
+            if (isIndexNotFoundException(e)) {
                 return null
             } else {
                 throw AlertingException.wrap(e)
             }
         }
     }
 
+    private fun isIndexNotFoundException(e: Throwable): Boolean {
```
eirsep, line 202: isthis a common method used every where. can we move it to commons and add an extra log line

eirsep (changes requested):

eirsep (approved):

alerting/src/main/kotlin/org/opensearch/alerting/MonitorMetadataService.kt
manaswini1920: ack

PR #2095, opened 2026-04-14
feat(scheduler): Disable internal job scheduler when multi-tenancy is enabled

yashgupte21:
When multi_tenancy_enabled is true, the internal job scheduling pipeline (JobSweeper, JobScheduler) is disabled since scheduling is handled externally. The sweeper is still instantiated to satisfy Guice dependency injection but isSweepingEnabled() returns false, making all scheduling operations no-op.

### Description
Disables the internal job scheduling pipeline (JobSweeper, JobScheduler) when multi_tenancy_enabled is true. When `plugins.alerting.multi_tenancy_enabled` is true , scheduling is handled externally. The sweeper is still instantiated to satisfy Guice dependency injection but isSweepingEnabled() returns false, making all scheduling operations no-op.

### Related Issues

### Check List

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
@@ -333,7 +333,7 @@ internal class AlertingPlugin : PainlessExtension, ActionPlugin, ScriptPlugin, R
         commentsIndices = CommentsIndices(environment.settings(), client, threadPool, clusterService)
         docLevelMonitorQueries = DocLevelMonitorQueries(client, clusterService)
         scheduler = JobScheduler(threadPool, runner)
-        sweeper = JobSweeper(environment.settings(), client, clusterService, threadPool, xContentRegistry, scheduler, ALERTING_JOB_TYPES)
+        sweeper = JobSweeper(environment.settings(), client, clusterService, threadPool, xContentRegistry, scheduler, ALERTING_JOB_TYPES, MULTI_TENANCY_ENABLED)
         destinationMigrationCoordinator = DestinationMigrationCoordinator(client, clusterService, threadPool, scheduledJobIndices)
         this.threadPool = threadPool
         this.clusterService = clusterService
--- a/core/src/main/kotlin/org/opensearch/alerting/core/JobSweeper.kt
+++ b/core/src/main/kotlin/org/opensearch/alerting/core/JobSweeper.kt
@@ -25,6 +25,7 @@ import org.opensearch.cluster.service.ClusterService
 import org.opensearch.common.lifecycle.LifecycleListener
 import org.opensearch.common.logging.Loggers
 import org.opensearch.common.lucene.uid.Versions
+import org.opensearch.common.settings.Setting
 import org.opensearch.common.settings.Settings
 import org.opensearch.common.unit.TimeValue
 import org.opensearch.common.util.concurrent.OpenSearchExecutors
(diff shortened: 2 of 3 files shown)


core/src/main/kotlin/org/opensearch/alerting/core/JobSweeper.kt
engechas, line 124: Should we add an update consumer for the multi-tenancy setting that enables the sweeper?

engechas, line 211: Do we need a multiTenancyEnabled check here as well?

yashgupte21: `MULTI_TENANCY_ENABLED` is nodeScope and final in AlertingSettings. This value would not change while the node is running.

yashgupte21: `initBackgroundSweep()` already short-circuits via `isSweepingEnabled()`, but adding an explicit guard here will make it clear. Will add.

alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
```diff
@@ -347,18 +333,31 @@ internal class AlertingPlugin : PainlessExtension, ActionPlugin, ScriptPlugin, R
         commentsIndices = CommentsIndices(environment.settings(), client, threadPool, clusterService)
         docLevelMonitorQueries = DocLevelMonitorQueries(client, clusterService)
         scheduler = JobScheduler(threadPool, runner)
-        sweeper = JobSweeper(environment.settings(), client, clusterService, threadPool, xContentRegistry, scheduler, ALERTING_JOB_TYPES)
+        sweeper = JobSweeper(environment.settings(), client, clusterService, threadPool, xContentRegistry, scheduler, ALERTING_JOB_TYPES, MULTI_TENANCY_ENABLED.get(settings))
```
engechas: Rather than passing this into the constructor, can we add a settings consumer directly into the JobSweeper? This would follow the same pattern as the other settings that it relies on and allow changes after the node is running

core/src/main/kotlin/org/opensearch/alerting/core/JobSweeper.kt
engechas: It should be able to change though, similar to other cluster settings. See comment above

alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
yashgupte21: Ack, will make this change.

core/src/main/kotlin/org/opensearch/alerting/core/JobSweeper.kt
```diff
@@ -76,7 +76,8 @@ class JobSweeper(
     private val threadPool: ThreadPool,
     private val xContentRegistry: NamedXContentRegistry,
     private val scheduler: JobScheduler,
-    private val sweepableJobTypes: List<String>
+    private val sweepableJobTypes: List<String>,
+    private val multiTenancyEnabled: Boolean = false
```
riysaxen-amzn:
```
@Volatile private var multiTenancyEnabled: Boolean = false

// In initSweeper():
clusterService.clusterSettings.addSettingsUpdateConsumer(MULTI_TENANCY_ENABLED) { multiTenancyEnabled = it }
```

yashgupte21: Addressing this by passing the `Setting<Boolean>` into the constructor (since JobSweeper is in core and can't import from alerting), reading it via `.get(settings)`. Same pattern as the other settings.

engechas (approved):

riysaxen-amzn (approved):

PR #2096, opened 2026-04-14
Add EventBridge Scheduler CRUD for external monitor scheduling

riysaxen-amzn:
### Description

Adds infrastructure for managing EventBridge Scheduler schedules as part of the
OpenSearch Alerting external monitoring workflow. When a monitor is
created, updated, or deleted, a corresponding EventBridge schedule is managed in the
customer's EB cell account via cross-account AssumeRole.

### Changes

**New files:**
- `ExternalSchedulerService.kt` — Concrete service object for EB schedule CRUD (create/update/delete). SDK calls are TODO stubs pending OSSA-608 wiring.
- `ScheduleTranslator.kt` — Converts OpenSearch `IntervalSchedule`/`CronSchedule` to EventBridge `rate()`/`cron()` expressions. Handles seconds→minutes ceiling division and Unix→EB day-of-week mapping.
- `MonitorPayloadBuilder.kt` — Builds the SQS message payload (gzip+base64 encoded monitor config) for the EB schedule target, within SQS's 256KB limit.

**Modified files:**
- `TransportIndexMonitorAction.kt` — Wires `createSchedule`/`updateSchedule` into monitor create/update flows. Extracts EB cell info from request headers with fail-fast validation (no empty string defaults).
- `TransportDeleteMonitorAction.kt` — Wires `deleteSchedule` into monitor delete flow. EB schedule is deleted **before** monitor metadata to preserve the monitor record if EB cleanup fails.

**Tests:**
- `ExternalSchedulerServiceTests.kt` — Unit tests for schedule naming, CRUD no-throw verification, and `ScheduleRequest` data class field coverage.
- `ScheduleTranslatorTests.kt` — Unit tests for interval→rate conversion (including ceiling division for non-divisible seconds), cron expression translation, and day-of-week mapping.

### Check List
- [x] New functionality includes testing
- [x] New functionality has been documented above
- [x] Commits are signed per the DCO using `--signoff`


alerting/src/main/kotlin/org/opensearch/alerting/util/ScheduleTranslator.kt
```diff
@@ -0,0 +1,104 @@
...
+            is IntervalSchedule -> Pair(translateInterval(schedule), null)
+            is CronSchedule -> Pair(translateCron(schedule), schedule.timezone)
+        }
+    }
+
+    private fun translateInterval(schedule: IntervalSchedule): String {
+        val unit = when (schedule.unit) {
+            ChronoUnit.SECONDS -> {
+                // EB doesn't support seconds — convert to minutes (ceiling division, minimum 1)
+                val minutes = maxOf(1, (schedule.interval + 59) / 60)
```
engechas: Should we throw an exception if the interval was less than one minute? It's better to be upfront that the request parameters are not supported rather than silently changing the behavior under the hood

alerting/src/main/kotlin/org/opensearch/alerting/util/MonitorPayloadBuilder.kt
```diff
@@ -0,0 +1,64 @@
...
+ * Builds the SQS message payload (EB Target.Input) per LLD section 5.2.
+ *
+ * Top-level fields for routing, nested monitorConfig (gzip+base64) with full monitor definition.
+ * job_start_time uses EB's <aws.scheduler.scheduled-time> context attribute.
+ */
+object MonitorPayloadBuilder {
+
+    private const val GZIP_THRESHOLD = 4096 // bytes — only compress if config exceeds this
+
+    fun buildTargetInput(
```
engechas: Why do we need encoding, compression, and XContentFactory for this?

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -565,6 +567,15 @@ class TransportIndexMonitorAction @Inject constructor(
                     throw t
                 }
 
+                // OSSA-606: Create EB schedule for monitor execution (OASIS path)
```
engechas: This comment should be removed

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -565,6 +567,15 @@ class TransportIndexMonitorAction @Inject constructor(
...
+                try {
+                    createExternalSchedule(request.monitor, tenantId)
```
engechas: Does this need to be behind the flag that enables external scheduling?

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -783,5 +802,61 @@ class TransportIndexMonitorAction @Inject constructor(
...
+        /**
+         * Reads EB cell info from ThreadContext and creates an external schedule.
+         * No-op when OASIS ActionFilter is not present (open-source standalone mode).
+         */
+        private fun createExternalSchedule(monitor: Monitor, tenantId: String?) {
+            val threadContext = client.threadPool().threadContext
+            val ebCellAccountId = threadContext.getTransient<String>(ExternalSchedulerService.EB_CELL_ACCOUNT_ID_KEY) ?: return
+            val ebCellRegion = threadContext.getTransient<String>(ExternalSchedulerService.EB_CELL_REGION_KEY) ?: return
+            val queueArn = threadContext.getTransient<String>(ExternalSchedulerService.EB_CELL_QUEUE_ARN_KEY) ?: return
+            val roleArn = threadContext.getTransient<String>(ExternalSchedulerService.EB_CELL_ROLE_ARN_KEY) ?: return
```
engechas: These should come from plugin settings and don't need to include `cell` in the name. The ebAccountId should additionally check if there is an entry in the threadcontext and override the plugin setting value if so

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -783,5 +802,61 @@ class TransportIndexMonitorAction @Inject constructor(
...
+
+            val targetInput = MonitorPayloadBuilder.buildTargetInput(
+                monitor = monitor,
+                appId = threadContext.getHeader("x-app-id")
+                    ?: throw OpenSearchStatusException("x-app-id header missing", RestStatus.BAD_REQUEST),
+                tenantId = tenantId
+                    ?: throw OpenSearchStatusException("Tenant ID missing", RestStatus.BAD_REQUEST),
+                workspaceId = threadContext.getHeader("x-workspace-id")
+                    ?: throw OpenSearchStatusException("x-workspace-id header missing", RestStatus.BAD_REQUEST),
```
engechas: What is the value of tenant ID? I am wondering if we need app/workspace or tenant is the differentiator

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -783,5 +802,61 @@ class TransportIndexMonitorAction @Inject constructor(
...
+                collectionEndpoint = threadContext.getHeader("x-collection-endpoint")
+                    ?: throw OpenSearchStatusException("x-collection-endpoint header missing", RestStatus.BAD_REQUEST)
+            )
```
engechas: This is all duplicated with the above. Can we push this logic into the ExternalSchedulerService and create a private method for it to remove the duplication

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -783,5 +802,61 @@ class TransportIndexMonitorAction @Inject constructor(
             }
             return null
         }
+
+        /**
+         * Reads EB cell info from ThreadContext and creates an external schedule.
+         * No-op when OASIS ActionFilter is not present (open-source standalone mode).
```
engechas: Comments need to be cleaned up. Applies throughout

alerting/src/main/kotlin/org/opensearch/alerting/service/ExternalSchedulerService.kt
```diff
@@ -0,0 +1,114 @@
...
+ */
+object ExternalSchedulerService {
+
+    private val log = LogManager.getLogger(ExternalSchedulerService::class.java)
+
+    const val SCHEDULE_NAME_PREFIX = "monitor-"
+    const val EB_CELL_ACCOUNT_ID_KEY = "oasis.eb_cell_account_id"
+    const val EB_CELL_REGION_KEY = "oasis.eb_cell_region"
+    const val EB_CELL_QUEUE_ARN_KEY = "oasis.eb_cell_queue_arn"
+    const val EB_CELL_ROLE_ARN_KEY = "oasis.eb_cell_role_arn"
```
engechas: See comment below which is related to this comment

We need the follow settings:
* Scheduler region
* Scheduler queue name
* Schedule role name
* Scheduler account ID
* Assume role name (optional)

With this information we can construct the ARNs for queue, scheduler role, and assume role.

If assume role name is specified, we construct the SDK client using StsAssumeRoleCredsProvider and the constructed assume role ARN. Otherwise it should use the default credentials provider

alerting/src/main/kotlin/org/opensearch/alerting/service/ExternalSchedulerService.kt
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * Manages external EventBridge schedules for OASIS monitor execution.
+ *
+ * Called from TransportIndexMonitorAction (create/update) and
+ * TransportDeleteMonitorAction (delete) during monitor CRUD.
+ *
+ * AWS SDK SchedulerClient calls will be wired in OSSA-608.
+ */
+object ExternalSchedulerService {
```
engechas, line 18: The logic to interact with EventBridge as an external scheduler is used by multiple plugins. This logic should be in common util

alerting/src/main/kotlin/org/opensearch/alerting/service/ExternalSchedulerService.kt
```diff
@@ -0,0 +1,114 @@
...
+        crossAccountRoleArn: String
+    ) {
+        log.info("Deleting EB schedule ${scheduleName(monitorId)} from cell $ebCellAccountId")
+        // TODO OSSA-608: SchedulerClient.deleteSchedule() with AssumeRole into EB cell account
+    }
+
+    /**
+     * Data class carrying all info needed to create/update an EB schedule.
+     */
+    data class ScheduleRequest(
```
engechas: Do we need a data class for this? We should be building the CreateScheduleRequest that needs this information as part of this class. If the information is not leaving this class then a data class seems unnecessary when we could instead construct the CreateScheduleRequest directly

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteMonitorAction.kt
```diff
@@ -120,6 +130,35 @@ class TransportDeleteMonitorAction @Inject constructor(
...
+        /**
+         * Writes EB cell info to ThreadContext so the OASIS post-filter can
+         * read it to decrement scheduleCount in DDB (LLD 6.3).
+         */
+        private fun writeEbCellToThreadContext(monitor: Monitor) {
+            // TODO OSSA-608: extract ebCellAccountId from monitor metadata once
+            // Neo DS alerting metadata fields are available. For now, the OASIS
+            // pre-filter already populates ThreadContext during cell selection.
+            val threadContext = client.threadPool().threadContext
+            val ebCellAccountId = threadContext.getTransient<String>(ExternalSchedulerService.EB_CELL_ACCOUNT_ID_KEY)
```
engechas: This needs to match the value supplied for IndexMonitor

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteMonitorAction.kt
```diff
@@ -120,6 +130,35 @@ class TransportDeleteMonitorAction @Inject constructor(
...
+            if (ebCellAccountId != null) {
```
engechas: We should log even if this is null for debugging

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteMonitorAction.kt
```diff
@@ -120,6 +130,35 @@ class TransportDeleteMonitorAction @Inject constructor(
...
+        /**
+         * Deletes the EB schedule BEFORE deleting the monitor from Neo DS.
+         * Fails the request if cleanup cannot complete — preserving the monitor
+         * record so the schedule deletion can be retried.
+         */
+        private fun deleteExternalSchedule(monitor: Monitor) {
+            val threadContext = client.threadPool().threadContext
+            val ebCellAccountId = threadContext.getTransient<String>(ExternalSchedulerService.EB_CELL_ACCOUNT_ID_KEY) ?: return
+            val ebCellRegion = threadContext.getTransient<String>(ExternalSchedulerService.EB_CELL_REGION_KEY) ?: return
+            val roleArn = threadContext.getTransient<String>(ExternalSchedulerService.EB_CELL_ROLE_ARN_KEY) ?: return
```
engechas: For delete/update this information should come from the monitor metadata

engechas (changes requested):

alerting/src/main/kotlin/org/opensearch/alerting/service/ExternalSchedulerService.kt
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    /** Returns account ID from ThreadContext override if present, otherwise plugin setting. */
+    private fun resolveAccountId(): String {
+        return client.threadPool().threadContext.getTransient<String>(ACCOUNT_ID_OVERRIDE_KEY) ?: accountId
+    }
+
+    /** Returns region from ThreadContext override if present, otherwise plugin setting. */
+    private fun resolveRegion(): String {
+        return client.threadPool().threadContext.getTransient<String>(REGION_OVERRIDE_KEY) ?: region
```
engechas: The threadContext override here shouldn't be necessary

alerting/src/main/kotlin/org/opensearch/alerting/service/ExternalSchedulerService.kt
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * Manages external EventBridge schedules for monitor execution.
+ *
+ * Reads scheduler configuration from plugin settings, with optional
+ * ThreadContext overrides for account ID and region (set by routing layer).
+ * Gated by the `plugins.alerting.external_scheduler.enabled` feature flag.
+ *
+ * TODO: Wire SDK SchedulerClient calls in a follow-up PR.
+ */
+class ExternalSchedulerService(
```
engechas: This class would also be shared across plugins that support external scheduling. It could/should be moved to common-utils so it can be reused without a dependency on alerting

The drawback would be requiring anyone with a dependency on common utils to take a transitive dependency on the eventbridge SDK library

alerting/src/main/kotlin/org/opensearch/alerting/service/ExternalSchedulerService.kt
```diff
@@ -0,0 +1,139 @@
...
+
+    private fun queueArn(): String = "arn:aws:sqs:${resolveRegion()}:${resolveAccountId()}:$queueName"
+    private fun roleArn(): String = "arn:aws:iam::${resolveAccountId()}:role/$roleName"
+
+    /**
+     * Creates an EventBridge schedule for a newly created monitor.
+     * No-op when external scheduling is disabled.
+     */
+    fun createSchedule(monitor: Monitor) {
+        if (!enabled) return
```
engechas: Should this check be in the caller instead of here?

If we leave it here it should throw an exception. Otherwise the caller doesn't know that no schedule was created

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteMonitorAction.kt
```diff
@@ -103,6 +105,10 @@ class TransportDeleteMonitorAction @Inject constructor(
                     )
                     return
                 } else if (canDelete) {
+                    // Delete EB schedule FIRST — if this fails, the monitor record is
+                    // preserved so the schedule can be retried.
+                    deleteExternalSchedule(monitor)
```
engechas: The check for whether the external scheduler is enabled would make more sense here imo

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteMonitorAction.kt
```diff
@@ -120,6 +126,10 @@ class TransportDeleteMonitorAction @Inject constructor(
             }
         }
 
+        /** Deletes the EB schedule BEFORE deleting the monitor record. No-op when disabled. */
+        private fun deleteExternalSchedule(monitor: Monitor) =
+            externalSchedulerService.deleteSchedule(monitor.id)
```
engechas: nit: do we need a new method for this 1-liner?

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -761,6 +772,14 @@ class TransportIndexMonitorAction @Inject constructor(
                     )
                     MonitorMetadataService.upsertMetadata(updatedMetadata, updating = true)
                 }
+                // Update external schedule with latest monitor config
+                try {
+                    updateExternalSchedule(request.monitor)
+                } catch (t: Exception) {
+                    log.error("Failed to update EB schedule for monitor ${request.monitorId}", t)
+                    // Update flow: EB failure is non-fatal — monitor metadata is already persisted.
+                    // The schedule will be reconciled on next update or by auditor.
```
engechas: What is the auditor?

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -761,6 +772,14 @@ class TransportIndexMonitorAction @Inject constructor(
                     )
                     MonitorMetadataService.upsertMetadata(updatedMetadata, updating = true)
                 }
+                // Update external schedule with latest monitor config
+                try {
+                    updateExternalSchedule(request.monitor)
+                } catch (t: Exception) {
+                    log.error("Failed to update EB schedule for monitor ${request.monitorId}", t)
```
engechas: I don't think swallowing the exception is the right behavior here. If the schedule is not updated, the monitor is not updated

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -783,5 +802,13 @@ class TransportIndexMonitorAction @Inject constructor(
             }
             return null
         }
+
+        /** Creates an external EB schedule. No-op when disabled. */
+        private fun createExternalSchedule(monitor: Monitor) =
+            externalSchedulerService.createSchedule(monitor)
```
engechas: Same nit as in DeleteMonitorAction, applies below too

alerting/src/test/kotlin/org/opensearch/alerting/service/ExternalSchedulerServiceTests.kt
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.alerting.service
+
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class ExternalSchedulerServiceTests {
```
engechas, line 18: Let's add unit tests for the other methods as well

We should also update the existing tests for index/delete monitor action to cover the new logic

alerting/src/main/kotlin/org/opensearch/alerting/service/ExternalSchedulerService.kt
riysaxen-amzn: The reusable utilities (ScheduleTranslator, SchedulePayloadBuilder) have been moved to common-utils in PR #939 (now merged). The ExternalSchedulerService remains in the alerting plugin because it contains plugin-specific orchestration (AWS SDK calls, AssumeRole, plugin settings) — keeping common-utils SDK-free per our convention. Other plugins can reuse the common-utils utilities and implement their own service layer.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteMonitorAction.kt
riysaxen-amzn: adressed In commit b9d7fee ("Gate external scheduler hooks behind plugins.alerting.external_scheduler.enabled"), the check is moved to the caller site in both transport actions

alerting/src/test/kotlin/org/opensearch/alerting/service/ExternalSchedulerServiceTests.kt
riysaxen-amzn: added

alerting/src/main/kotlin/org/opensearch/alerting/service/SchedulerRoutingResolver.kt
```diff
@@ -0,0 +1,51 @@
...
+ * and is shared by [org.opensearch.alerting.transport.TransportIndexMonitorAction] and
+ * [org.opensearch.alerting.transport.TransportDeleteMonitorAction].
+ */
+object SchedulerRoutingResolver {
+
+    /** Full routing info needed by create/update. Null when any required field is blank. */
+    data class Routing(val accountId: String, val queueArn: String, val roleArn: String)
+
+    /** Delete path only needs accountId + roleArn. Null when either is blank. */
+    data class DeleteRouting(val accountId: String, val roleArn: String)
```
engechas: minor: There's enough overlap that we can just reuse the model with queueArn not being used in the delete flow. That would then simplify this

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteMonitorAction.kt
```diff
@@ -120,6 +142,31 @@ class TransportDeleteMonitorAction @Inject constructor(
             }
         }
 
+        /**
+         * Logs scheduler routing info from ThreadContext for observability.
+         */
+        private fun logSchedulerContext(monitor: Monitor) {
```
engechas: We should log this info in the service class which has the concrete details about where the scheduler is being created

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteMonitorAction.kt
engechas: This doesn't appear to be addressed in the latest changes

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -798,5 +838,45 @@ class TransportIndexMonitorAction @Inject constructor(
...
+        /**
+         * Reads scheduler routing info from plugin settings (with optional ThreadContext
+         * override for account id) and creates an external schedule.
+         * No-op when required settings are blank.
+         */
+        private fun createExternalSchedule(monitor: Monitor, tenantId: String?) {
+            val routing = resolveRouting() ?: return
+            val targetInput = SchedulePayloadBuilder.buildTargetInput(
+                monitor = monitor,
+                jobStartTimePlaceholder = "<aws.scheduler.scheduled-time>"
```
engechas: minor: this is presumably a constant. We can push the logic into the SchedulePayloadBuilder

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -798,5 +838,45 @@ class TransportIndexMonitorAction @Inject constructor(
...
+            )
+            ExternalSchedulerService.createSchedule(
+                monitor, routing.accountId, routing.queueArn, routing.roleArn, targetInput
```
engechas: minor: we can pass the full routing object instead of breaking it apart here

engechas (changes requested):
1 blocking comment related to guarding the deletion logic behind the external scheduler flag value. The other comments are non-blocking

engechas (approved):

build.gradle
```diff
@@ -22,7 +22,8 @@ buildscript {
             opensearch_build += "-SNAPSHOT"
         }
         opensearch_no_snapshot = opensearch_version.replace("-SNAPSHOT","")
-        common_utils_version = System.getProperty("common_utils.version", opensearch_build)
+        // TODO: revert to opensearch_build after alerting version bump to 3.7.0
+        common_utils_version = System.getProperty("common_utils.version", "3.7.0.0-SNAPSHOT")
```
AWSHurneyt, line 26: @riysaxen-amzn is this change still needed? Looks like common utils has been bumped?
https://github.com/opensearch-project/common-utils/blob/main/build.gradle#L9

This can be fixed in a separate PR.

AWSHurneyt (approved):

build.gradle
riysaxen-amzn: yes for the CI's to pass, i made the change.. will revert it once we have alerting bump

PR #2098, opened 2026-04-17
Bucket level multi-tenant trigger

eirsep:
Replace the custom BucketSelectorExt aggregation with standard
bucket_selector for bucket-level trigger evaluation when
multi_tenant_trigger_eval_enabled is true. This enables trigger
evaluation on user clusters that do not have the alerting plugin
installed.

When the flag is on, bucket-level monitors are limited to 1 trigger.
The standard bucket_selector is injected directly into the monitor's
search query so a single search call performs both data collection and
trigger evaluation. Include/exclude filtering from BucketSelectorExtFilter
is replicated post-response by BucketKeyFilter.

The existing BucketSelectorExt path is completely unchanged when the
flag is off (the default).

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/MonitorMetadataService.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/MonitorMetadataService.kt
@@ -56,7 +56,7 @@ object MonitorMetadataService :
     private lateinit var xContentRegistry: NamedXContentRegistry
     private lateinit var clusterService: ClusterService
     private lateinit var settings: Settings
-    private lateinit var sdkClient: SdkClient
+    lateinit var sdkClient: SdkClient
 
     @Volatile
     private lateinit var indexTimeout: TimeValue
--- a/alerting/src/main/kotlin/org/opensearch/alerting/TriggerService.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/TriggerService.kt
@@ -12,6 +12,7 @@ import org.opensearch.alerting.script.ChainedAlertTriggerExecutionContext
 import org.opensearch.alerting.script.QueryLevelTriggerExecutionContext
 import org.opensearch.alerting.script.TriggerScript
 import org.opensearch.alerting.triggercondition.parsers.TriggerExpressionParser
+import org.opensearch.alerting.util.BucketKeyFilter
 import org.opensearch.alerting.util.CrossClusterMonitorUtils
 import org.opensearch.alerting.util.getBucketKeysHash
 import org.opensearch.cluster.service.ClusterService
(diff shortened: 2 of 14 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/util/BucketKeyFilter.kt
```diff
@@ -0,0 +1,87 @@
...
+        if (includeRegex != null && !includeRegex.matcher(value).matches()) return false
+        if (excludeRegex != null && excludeRegex.matcher(value).matches()) return false
+        return true
+    }
+
+    /**
+     * Extracts include/exclude regex patterns from [IncludeExclude].
+     * Uses reflection to access private fields since no public API exposes the raw patterns.
+     */
+    private fun extractPatterns(includeExclude: IncludeExclude): Pair<Pattern?, Pattern?> {
```
engechas, line 74: Reflection is an anti-pattern and won't be supported in some environments. Is there a way to take a dependency on the IncludeExclude object or recreate it in a way that we can access the fields without reflection?

alerting/src/main/kotlin/org/opensearch/alerting/util/BucketSelectorQueryBuilder.kt
```diff
@@ -0,0 +1,87 @@
...
+            found = null
+            for (agg in aggBuilders) {
+                if (agg.name == element) {
+                    found = agg
+                    aggBuilders = agg.subAggregations
+                    break
+                }
+            }
+            if (found == null) {
+                throw IllegalArgumentException("ParentBucketPath: $parentBucketPath not found in query aggregations")
```
engechas, line 68: Throwing the exception here would only allow one iteration of the for loop right? I'm wondering if that's the intended behavior or if we are looking for the parentAgg in any of the available pathElements

alerting/src/main/kotlin/org/opensearch/alerting/util/BucketSelectorQueryBuilder.kt
```diff
@@ -0,0 +1,87 @@
...
+
+        return found!!
+    }
+
+    /**
+     * Extracts the bucketsPathsMap from [BucketSelectorExtAggregationBuilder] via reflection
+     * since the field is private with no public getter in the commons library.
+     */
+    @Suppress("UNCHECKED_CAST")
+    private fun extractBucketsPathsMap(selector: BucketSelectorExtAggregationBuilder): Map<String, String> {
```
engechas: Same comment on reflection

alerting/src/main/kotlin/org/opensearch/alerting/MonitorMetadataService.kt
```diff
@@ -201,6 +201,15 @@ object MonitorMetadataService :
         }
     }
 
+    private fun isIndexNotFoundException(e: Throwable): Boolean {
```
engechas: Was this from another PR? It looks familiar and doesn't appear to be used anywhere

alerting/src/main/kotlin/org/opensearch/alerting/TriggerService.kt
```diff
@@ -181,6 +182,35 @@ class TriggerService(val scriptService: ScriptService) {
         }
     }
 
+    @Suppress("UNCHECKED_CAST")
```
engechas: We've seen a number of issues related to casting in Alerting. I am not sure if there's a way to avoid unchecked casting here, but if there is we should prefer that

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -418,6 +423,14 @@ class TransportIndexMonitorAction @Inject constructor(
             require(monitor.triggers.size <= maxTriggersPerMonitor) {
                 "The current cluster settings only allow up to $maxTriggersPerMonitor triggers per monitor."
             }
+            if (multiTenantTriggerEvalEnabled &&
+                Monitor.MonitorType.valueOf(monitor.monitorType.uppercase(Locale.ROOT)) ==
+                Monitor.MonitorType.BUCKET_LEVEL_MONITOR
+            ) {
+                require(monitor.triggers.size <= 1) {
```
AWSHurneyt, line 465: Would we need this same validation in the ExecuteMonitor API as well?

eirsep: added in runMonitor code for bucketLeveMonitor

alerting/src/main/kotlin/org/opensearch/alerting/util/BucketKeyFilter.kt
eirsep: changed to parse the json

alerting/src/main/kotlin/org/opensearch/alerting/util/BucketSelectorQueryBuilder.kt
eirsep: There are two nested loops:

- **Outer loop**: iterates over pathElements (e.g., for agg1>agg2, it walks agg1 then agg2)
- **Inner loop**: searches the current level's agg builders for the matching name

The throw is inside the outer loop but outside the inner loop. If agg1 is found, the outer loop continues to look for agg2 inside agg1's sub-aggregations. The exception only fires if a path element isn't found at its expected level — which is correct behavior imo.

alerting/src/main/kotlin/org/opensearch/alerting/MonitorMetadataService.kt
eirsep: removed

alerting/src/main/kotlin/org/opensearch/alerting/TriggerService.kt
eirsep: removed suppress

engechas (approved):

AWSHurneyt (approved):

PR #2099, opened 2026-04-17
Block unsupported actions when multi-tenancy is enabled

engechas:
### Description
Adds checks within workflow, cluster metrics, and doc level monitor actions to reject the request as unsupported when multi-tenancy is enabled

### Related Issues

### Check List
- [x] New functionality has been documented.
- [x] API changes companion pull request [created](https://github.com/opensearch-project/opensearch-api-specification/blob/main/DEVELOPER_GUIDE.md).
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose).

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteWorkflowAction.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteWorkflowAction.kt
@@ -84,11 +84,22 @@ class TransportDeleteWorkflowAction @Inject constructor(
 
     @Volatile override var filterByEnabled = AlertingSettings.FILTER_BY_BACKEND_ROLES.get(settings)
 
+    private val multiTenancyEnabled = AlertingSettings.MULTI_TENANCY_ENABLED.get(settings)
+
     init {
         listenFilterBySettingChange(clusterService)
     }
 
     override fun doExecute(task: Task, request: ActionRequest, actionListener: ActionListener<DeleteWorkflowResponse>) {
+        if (multiTenancyEnabled) {
+            actionListener.onFailure(
+                AlertingException.wrap(
+                    OpenSearchStatusException("Workflow operations are not allowed when multi-tenancy is enabled.", RestStatus.METHOD_NOT_ALLOWED)
+                )
+            )
+            return
+        }
+
         val transformedRequest = request as? DeleteWorkflowRequest
             ?: recreateObject(request) { DeleteWorkflowRequest(it) }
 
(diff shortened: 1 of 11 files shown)


eirsep (commented):
these changes LGTM but there are some more actions that we need to block like
- RestAcknowledgeChainedAlertAction
- RestScheduledJobStatsHandler
- RestGetWorkflowAlertsAction
- RestGetFindingsAction
- RestGetRemoteIndexesAction


you need to check with @MohammedAghil if  the following are needed or not(IMO they aren't)
- RestSearchEmailAccountAction
- RestGetEmailAccountAction
- RestSearchEmailGroupAction
- RestGetEmailGroupAction
- RestGetDestinationsAction

eirsep (approved):

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportDeleteWorkflowAction.kt
```diff
@@ -84,11 +84,25 @@ class TransportDeleteWorkflowAction @Inject constructor(
 
     @Volatile override var filterByEnabled = AlertingSettings.FILTER_BY_BACKEND_ROLES.get(settings)
 
+    private val multiTenancyEnabled = AlertingSettings.MULTI_TENANCY_ENABLED.get(settings)
```
riysaxen-amzn, line 87: is the setting is meant to be toggled at runtime via cluster settings update? if yes, then we need to do @Volatile + clusterService.clusterSettings.addSettingsUpdateConsumer(...) like `filterByEnabled` does?

engechas: It's marked as final (not able to be changed after startup) -> https://github.com/opensearch-project/alerting/blob/main/alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt#L324

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportExecuteMonitorAction.kt
```diff
@@ -133,6 +136,17 @@ class TransportExecuteMonitorAction @Inject constructor(
                                 getResponse.sourceAsBytesRef, XContentType.JSON
                             ).use { xcp ->
                                 val monitor = ScheduledJob.parse(xcp, getResponse.id, getResponse.version) as Monitor
+                                if (multiTenancyEnabled && monitor.isUnsupportedMultiTenantMonitorType()) {
```
riysaxen-amzn, line 139: The monitor type check appears in two places — once for the "fetch by ID" path and once for the "inline monitor" path (line 173).Should we consider extracting to a private helper method?

engechas: The monitor type check is already abstracted. I am against clubbing the two booleans into a single check because:
* The verbosity makes it clear that the condition is based on two independent factors
* The check is extensible if the following conditions need consideration:
** Multi tenancy enabled, but monitor type is supported

riysaxen-amzn (approved):

PR #2107, opened 2026-04-28
 Make MonitorJobPoller populate thread context for downstream request interception

toepkerd:
### Description
Job poller now populates the thread context with information about the external data source so downstream flows can properly intercept it and make Search and PPL calls to external data source.

### Check List

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/service/MonitorJobPoller.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/service/MonitorJobPoller.kt
@@ -5,6 +5,8 @@
 
 package org.opensearch.alerting.service
 
+import java.time.Instant
+import java.util.concurrent.atomic.AtomicInteger
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.SupervisorJob
@@ -30,8 +32,6 @@ import software.amazon.awssdk.services.sqs.SqsClient
 import software.amazon.awssdk.services.sqs.model.DeleteMessageRequest
 import software.amazon.awssdk.services.sqs.model.Message
 import software.amazon.awssdk.services.sqs.model.ReceiveMessageRequest
-import java.time.Instant
-import java.util.concurrent.atomic.AtomicInteger
 
 /**
  * Polls SQS queues for monitor execution messages and dispatches them
@@ -65,6 +65,7 @@ class MonitorJobPoller(
         }
         val provider = requireNotNull(accountIdProvider) { "accountIdProvider must be set before starting" }
         val sqs = requireNotNull(sqsClient) { "sqsClient must be set before starting" }
+        require(region.isNotBlank()) { "region must be set before starting" }
 
         logger.info("Starting MonitorJobPoller with $POLLER_THREAD_COUNT workers")
         repeat(POLLER_THREAD_COUNT) { scope.launch { pollLoop(provider, sqs, region, queueName) } }
@@ -194,28 +195,41 @@ class MonitorJobPoller(
             )
         }
 
-        if (region.isBlank()) {
+        if (monitor.target!!.type.isBlank()) {
             throw AlertingException.wrap(
-                IllegalStateException("No region configured when populating thread context from job poller")
+                IllegalStateException("Monitor target received by Job Poller did not contain target type")
+            )
+        }
+
+        if (monitor.target!!.endpoint.isBlank()) {
+            throw AlertingException.wrap(
+                IllegalStateException("Monitor target received by Job Poller did not contain endpoint")
             )
         }
 
         val threadContext = client.threadPool().threadContext
 
-        // Request interception checks for this flag to know that because this is
-        // a scheduled background monitor execution, there will be
+        // Request interception checks for this flag to know that this is
+        // a scheduled background monitor execution, meaning there will be
         // no user credentials to make the search/ppl call to customer
         // data source with, and it must use service credentials
         threadContext.putHeader(IS_BACKGROUND_JOB_HEADER, "true")
 
-        // TODO: in long term, may need to generalize to aos data source type
-        threadContext.putHeader(SERVICE_NAME_HEADER, "aoss")
+        threadContext.putHeader(SERVICE_NAME_HEADER, mapTargetTypeToServiceName(monitor.target!!.type))
 
         // external customer data source endpoint, to run search/ppl against
         threadContext.putHeader(OPENSEARCH_ENDPOINT_HEADER, monitor.target!!.endpoint)
 
         // populated upstream in AlertingPlugin.kt with REMOTE_METADATA_REGION.get(settings)
         threadContext.putHeader(REGION_HEADER, region)
+    }
+
+    private fun mapTargetTypeToServiceName(targetType: String): String {
+        return when (targetType) {
+            "AOSS_COLLECTION" -> "aoss"
+            "AOS_DOMAIN" -> "es"
+            else -> throw AlertingException.wrap(IllegalStateException("Received unknown target type in Job Poller: " + targetType))
+        }
     }
 
     companion object {
--- a/alerting/src/main/kotlin/org/opensearch/alerting/service/MonitorJobPoller.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/service/MonitorJobPoller.kt
@@ -134,7 +134,7 @@ class MonitorJobPoller(
     }
 
     private suspend fun executeMonitor(monitor: Monitor, jobStartTime: Instant) {
-        // populate thread context for downstream Oasis interception the moment
+        // populate thread context for downstream request interception the moment
         // Monitor config is in hand
         populateThreadContext(monitor)
 
@@ -184,8 +184,8 @@ class MonitorJobPoller(
         }
     }
 
-    // populates thread context with KVs that Oasis will need
-    // when intercepting search or PPL calls to external customer
+    // populates thread context with KVs that downstream interception will
+    // need when intercepting search or PPL calls to external customer
     // data source
     internal fun populateThreadContext(monitor: Monitor) {
         if (monitor.target == null) {
@@ -202,7 +202,7 @@ class MonitorJobPoller(
 
         val threadContext = client.threadPool().threadContext
 
-        // Oasis checks for this flag to know that because this is
+        // Request interception checks for this flag to know that because this is
         // a scheduled background monitor execution, there will be
         // no user credentials to make the search/ppl call to customer
         // data source with, and it must use service credentials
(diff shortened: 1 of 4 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/service/MonitorJobPoller.kt
```diff
@@ -134,6 +134,10 @@ class MonitorJobPoller(
     }
 
     private suspend fun executeMonitor(monitor: Monitor, jobStartTime: Instant) {
+        // populate thread context for downstream Oasis interception the moment
+        // Monitor config is in hand
+        populateThreadContext(monitor)
```
riysaxen-amzn: populateThreadContext() calls threadContext.putHeader() but never cleans up. If the poller processes multiple monitors in the same thread (it uses a thread pool of 10), headers from monitor A could leak into monitor B's execution if B fails the null-target check and throws before setting its own headers.

The standard OpenSearch pattern is to use threadContext.stashContext() which returns a StoredContext that restores the original headers when closed

```
client.threadPool().threadContext.stashContext().use {
    populateThreadContext(monitor)
    // ... rest of execution
}
```

This ensures each monitor execution starts with a clean thread context.

alerting/src/main/kotlin/org/opensearch/alerting/service/MonitorJobPoller.kt
```diff
@@ -180,8 +184,48 @@ class MonitorJobPoller(
...
+
+        // populated upstream in AlertingPlugin.kt with REMOTE_METADATA_REGION.get(settings)
+        threadContext.putHeader(REGION_HEADER, region)
+    }
+
     companion object {
         const val POLLER_THREAD_COUNT = 10
         const val POLL_INTERVAL_MS = 1000L
+
+        // thread context header keys for request interception
```
riysaxen-amzn, line 246: nit: I think we should follow a consistent naming convention for all these headers but needs to be discussed this across team

alerting/src/main/kotlin/org/opensearch/alerting/service/MonitorJobPoller.kt
```diff
@@ -180,8 +184,48 @@ class MonitorJobPoller(
...
+        val threadContext = client.threadPool().threadContext
+
+        // Request interception checks for this flag to know that because this is
+        // a scheduled background monitor execution, there will be
+        // no user credentials to make the search/ppl call to customer
+        // data source with, and it must use service credentials
+        threadContext.putHeader(IS_BACKGROUND_JOB_HEADER, "true")
+
+        // TODO: in long term, may need to generalize to aos data source type
+        threadContext.putHeader(SERVICE_NAME_HEADER, "aoss")
```
engechas: Is this information in the monitor config already? If so we can pull it from there instead of hardcoding

alerting/src/main/kotlin/org/opensearch/alerting/service/MonitorJobPoller.kt
toepkerd: The job poller executes a monitor with `client.execute()` which calls the TransportExecuteMonitorAction. This class already stashes context before actually running the Monitor ([reference](https://github.com/opensearch-project/alerting/blob/aaf7338da1f3a54cc8cdf76b4db19c261facb379/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportExecuteMonitorAction.kt#L76)), ensuring the desired clean slate.

I should note that these downstream thread context stashes should not interfere with logic even further downstream to intercept Search and PPL calls to external data source.

alerting/src/main/kotlin/org/opensearch/alerting/service/MonitorJobPoller.kt
toepkerd: These constants are actually existing header names from other code owners that we are reusing. The only new and original const here is `IS_BACKGROUND_JOB_HEADER = "alerting-is-background-job"`

alerting/src/main/kotlin/org/opensearch/alerting/service/MonitorJobPoller.kt
toepkerd: I don't think it's anywhere in the Monitor config, but there is a `REMOTE_METADATA_SERVICE_NAME` setting I could retrieve the value from. Will explore reading from this value instead of hard-coding `aoss`.

toepkerd: If reading from `REMOTE_METADATA_SERVICE_NAME` approach is used, we must be mindful of conventions for naming services. For collection data sources, I think we can trust the service name will always be `aoss`.

In the long term, if we support domain data sources, we will need to match existing service name conventions in the interception logic. Whether they favor `aos` or `es` as the service name, our logic for populating these remote metadata settings (outside the scope of this PR) will need to be ready to match that convention if reading from settings is to work.

alerting/src/main/kotlin/org/opensearch/alerting/service/MonitorJobPoller.kt
```diff
@@ -180,8 +184,48 @@ class MonitorJobPoller(
...
+    // need when intercepting search or PPL calls to external customer
+    // data source
+    internal fun populateThreadContext(monitor: Monitor) {
+        if (monitor.target == null) {
+            throw AlertingException.wrap(
+                IllegalStateException("Monitor received by Job Poller did not contain target")
+            )
+        }
+
+        if (region.isBlank()) {
```
eirsep: this should be a check before poller threads are bootstrapped itself near
        val provider = requireNotNull(accountIdProvider) { "accountIdProvider must be set before starting" }
        val sqs = requireNotNull(sqsClient) { "sqsClient must be set before starting" }

alerting/src/main/kotlin/org/opensearch/alerting/service/MonitorJobPoller.kt
eirsep: [`monitor.target.type` ](https://github.com/opensearch-project/common-utils/blob/main/src/main/kotlin/org/opensearch/commons/alerting/model/Monitor.kt#L49) will have this header info 

https://github.com/opensearch-project/common-utils/blob/main/src/main/kotlin/org/opensearch/commons/alerting/model/Target.kt

eirsep: @riysaxen-amzn @MohammedAghil can help ensure that this header's value will match exactly what the header needs or what a setting can hold as a kv map where key will be `target.type` and value would be header value

eirsep (changes requested):

alerting/src/main/kotlin/org/opensearch/alerting/service/MonitorJobPoller.kt
```diff
@@ -180,8 +184,48 @@ class MonitorJobPoller(
...
+        // a scheduled background monitor execution, there will be
+        // no user credentials to make the search/ppl call to customer
+        // data source with, and it must use service credentials
+        threadContext.putHeader(IS_BACKGROUND_JOB_HEADER, "true")
+
+        // TODO: in long term, may need to generalize to aos data source type
+        threadContext.putHeader(SERVICE_NAME_HEADER, "aoss")
+
+        // external customer data source endpoint, to run search/ppl against
+        threadContext.putHeader(OPENSEARCH_ENDPOINT_HEADER, monitor.target!!.endpoint)
```
eirsep: plz check that it's not null/empty before you set it.

alerting/src/main/kotlin/org/opensearch/alerting/service/MonitorJobPoller.kt
toepkerd: I see, I misunderstood the monitor.target.type field, thanks for clarifying. I will double check the values that this field can take on (other than the default value of "local"). I will add fail-fast validations for the type.

alerting/src/main/kotlin/org/opensearch/alerting/service/MonitorJobPoller.kt
```diff
@@ -180,8 +186,76 @@ class MonitorJobPoller(
...
+        threadContext.putHeader(SERVICE_NAME_HEADER, mapTargetTypeToServiceName(target.type))
+
+        // external customer data source endpoint, to run search/ppl against
+        threadContext.putHeader(OPENSEARCH_ENDPOINT_HEADER, target.endpoint)
+
+        // populated upstream in AlertingPlugin.kt with REMOTE_METADATA_REGION.get(settings)
+        threadContext.putHeader(REGION_HEADER, region)
+    }
+
+    private fun mapTargetTypeToServiceName(targetType: String): String {
```
eirsep, line 229: we shouldn't hardcode.
plz move this to a setting holding a map

alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
```diff
@@ -404,5 +404,10 @@ class AlertingSettings {
             "plugins.alerting.external_scheduler.job_queue_message_group_key_name",
             Setting.Property.NodeScope, Setting.Property.Dynamic
         )
+
+        val JOB_QUEUE_TARGET_TYPE_TO_SERVICE_NAME = Setting.groupSetting(
```
eirsep: will this default to empty map

alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
```diff
@@ -404,5 +404,10 @@ class AlertingSettings {
...
+            "plugins.alerting.external_scheduler.type_to_service.",
```
eirsep: can this be renamed to `plugins.alerting.monitor.target_type_to_service_name.`

eirsep (approved):
minor comments but rest LGTM

alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
toepkerd: Yes.

If no settings with this prefix is configured anywhere, MonitorJobPoller construction in AlertingPlugin.kt effectively gets an empty keySet() (like `get(SETTINGS.Empty)`). This means every Monitor Job Poller job will fail at thread context writing with "invalid target type" exception. This is intentional behavior: values for these settings must be configured in the following way in `opensearch.yml`:

```
plugins.alerting.monitor.target_type_to_service_name.target_type_1: service_name_1
plugins.alerting.monitor.target_type_to_service_name.target_type_2: service_name_2
```

alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
toepkerd: Done

riysaxen-amzn (approved):

PR #2128, opened 2026-05-07
PPL Alerting: Create, Read, Update, Delete

toepkerd:
### Description
PPL Alerting Index Monitor changes and other CRUD

### Check List

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/service/DeleteMonitorService.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/service/DeleteMonitorService.kt
@@ -184,7 +184,11 @@ object DeleteMonitorService :
     }
 
     private suspend fun deleteLock(monitor: Monitor) {
-        client.suspendUntil<Client, Boolean> { lockService.deleteLock(LockModel.generateLockId(monitor.id), it) }
+        deleteLock(monitor.id)
+    }
+
+    private suspend fun deleteLock(monitorId: String) {
+        client.suspendUntil<Client, Boolean> { lockService.deleteLock(LockModel.generateLockId(monitorId), it) }
     }
 
     /**
--- a/alerting/src/main/kotlin/org/opensearch/alerting/service/DeleteMonitorService.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/service/DeleteMonitorService.kt
@@ -184,11 +184,7 @@ object DeleteMonitorService :
     }
 
     private suspend fun deleteLock(monitor: Monitor) {
-        deleteLock(monitor.id)
-    }
-
-    private suspend fun deleteLock(monitorId: String) {
-        client.suspendUntil<Client, Boolean> { lockService.deleteLock(LockModel.generateLockId(monitorId), it) }
+        client.suspendUntil<Client, Boolean> { lockService.deleteLock(LockModel.generateLockId(monitor.id), it) }
     }
 
     /**
--- a/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
@@ -102,6 +102,7 @@ import org.opensearch.remote.metadata.client.PutDataObjectRequest
 import org.opensearch.remote.metadata.client.SdkClient
 import org.opensearch.rest.RestRequest
 import org.opensearch.search.builder.SearchSourceBuilder
+import org.opensearch.sql.plugin.transport.TransportPPLQueryResponse
 import org.opensearch.tasks.Task
 import org.opensearch.transport.TransportService
 import org.opensearch.transport.client.Client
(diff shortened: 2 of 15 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -292,6 +327,200 @@ class TransportIndexMonitorAction @Inject constructor(
...
+
+            override fun onFailure(e: Exception) {
+                actionListener.onFailure(e)
+            }
+        }
+
+        // initiate the PPL monitor and PPL query validations
+        client.threadPool().threadContext.stashContext().use {
+            scope.launch {
+                val singleThreadContext = newSingleThreadContext("IndexPPLMonitorActionThread")
```
eirsep: why are we required to do this?

eirsep: newSingleThreadContext("IndexPPLMonitorActionThread") creates a new ExecutorCoroutineDispatcher (backed by a single-thread ExecutorService) on every invocation of checkPPLQueryAndExecute. This dispatcher is 
never .close()'d, so the underlying thread and executor are leaked. Under load, each monitor create/update permanently leaks one thread.

Either:
- Call singleThreadContext.close() in a finally block after withContext completes (I don't prefer this)
- Or better, replace with a shared dispatcher (e.g. Dispatchers.IO or a class-level newSingleThreadContext created only one and not per-request) since the per-request isolation isn't necessary here — the restored ThreadContext is only 
needed for the duration of the PPL query call and can be cleaned up afterward.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -292,6 +327,200 @@ class TransportIndexMonitorAction @Inject constructor(
...
+            if (pplTrigger.conditionType == PPLTrigger.ConditionType.NUMBER_OF_RESULTS &&
+                pplTrigger.numResultsValue!! > maxQueryResults
+            ) {
+                validationListener.onFailure(
+                    AlertingException.wrap(
+                        IllegalArgumentException(
+                            "Trigger ${trigger.id} checks for number of results threshold of ${trigger.numResultsValue}, " +
+                                "but PPL Alerting is configured only to retrieve $maxQueryResults query results maximum. " +
+                                "Please lower the number of results value to one below this maximum value, or adjust the cluster " +
+                                "setting: $PPL_QUERY_RESULTS_MAX_DATAROWS.key}"
```
eirsep: missing `{`

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -101,7 +116,7 @@ private val log = LogManager.getLogger(TransportIndexMonitorAction::class.java)
 private val scope: CoroutineScope = CoroutineScope(Dispatchers.IO)
 
 class TransportIndexMonitorAction @Inject constructor(
-    transportService: TransportService,
+    val transportService: TransportService,
```
eirsep: why did we do this?

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -292,6 +327,200 @@ class TransportIndexMonitorAction @Inject constructor(
...
+            scope.launch {
+                val singleThreadContext = newSingleThreadContext("IndexPPLMonitorActionThread")
+                withContext(singleThreadContext) {
+                    it.restore()
+
+                    val pplMonitor = indexMonitorRequest.monitor
+
+                    // validate the PPL query syntax and that user has permissions to
+                    // the indices being queried
+                    val pplQueryValid = validatePPLQuery(pplMonitor, validationListener)
```
eirsep: validatePPLMonitor only does in-memory checks (query length, trigger thresholds, notification template lengths). It should run before validatePPLQuery, which actually executes the query against the PPL engine.

eirsep (changes requested):

alerting/src/main/kotlin/org/opensearch/alerting/service/DeleteMonitorService.kt
```diff
@@ -184,7 +184,11 @@ object DeleteMonitorService :
     }
 
     private suspend fun deleteLock(monitor: Monitor) {
```
eirsep, line 186: we can revert this??

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
toepkerd: Will modify to get a pooled thread from Dispatchers.IO

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
toepkerd: This is a vestige of using the TransportService directly to make transport PPL call, but now client.execute(PPLQueryAction) is used. I will revert this.

alerting/src/main/kotlin/org/opensearch/alerting/util/AlertingUtils.kt
```diff
@@ -293,3 +306,102 @@ fun printsSampleDocData(trigger: Trigger): Boolean {
...
+    monitorCtx: MonitorRunnerExecutionContext
+): NotificationActionConfigs {
+    var destination: Destination? = null
+    var notificationPermissionException: Exception? = null
+
+    var channel: NotificationConfigInfo? = null
+    try {
+        channel = getNotificationConfigInfo(monitorCtx.client as NodeClient, action.destinationId)
+    } catch (e: OpenSearchSecurityException) {
+        notificationPermissionException = e
```
engechas, line 366: Should we return from here? Not sure if it makes sense to continue if we can't resolve the channel

engechas (approved):

alerting/src/main/kotlin/org/opensearch/alerting/util/AlertingUtils.kt
toepkerd: I believe we should not return early and keep this logic as is. Downstream logic checks for the case of the user using legacy Notifications and tries to resolve the channel from there. If we return here, it would be a breaking change for customers who have not migrated from legacy to the Notifications of today.

eirsep (approved):

eirsep (approved):

PR #2130, opened 2026-05-08
PPL Alerting: PPL Monitor Execution

toepkerd:
### Description
This PR primarily contains logic for running PPL Monitors. It also adds RBAC checks for manual Monitor executions, a gap revealed by a security review.

To support RBAC filtering for manual monitor execution, a new field `manual` is added to ExecuteMonitorRequest/Response. *Many of the file changes are simply to accommodate this new field and can be ignored in review.*

### The only files that need deep review are:
`TriggerService.kt`
`InputService.kt`
`AlertService.kt`
`QueryLevelMonitorRunner.kt`
`TransportExecuteMonitorAction.kt`
`QueryLevelTriggerExecutionContext.kt`

### New `manual` Field
The manual field in ExecuteMonitorRequest distinguishes between user-initiated executions (via the REST API) and system-initiated executions (via the job scheduler or MonitorJobPoller). When a user calls the Execute Monitor API, the REST handler sets manual = true. When the scheduler or poller triggers execution internally, it sets manual = false. In TransportExecuteMonitorAction, this field gates the RBAC check: when manual is true and the monitor was fetched by ID, checkUserPermissionsWithResource is called to verify the calling user's backend roles overlap with the monitor owner's backend roles (enforcing filter_by_backend_roles). When manual is false, this check is skipped entirely — scheduled executions run as the system and must not be blocked by user-level RBAC. This ensures that monitors continue to fire on schedule regardless of which user created them, while still preventing unauthorized users from manually triggering another user's monitor via the API.

### Check List


alerting/src/main/kotlin/org/opensearch/alerting/script/QueryLevelTriggerExecutionContext.kt
```diff
@@ -9,31 +9,43 @@ import org.opensearch.alerting.model.AlertContext
...
 import org.opensearch.commons.alerting.model.QueryLevelTriggerRunResult
+import org.opensearch.commons.alerting.model.Trigger
 import java.time.Instant
 
 data class QueryLevelTriggerExecutionContext(
     override val monitor: Monitor,
-    val trigger: QueryLevelTrigger,
+    val trigger: Trigger,
     override val results: List<Map<String, Any>>,
+    val pplQueryResults: List<Map<String, Any?>>, // each list element is a result row
```
lezzago, line 51: Can you explain the difference between the two and how the value differs for ppl monitors?

alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
```diff
@@ -309,8 +309,8 @@ class AlertingSettings {
             Setting.Property.NodeScope, Setting.Property.Dynamic
         )
 
-        val PPL_MONITOR_EXECUTION_MAX_DURATION = Setting.positiveTimeSetting(
```
lezzago, line 312: Why are we removing this?

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportExecuteMonitorAction.kt
```diff
@@ -159,6 +154,37 @@ class TransportExecuteMonitorAction @Inject constructor(
...
+                                )
+                                return@whenComplete
+                            }
+
+                            // in the case that the Monitor is already indexed, do an RBAC
+                            // check if this Monitor is being executed manually via API
+                            if (execMonitorRequest.manual && !checkUserPermissionsWithResource(
+                                    user, monitor.user, actionListener,
+                                    "monitor", execMonitorRequest.monitorId
+                                )
```
lezzago: Instead should we call the get monitor api with the user's context, so it would automatically validate correctly if the user has permissions to even fetch the monitor as the get monitor api would do the RBAC check.

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
```diff
@@ -202,6 +205,10 @@ class AlertService(
                 }
             }
 
+        // populate PPL Monitor specific fields
+        val query = if (ctx.monitor.isPPLMonitor()) (ctx.monitor.inputs[0] as PPLInput).query else null
+        val queryResults = if (ctx.monitor.isPPLMonitor()) ctx.pplQueryResults else emptyList()
```
lezzago, line 213: Rename this to `pplQuery` and `pplQueryResults` if this is only meant for ppl monitors.

alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
toepkerd: it's replaced with a per-trigger max execution duration setting as opposed to a per-monitor max execution duration. Each custom trigger can run its own query, so the timeboxing is done at a per trigger level instead of a per monitor level.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportExecuteMonitorAction.kt
toepkerd: I'm not sure about this approach. Get Monitor calls `checkUserPermissionsWithResource` just like ExecuteMonitor is doing here ([reference](https://github.com/opensearch-project/alerting/blob/93fcb58052aa40264fa3751c5b72c8508eddefa3/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportGetMonitorAction.kt#L134)), adding an extra hop from ExecuteMonitor to GetMonitor to call `checkUserPermissionsWithResource` doesn't seem very clean when ExecuteMonitor can do that itself.

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
toepkerd: This naming is used so that the query and query results in Alerts specifically can be included for more than just PPL Monitors, in response to this [feedback from common-utils](https://github.com/opensearch-project/common-utils/pull/940#discussion_r3102325771). This generalization (of creating a field that either PPL Monitors or existing Monitor types can use) only exists for Alerts currently. In internal Monitor execution logic, a distinction is made between DSL results and PPL results because their shapes are different despite both being List<Map<String, Any>>. See: https://github.com/opensearch-project/alerting/pull/2130#discussion_r3220438584.

alerting/src/main/kotlin/org/opensearch/alerting/script/QueryLevelTriggerExecutionContext.kt
toepkerd: Despite both being List<Map<String, Any>>, the shape/structure of the data inside is different between the two, so the code can't assume it can parse/read that field the same way in either a PPL or DSL Monitor run.

Example `results` value (used by Query-Level Monitors and other DSL-based Monitors):
```
[
    {
      "_shards": {"total": 1, "failed": 0, "successful": 1, "skipped": 0},
      "hits": {
        "hits": [{"_id": "abc", "_source": {"status": 500, "endpoint": "/api"}}],
        "total": {"value": 3, "relation": "eq"},
        "max_score": 1.0
      },
      "took": 45,
      "timed_out": false
    }
  ]
```

Example `pplQueryResults` value (used only by PPL Monitors):
```
[
    {"endpoint": "/api/orders", "error_count": 3},
    {"endpoint": "/api/checkout", "error_count": 1}
]
```

alerting/src/main/kotlin/org/opensearch/alerting/InputService.kt
```diff
@@ -220,6 +228,84 @@ class InputService(
...
+
+        // TODO: change name to trigger max duration
+        val monitorExecutionDuration = monitorCtx
+            .clusterService!!
+            .clusterSettings
+            .get(AlertingSettings.PPL_TRIGGER_EXECUTION_MAX_DURATION)
+
+        var queryResponseJson: JsonNode? = null
+
+        withTimeout(monitorExecutionDuration.millis) {
```
lezzago: if it times out, do we want any logging and catching the exception?

toepkerd: That is already done. QueryLevelMonitorRunner calls `collectInputResultsForPPLMonitor()`, which wraps `runPPLBaseQuery()` in a try catch. `collectInputResultsForPPLMonitor()`'s try catch catches the timeout exception (or any other exception), logs an error log, then returns an `InputRunResults` with a non-empty error. This causes the generation of an ERROR alert.

alerting/src/main/kotlin/org/opensearch/alerting/script/QueryLevelTriggerExecutionContext.kt
toepkerd: Will add comment explaining the difference between these fields.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportExecuteMonitorAction.kt
toepkerd: Exploring running checkUserPermissionsWithResource() directly, without checking manual field first.

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
toepkerd: Need to clarify comment that these fields are for PPL Monitors for now, but can be extended for other Monitor types to use.

alerting/src/main/kotlin/org/opensearch/alerting/AlertingPlugin.kt
```diff
@@ -403,6 +402,8 @@ internal class AlertingPlugin : PainlessExtension, ActionPlugin, ScriptPlugin, R
             }
         }
 
+        ExternalSchedulerService.initialize(settings)
```
eirsep: why are these changes appearing as an addition from this CR?

toepkerd: This must have been an oversight on my part when fixing merge conflicts. `ExternalSchedulerService.initialize(settings)` is already called on line 393 (and it's not showing up as an addition in the PR). Will remove this.

alerting/src/main/kotlin/org/opensearch/alerting/InputService.kt
```diff
@@ -220,6 +228,84 @@ class InputService(
...
+    suspend fun collectInputResultsForPPLMonitor(
+        monitor: Monitor,
+        monitorCtx: MonitorRunnerExecutionContext
+    ): InputRunResults {
+        return try {
+            // PPL Alerting:
+            // these query results are for number_of_results PPL triggers,
+            // only the number of results returned by base query matters
+            // for those triggers, not the contents themselves
+            val basePplQueryResults = runPPLBaseQuery(
```
eirsep, line 251: if there are no numResults triggers skip runnin base query

alerting/src/main/kotlin/org/opensearch/alerting/InputService.kt
```diff
@@ -220,6 +228,84 @@ class InputService(
         }
     }
 
+    suspend fun collectInputResultsForPPLMonitor(
```
eirsep, line 232: i would bat for PPL monitor to have only 1 custom PPL trigger and if it has any custom trigger then it shouldn't have a num results trigger.. but it can have multiple num_results triggers in the same monitor.

running one monitor sequentially would be a nightmare because partial failures would be tough to ahndle and if the job crahes after processing 2 out 5 custom triggers, then when it retries there would be duplicate

just an open discussion..

eirsep (changes requested):

alerting/src/main/kotlin/org/opensearch/alerting/InputService.kt
eirsep: What is the timeout value.. also what is max number of PPL triggers allowed?
is <timeout> x <max triggers> less than 5 minutes. 

Am asking because the  monitor cannot run longer than 5 minutes in total since the job scheduler lock expires and will fire a duplicate job

alerting/src/main/kotlin/org/opensearch/alerting/action/ExecuteMonitorRequest.kt
```diff
@@ -15,19 +15,22 @@ import java.io.IOException
 
 class ExecuteMonitorRequest : ActionRequest {
     val dryrun: Boolean
+    val manual: Boolean
```
eirsep: check if the value set by bg job in https://github.com/opensearch-project/alerting/pull/2132/changes is skipping RBAC requirement correctly.

alerting/src/main/kotlin/org/opensearch/alerting/InputService.kt
toepkerd: Alerts are collected while PPL Triggers are run (just like Query-Level Triggers), then bulk saved all at once after all Triggers are executed. If the job crashes in the middle of processing its Triggers, no Alerts would have been written. This means no duplicate Alerts when the PPL Monitor retries.

alerting/src/main/kotlin/org/opensearch/alerting/TriggerService.kt
```diff
@@ -271,4 +283,156 @@ class TriggerService(val scriptService: ScriptService) {
...
+                            limitedQueryToExecute,
+                            false,
+                            monitorCtx.client!! as NodeClient
+                        )
+                    }
+                }
+                logger.debug("query results for trigger ${pplTrigger.id}: $queryResponseJson")
+                logger.debug("time taken to execute query against sql/ppl plugin: $timeTaken")
+
+                // val numPplResults = basePplQueryResults.getLong("total")
```
lezzago: Did you mean to delete the commented code?

alerting/src/main/kotlin/org/opensearch/alerting/InputService.kt
toepkerd: Will reduce max ppl query duration setting default value from 1m to 30s, and add a new max ppl triggers setting whose max value is 10.

alerting/src/main/kotlin/org/opensearch/alerting/workflow/WorkflowRunner.kt
```diff
@@ -93,107 +78,6 @@ abstract class WorkflowRunner {
         }
     }
 
-    protected suspend fun getConfigAndSendNotification(
```
eirsep, line 96: why did we remove this? why are we re-using workflow's send notification instead of the one from monitor runner

alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
```diff
@@ -309,9 +309,17 @@ class AlertingSettings {
...
+            10,
+            0,
+            10,
+            Setting.Property.NodeScope, Setting.Property.Dynamic
+        )
+
         val PPL_QUERY_EXECUTION_MAX_DURATION = Setting.positiveTimeSetting(
             "plugins.alerting.ppl_query_max_execution_duration",
-            TimeValue(1, TimeUnit.MINUTES),
+            TimeValue(30, TimeUnit.SECONDS),
```
lezzago, line 322: Note that in the documentation page, we should inform users that if they increase this with 10 triggers, and have the monitor run every 5 minutes, there can be a duplicate monitor that gets executed while this monitor is trying to be completed.

lezzago (approved):

alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
```diff
@@ -309,9 +309,17 @@ class AlertingSettings {
             Setting.Property.NodeScope, Setting.Property.Dynamic
         )
 
-        val PPL_MONITOR_EXECUTION_MAX_DURATION = Setting.positiveTimeSetting(
-            "plugins.alerting.ppl_monitor_max_execution_duration",
-            TimeValue(1, TimeUnit.MINUTES),
+        val MAX_PPL_TRIGGERS_PER_MONITOR = Setting.intSetting(
```
AWSHurneyt, line 312: @toepkerd for my own understanding, could you clarify why we need separate settings for PPL monitors as opposed to just using the pre-existing `MAX_TRIGGERS_PER_MONITOR` setting?

eirsep (approved):

toepkerd:
On the side, here are some CIs that ran on my fork, given that CIs are blocked on the official repo: https://github.com/toepkerd/alerting/actions/runs/25764960009

alerting/src/main/kotlin/org/opensearch/alerting/InputService.kt
```diff
@@ -220,6 +229,110 @@ class InputService(
...
+
+            logger.debug("executing the base PPL query of monitor: ${pplMonitor.id}")
+            val (queryResponseJsonReceived, timeTaken) = measureTimedValue {
+                executePplQuery(
+                    limitedQueryToExecute,
+                    false,
+                    monitorCtx.client!! as NodeClient
+                )
+            }
+            logger.debug("base query results: $queryResponseJsonReceived")
```
AWSHurneyt, line 327: **Not blocking:**
If we're planning to log the query results, it makes sense to hide those at the debug level. However, are there any generic details you can think of that would be helpful to log at the info level to help troubleshoot issues? For example, rather than logging the query results at the info level, would it be helpful to log the number of results at the info level?

You don't have to revise the PR to add more logs right now. That can always be done as a follow-up item. This is more just food for thought.

alerting/src/main/kotlin/org/opensearch/alerting/InputService.kt
```diff
@@ -220,6 +229,110 @@ class InputService(
...
+            logger.debug("time taken to execute base query against sql/ppl plugin: $timeTaken")
```
AWSHurneyt, line 328: **Not blocking:**
I don't think we have a mechanism for this right now, but this could be a helpful metric to capture somewhere.

It would potentially give us the option of warning a user that their PPL with 5 triggers is currently taking 4 minutes and 50 seconds to complete on average.

Again, just food for thought.

alerting/src/main/kotlin/org/opensearch/alerting/TriggerService.kt
```diff
@@ -271,4 +283,154 @@ class TriggerService(val scriptService: ScriptService) {
...
+
+        if (pplTrigger.customCondition == null) {
+            return QueryLevelTriggerRunResult(
+                pplTrigger.name,
+                true,
+                IllegalStateException("No custom condition found for trigger: ${pplTrigger.id}")
+            )
+        }
+
+        // TODO: change name to trigger max duration
```
AWSHurneyt: Are these TODO tasks being tracked somewhere?

AWSHurneyt (approved):

alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt
toepkerd: The existing MAX_TRIGGERS allows for up to 50 triggers. In PPL Alerting, custom Triggers run their own PPL query. Each query is timeboxed to 30 seconds. In the worst case, 50 PPL Triggers running 30 second queries is a Monitor execution time of 25 minutes. This exceeds the 5 minute lifetime of a scheduled job lock. With the lock released before the Monitor is done executing, duplicate jobs can be spawned. The new setting is introduced to enforce PPL Monitors having at most 10 PPL Triggers.

alerting/src/main/kotlin/org/opensearch/alerting/InputService.kt
toepkerd: This seems like quite a big feature request. My biggest question is where we would display that warning, an ERROR Alert? As a UI popup? If a UI popup, how will the UI read in information about the warning?

alerting/src/main/kotlin/org/opensearch/alerting/TriggerService.kt
toepkerd: This is a stale TODO actually, it has already been completed

PR #2131, opened 2026-05-08
feat: Add tenant_id to monitor metadata and propagate to AlertService ADK calls via background and APIs

eirsep:
- Store tenant_id in monitor.metadata map during create flow in TransportIndexMonitorAction when multi-tenancy is enabled
- Set x-tenant-id header in MonitorJobPoller.populateThreadContext() from monitor metadata before executing monitors from SQS
- Add tenantId(currentTenantId()) to all AlertService SDK client calls (search, put, delete) for proper multi-tenant routing
- Add debug logging in AlertService for tenant context visibility
- Add unit tests for metadata key logic


toepkerd (approved):

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
```diff
@@ -701,6 +709,7 @@ class AlertService(
                 Alert.State.ACTIVE, Alert.State.ERROR -> {
                     putRequests.add(
                         PutDataObjectRequest.builder()
+                            .tenantId(currentTenantId())
```
engechas: Tenant ID is already present in these

engechas (approved):

alerting/src/main/kotlin/org/opensearch/alerting/AlertService.kt
eirsep: removed

PR #2141, opened 2026-05-14
feat(alerting): Block non-PPL monitor CRUD on pluggable dataformat domains

yashgupte21:
feat(alerting): Block non-PPL monitor CRUD on pluggable dataformat domains

Reject creation, update, and execution of non-PPL monitors when the pluggable dataformat feature flag is enabled. Only PPL monitors are allowed on these domains. Delete and read operations are not blocked.

Uses FeatureFlags.isEnabled(PLUGGABLE_DATAFORMAT_EXPERIMENTAL_FLAG) from OpenSearch core to gate the behavior.

Also fixes missing settings registration in TransportIndexMonitorAction unit tests (MULTI_TENANT_TRIGGER_EVAL_ENABLED, PPL_MAX_QUERY_LENGTH, PPL_QUERY_RESULTS_MAX_DATAROWS, NOTIFICATION_SUBJECT_SOURCE_MAX_LENGTH, NOTIFICATION_MESSAGE_SOURCE_MAX_LENGTH) which were not reflected in the test setup.

### Description
[Describe what this change achieves]

### Related Issues

### Check List

### DIFF
--- a/alerting/build.gradle
+++ b/alerting/build.gradle
@@ -392,6 +392,46 @@ integTest {
             excludeTestsMatching "org.opensearch.alerting.bwc.*IT"
         }
     }
+
+    task integTestPluggableDataformat(type: RestIntegTestTask) {
+        description = "Runs integration tests with pluggable dataformat feature flag enabled"
+        dependsOn bundlePlugin
+        testClassesDirs = sourceSets.test.output.classesDirs
+        classpath = sourceSets.test.runtimeClasspath
+
+        systemProperty 'tests.security.manager', 'false'
+        systemProperty 'java.io.tmpdir', opensearch_tmp_dir.absolutePath
+
+        filter {
+            includeTestsMatching "org.opensearch.alerting.transport.PluggableDataFormatMonitorBlockIT"
+        }
+    }
+
+    testClusters.integTestPluggableDataformat {
+        testDistribution = "ARCHIVE"
+        systemProperty 'opensearch.experimental.feature.pluggable.dataformat.enabled', 'true'
+
+        // Include same plugins as main integTest cluster
+        plugin(provider({ new RegularFile() { @Override File getAsFile() { return configurations.zipArchive.asFileTree.matching { include '**/opensearch-notifications-core*' }.singleFile } } }))
+        plugin(provider({ new RegularFile() { @Override File getAsFile() { return configurations.zipArchive.asFileTree.matching { include '**/notifications*' }.singleFile } } }))
+        plugin(provider({ new RegularFile() { @Override File getAsFile() { return configurations.zipArchive.asFileTree.matching { include '**/opensearch-job-scheduler*' }.singleFile } } }))
+        plugin(provider({ new RegularFile() { @Override File getAsFile() { return configurations.zipArchive.asFileTree.matching { include '**/opensearch-sql-plugin*' }.singleFile } } }))
+
+    }
+
+    // Install alerting plugin on the pluggable dataformat test cluster
+    integTestPluggableDataformat.dependsOn(bundle)
+    integTestPluggableDataformat.getClusters().forEach{c -> c.plugin(project.getObjects().fileProperty().value(bundle.getArchiveFile()))}
+
+    // Exclude pluggable dataformat tests from the main integTest task
+    integTest {
+        filter {
+            excludeTestsMatching "org.opensearch.alerting.transport.PluggableDataFormatMonitorBlockIT"
+        }
+    }
+
+    check.dependsOn integTestPluggableDataformat
+    integTest.finalizedBy integTestPluggableDataformat
 }
 
 task integTestRemote(type: RestIntegTestTask) {
--- a/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportExecuteMonitorAction.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportExecuteMonitorAction.kt
@@ -27,6 +27,7 @@ import org.opensearch.alerting.util.use
 import org.opensearch.cluster.service.ClusterService
 import org.opensearch.common.inject.Inject
 import org.opensearch.common.settings.Settings
+import org.opensearch.common.util.FeatureFlags
 import org.opensearch.common.xcontent.LoggingDeprecationHandler
 import org.opensearch.common.xcontent.XContentHelper
 import org.opensearch.common.xcontent.XContentType
@@ -35,6 +36,7 @@ import org.opensearch.commons.alerting.model.Monitor
 import org.opensearch.commons.alerting.model.ScheduledJob
 import org.opensearch.commons.alerting.util.AlertingException
 import org.opensearch.commons.alerting.util.isMonitorOfStandardType
+import org.opensearch.commons.alerting.util.isPPLMonitor
 import org.opensearch.commons.authuser.User
 import org.opensearch.commons.utils.TenantContext
 import org.opensearch.core.action.ActionListener
@@ -201,6 +203,22 @@ class TransportExecuteMonitorAction @Inject constructor(
                             OpenSearchStatusException(
                                 "${monitor.monitorType} monitors are not allowed when multi-tenancy is enabled.",
                                 RestStatus.METHOD_NOT_ALLOWED
+                            )
+                        )
+                    )
+                    return@use
+                }
+
+                // Block non-PPL monitors on pluggable dataformat domains
+                if (FeatureFlags.isEnabled(FeatureFlags.PLUGGABLE_DATAFORMAT_EXPERIMENTAL_FLAG) &&
+                    !monitor.isPPLMonitor()
+                ) {
+                    actionListener.onFailure(
+                        AlertingException.wrap(
+                            OpenSearchStatusException(
+                                "Only PPL monitors are supported on this domain." +
+                                    " ${monitor.monitorType} monitors are not allowed.",
+                                RestStatus.FORBIDDEN
                             )
                         )
                     )
--- a/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
@@ -62,6 +62,7 @@ import org.opensearch.cluster.service.ClusterService
 import org.opensearch.common.inject.Inject
 import org.opensearch.common.settings.Settings
 import org.opensearch.common.unit.TimeValue
+import org.opensearch.common.util.FeatureFlags
 import org.opensearch.common.xcontent.LoggingDeprecationHandler
 import org.opensearch.common.xcontent.XContentFactory
 import org.opensearch.common.xcontent.XContentHelper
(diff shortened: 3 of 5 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportExecuteMonitorAction.kt
```diff
@@ -148,6 +150,24 @@ class TransportExecuteMonitorAction @Inject constructor(
                                     )
                                     return@whenComplete
                                 }
+
+                                // Block non-PPL monitors on pluggable dataformat domains
+                                if (FeatureFlags.isEnabled(
```
toepkerd: Since we are enforcing the invariant that a pluggable dataformat cluster is blocked from creating a non-PPL Monitor, this check may not be necessary.

For context, this if else branch handles 2 cases of Monitor execution:
1. The Monitor was already created and exists in the cluster, execute it via its ID
2. The Monitor does not exist in the cluster, but the configuration of the Monitor to be executed is passed into the Execute Monitor request.

Right now, you've added the check for both cases. I believe that since we're blocking non-PPL Monitor creation, we can remove the check in this if branch (case 1), but we must keep the check in the else branch (case 2).

alerting/src/main/kotlin/org/opensearch/alerting/util/AlertingUtils.kt
```diff
@@ -107,6 +107,12 @@ fun Monitor.isUnsupportedMultiTenantMonitorType(): Boolean {
     return type == Monitor.MonitorType.DOC_LEVEL_MONITOR || type == Monitor.MonitorType.CLUSTER_METRICS_MONITOR
 }
 
+fun Monitor.isNonPPLMonitor(): Boolean {
```
toepkerd: An isPPLMonitor() helper already exists in common utils, let's use that instead: https://github.com/opensearch-project/common-utils/blob/ae0f22333c41fdfb1f649a837fc44af0403b3e95/src/main/kotlin/org/opensearch/commons/alerting/util/IndexUtils.kt#L53

Furthermore, PPL Monitors are considered a standard type, so saying a Monitor is non-PPL when it is not a standard Monitor type will not give expected behavior.

alerting/src/test/kotlin/org/opensearch/alerting/transport/PluggableDataFormatMonitorBlockIT.kt
```diff
@@ -0,0 +1,277 @@
...
+class PluggableDataFormatMonitorBlockIT : AlertingSingleNodeTestCase() {
+
+    override fun nodeSettings(): Settings {
+        return Settings.builder()
+            .put(super.nodeSettings())
+            .put(FeatureFlags.PLUGGABLE_DATAFORMAT_EXPERIMENTAL_FLAG, true)
+            .build()
+    }
+
+    fun `test create query-level monitor blocked on pluggable dataformat domain`() {
```
toepkerd: You've added tests to ensure creation of query-level and document-level Monitors are blocked. For thoroughness, let's also add individual tests to ensure each non-PPL Monitor type gets blocked. In other words, let's add this exact test for bucket-level monitors and cluster metrics Monitors.

Open discussion: should we block the creation of workflows (Composite Monitors)? My current inclination is yes since PPL Monitors as part of chained Monitors haven't been tested yet.

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportExecuteMonitorAction.kt
yashgupte21: Yeah that makes sense, will update this check and keep it only for else branch (case2).

alerting/src/main/kotlin/org/opensearch/alerting/util/AlertingUtils.kt
yashgupte21: Ack, will use the helper `isPPLMonitor()` from common utils.

alerting/src/test/kotlin/org/opensearch/alerting/transport/PluggableDataFormatMonitorBlockIT.kt
yashgupte21: Adding tests for bucket level and cluster level metrics monitors.

>Open discussion: should we block the creation of workflows (Composite Monitors)? My current inclination is yes since PPL Monitors as part of chained Monitors haven't been tested yet.

I agree with this point about creation of workflows composite monitors. I'll add the check in `TransportIndexWorkflowAction` for composite monitors.

alerting/src/test/kotlin/org/opensearch/alerting/transport/PluggableDataFormatMonitorBlockIT.kt
```diff
@@ -0,0 +1,371 @@
...
+                assertFalse(
+                    "PPL monitor should not be blocked by pluggable dataformat gate, got: $message",
+                    message.contains("Only PPL monitors are supported") ||
+                        message.contains("monitors are not allowed")
+                )
+            }
+        }
+    }
+
+    fun `test update query-level monitor blocked on pluggable dataformat domain`() {
```
toepkerd: Can you briefly explain this test? If query level monitor creation is blocked, it should be impossible to update a query-level monitor because one would never have existed in the first place.

yashgupte21: That test is checking if it also fires for `PUT` requests and not just `POST`.  Since `TransportIndexMonitorAction` handles both create and update, the test verifies the gate blocks both paths. The monitor doesn't need to exist, the check rejects based on the monitor type in the request body before any index lookup happens. Keeping it to cover both paths.

alerting/src/test/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorActionTests.kt
```diff
@@ -69,6 +97,12 @@ class TransportIndexMonitorActionTests : OpenSearchTestCase() {
         settingSet.add(AlertingSettings.JOB_QUEUE_NAME)
         settingSet.add(AlertingSettings.EXTERNAL_SCHEDULER_ROLE_NAME)
         settingSet.add(AlertingSettings.EXTERNAL_SCHEDULER_EXECUTION_ROLE_NAME)
+        settingSet.add(AlertingSettings.MULTI_TENANT_TRIGGER_EVAL_ENABLED)
```
toepkerd: Is this setting necessary? I ctrl-F ed the PR file changes page for this constant and it only appears here

toepkerd: EDIT: this seems to be the case for each of the PPL Alerting related settings added here, may be an artifact of rebase conflict resolutions?

yashgupte21: Yes the `TransportIndexMonitorAction` 's init block registers an update consumer for it. Without it in the test's `ClusterSetting` the constructor is throwing a `SettingsException`.  The test fail on `main` were missing these.

alerting/src/test/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorActionTests.kt
```diff
@@ -228,4 +262,338 @@ class TransportIndexMonitorActionTests : OpenSearchTestCase() {
     fun `test tenant_id metadata key constant value`() {
         assertEquals("tenant_id", AlertingPlugin.TENANT_ID_METADATA_KEY)
     }
+
+    fun `test index query-level monitor blocked on pluggable dataformat domain`() {
```
toepkerd: These unit tests mock case scenarios that your integ tests actually run on a spun up local node. Let's dedupe these tests that mock case scenarios already covered by your integ tests.

yashgupte21: Got it, will remove unit tests and keep integ.

alerting/build.gradle
```diff
@@ -392,6 +392,46 @@ integTest {
             excludeTestsMatching "org.opensearch.alerting.bwc.*IT"
         }
     }
+
+    task integTestPluggableDataformat(type: RestIntegTestTask) {
```
eirsep, line 396: how will this task run in our github CIs 
can you edit github workflow.yml and add this too

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexWorkflowAction.kt
```diff
@@ -150,6 +151,22 @@ class TransportIndexWorkflowAction @Inject constructor(
             return
         }
 
+        // Block workflow creation on pluggable dataformat domains
+        if (FeatureFlags.isEnabled(
```
eirsep, line 155: shouldn't we be blocking get workflow alerts, acknowledge workflow alerts, chained alerts etc.

eirsep: can be taken up on followup pr

eirsep (dismissed):

alerting/build.gradle
yashgupte21: In build.gradle added [integTest.finalizedBy](https://github.com/opensearch-project/alerting/pull/2141/changes#diff-612ccdeaa466e3f74bb985ad6fab92b6f9006e8f883c2582f03e8b49fbb3b556R434)  which automatically runs [integTestPluggableDataFormat](https://github.com/opensearch-project/alerting/actions/runs/25938516461/job/76249966986?pr=2141#step:6:243) task after integTest task , they're included in the CIs. Adding it to workflows could be taken in follow up PR to double guard this .

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexWorkflowAction.kt
yashgupte21: Agree since workflows can not be created on pluggable dataformat domains, those API would never have data to operate on, but adding explicit blocks for completeness makes sense and can take that as a follow up PR.

toepkerd (approved):

alerting/src/main/kotlin/org/opensearch/alerting/transport/TransportIndexMonitorAction.kt
```diff
@@ -213,6 +215,22 @@ class TransportIndexMonitorAction @Inject constructor(
...
+                AlertingException.wrap(
+                    OpenSearchStatusException(
+                        "Only PPL monitors are supported on this domain." +
+                            " ${transformedRequest.monitor.monitorType} monitors are not allowed.",
+                        RestStatus.FORBIDDEN
+                    )
+                )
+            )
+            return
+        }
```
vikhy-aws, line 232: nit: Code repetition. Can we pull this out into a utility method?

vikhy-aws (approved):

PR #2143, opened 2026-05-15
feat: Propagate tenant headers for remote monitor execution

eirsep:
Parse monitor target ARN to extract account ID and resource ID, then set them as thread context headers so search requests can be routed to the correct remote data source.

Changes:
- Add plugins.alerting.tenant.account_id_header setting
- Add plugins.alerting.tenant.resource_id_header.<TYPE> group setting
- Add ARN parsing and putTransient() in MonitorJobPoller.populateThreadContext
- Add reinjectHeaders() in QueryLevelMonitorRunner and BucketLevelMonitorRunner to re-set headers inside withClosableContext
- Add ArnValidator for create monitor request validation
- Fix RemoteQueryLevelTriggerEvaluator to handle typed_keys agg key prefix
- Register new settings in AlertingPlugin.getSettings()
- Add tests for ARN parsing and tenant header propagation

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunner.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunner.kt
@@ -9,6 +9,9 @@ import org.opensearch.alerting.opensearchapi.InjectorContextElement
 import org.opensearch.alerting.opensearchapi.withClosableContext
 import org.opensearch.alerting.script.QueryLevelTriggerExecutionContext
 import org.opensearch.alerting.script.TriggerExecutionContext
+import org.opensearch.alerting.service.MonitorJobPoller
+import org.opensearch.alerting.settings.AlertingSettings
+import org.opensearch.alerting.util.ArnHelper
 import org.opensearch.alerting.util.getConfigAndSendNotification
 import org.opensearch.alerting.util.use
 import org.opensearch.commons.ConfigConstants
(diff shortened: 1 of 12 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/service/MonitorJobPoller.kt
```diff
@@ -51,7 +51,9 @@ class MonitorJobPoller(
     private val accountIdProvider: JobQueueAccountIdProvider?,
     private val region: String,
     private val queueName: String,
-    private val targetTypeToServiceName: Map<String, String>
+    private val targetTypeToServiceName: Map<String, String>,
+    private val tenantAccountIdHeader: String = "",
+    private val tenantResourceIdHeaders: Map<String, String> = emptyMap()
```
engechas: nit: tenantAccountIdHeaderName and tenantResourceIdHeaderNames

vikhy-aws (approved):

alerting/src/main/kotlin/org/opensearch/alerting/service/MonitorJobPoller.kt
```diff
@@ -249,6 +251,29 @@ class MonitorJobPoller(
...
+                threadContext.putTransient(tenantAccountIdHeader, accountId)
+            }
+            val resourceHeader = tenantResourceIdHeaders[target.type]
+            if (!resourceHeader.isNullOrEmpty()) {
+                threadContext.putTransient(resourceHeader, resourceId)
+            }
+        }
+    }
+
+    internal fun parseArn(arn: String): Pair<String, String> {
```
engechas: We should have access to the Arn construct from the AWS SDK dependencies. We can use that to validate and deconstruct the ARN

alerting/src/main/kotlin/org/opensearch/alerting/MonitorRunner.kt
```diff
@@ -85,4 +87,41 @@ abstract class MonitorRunner {
...
+        if (threadContext.getHeader(MonitorJobPoller.SERVICE_NAME_HEADER) == null) {
+            val serviceNameMap = AlertingSettings.TARGET_TYPE_TO_SERVICE_NAME.get(settings).let { it.keySet().associateWith { key -> it.get(key) } }
+            threadContext.putHeader(MonitorJobPoller.SERVICE_NAME_HEADER, serviceNameMap[target.type] ?: target.type)
+        }
+        if (threadContext.getHeader(MonitorJobPoller.REGION_HEADER) == null) {
+            threadContext.putHeader(MonitorJobPoller.REGION_HEADER, AlertingSettings.REMOTE_METADATA_REGION.get(settings) ?: "")
+        }
+
+        if (target.arn.isNotBlank()) {
+            val parts = target.arn.split(":")
```
engechas: Same comment on using `Arn` from the SDK. We could also consolidate this logic with the Poller and the Validator to instead have an ArnHelper class

vikhy-aws: Can we reuse `parseArn` here? Reusing it should also handle ARN validation.

engechas (approved):
No blocking comments but it would great to have the comments addressed before merging

engechas (approved):

PR #2158, opened 2026-05-23
Fix: Skip alias-type fields in doc-level monitor query index mapping

### LINKED ISSUE #2157: [BUG] Doc-level monitor upsertQueryIndex fails with HTTP 500 when source index contains alias-type fields
# [BUG] Doc-level monitor upsertQueryIndex fails with HTTP 500 when source index contains alias-type fields

## Describe the issue

When a doc-level monitor is created (or updated) and one or more of its source indices contains fields with `"type": "alias"`, the alerting plugin's `DocLevelMonitorQueries.upsertQueryIndex` method fails with an HTTP 500 error. The monitor is never created and no useful error is surfaced at production log levels.

The failure occurs in `leafNodeProcessor`, which traverses all field mappings from all backing indices and copies them into the percolate query index (`.opensearch-alerting-queries-*`). When it encounters an alias field it rewrites the alias `path` value by appending `_<indexName>_<monitorId>`. This synthesized path does not exist as a concrete field in the query index. OpenSearch rejects the subsequent `PUT _mapping` call because an alias must refer to an existing field in the same index's mappings.

The plugin is not idempotent with respect to its own prior output: alias fields written to backing indices during a previous monitor lifecycle are re-discovered on the next `upsertQueryIndex` call and trigger the same failure.

## Expected behavior

`upsertQueryIndex` completes successfully regardless of whether source indices contain `type: alias` fields. Alias fields are silently skipped during mapping traversal because they are index-local contracts and cannot be transplanted to an index with a different schema. The percolate query index holds stored `query_string` queries, not documents; alias resolution is never used when matching percolate queries, so alias fields in the query index serve no semantic purpose.

## Actual behavior

`PUT _mapping` on the query index fails with:

```
Invalid [path] value [<originalPath>_<backingIndexName>_<monitorId>] for field alias [...]:
an alias must refer to an existing field in the mappings.
[type=security_analytics_exception]
```

The initial failure is logged at `log.debug`, making it invisible at the default production log level. A `deleteQueryIndexInEveryRun=true` retry path is attempted but produces the same error because the same source mappings generate the same invalid synthesized path.

## Steps to reproduce

1. Create an index (or use a data stream) that contains at least one field mapping with `"type": "alias"`. Example mapping fragment:

   ```json
   PUT myIndex/_mapping
   {
     "properties": {
       "myAlias": {
         "type": "alias",
         "path": "myConcreteField"
       },
       "myConcreteField": {
         "type": "keyword"
       }
     }
   }
   ```

2. Create a doc-level monitor that references `myIndex` as its source index.
3. Delete the monitor.
4. Re-create the doc-level monitor referencing the same index.
5. The second creation triggers `upsertQueryIndex`, which re-traverses the backing index. The alias field `myAlias` is present from step 1. The plugin produces a synthesized path `myConcreteField_myIndex_<monitorId>` and submits it in a `PUT _mapping` call that OpenSearch rejects with HTTP 500.

The bug also triggers on the first monitor creation if alias fields pre-exist in the source index for any reason.

## Root cause analysis

**File:** `alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt`

**Primary defect — `leafNodeProcessor`:**
The function processes every leaf field in the source index mappings. When it encounters a field with `"type": "alias"`, it falls through to the general rewrite branch, which appends `_<indexName>_<monitorId>` to the alias `path` value. The resulting path is synthetic and does not exist in the query index, causing `PUT _mapping` to be rejected.

Both the conflicting-name branch and the non-conflicting branch of `leafNodeProcessor` rewrite the alias `path` to a non-existent value. Neither branch produces a valid outcome for alias fields.

**Secondary defect — in-place mutation:**
`traverseMappingsAndUpdate` mutates `MappingMetadata.sourceAsMap` in-place. Because `MappingMetadata.sourceAsMap()` is documented to return a reference to the underlying map, concurrent monitor operations on the same index share and corrupt this state.

**Fix (primary — two-line guard):**
Add a guard at the entry of `leafNodeProcessor` that returns early for any field where `props["type"] == "alias"`. The returned triple uses an empty `mutableMapOf()` so the field is excluded from `updatedProperties` and never submitted to `PUT _mapping`. A matching nil-check must be added in the `properties.forEach` accumulation loop that calls `leafNodeProcessor`:

```kotlin
// In leafNodeProcessor, before val newProps = props.toMutableMap():
if (props["type"] == "alias") {
    return Triple(fieldName, fieldName, mutableMapOf())
}
```

The accumulation loop should skip entries where the returned properties map is empty (i.e., do not write an empty property entry to `updatedProperties`).

## Related components

- `opensearch-project/alerting` — primary fix location (`DocLevelMonitorQueries.kt`)
- `opensearch-project/security-analytics` — secondary issue: the SA plugin writes `type: alias` fields to backing indices when a detector is created and does not clean them up on deletion. This permanent residue is the most common trigger for the alerting bug. Tracked separately in `opensearch-project/security-analytics`.

thecodingshrimp:
# Fix: Skip alias-type fields in doc-level monitor query index mapping

## Related issue

Fixes: #2157 

---

## Summary

`DocLevelMonitorQueries.upsertQueryIndex` fails with HTTP 500 whenever a source index contains fields with `"type": "alias"`. This PR adds a two-line guard in `leafNodeProcessor` that skips alias fields during mapping traversal, preventing invalid `PUT _mapping` calls to the percolate query index.

The fix is minimal and surgical: it does not change how any other field type is handled.

---

## Root cause

**File:** `alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt`
**Function:** `leafNodeProcessor`

When `upsertQueryIndex` builds the query index mappings it calls `traverseMappingsAndUpdate`, which recursively walks every leaf field of every source index and delegates to `leafNodeProcessor`. For each leaf, `leafNodeProcessor` rewrites field names and the alias `path` value by appending `_<indexName>_<monitorId>` to produce a per-monitor suffix. For `type: alias` fields this produces a synthesized path that does not exist as a concrete field in the query index. OpenSearch then rejects the `PUT _mapping` call:

```
Invalid [path] value [<originalPath>_<backingIndexName>_<monitorId>] for field alias [...]:
an alias must refer to an existing field in the mappings.
```

The percolate query index (`.opensearch-alerting-queries-*`) stores `query_string` queries, not documents. Alias resolution is not used during percolate matching. Alias fields copied to the query index therefore serve no semantic purpose — they are index-local contracts that are only meaningful in the context of the index where they are defined.

The plugin is not idempotent: alias fields left in a source index by a prior monitor lifecycle are re-discovered on every subsequent `upsertQueryIndex` call and trigger the same failure. The `deleteQueryIndexInEveryRun=true` retry path does not escape this because the same source mappings produce the same invalid synthesized path on retry.

---

## Changes

**`alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt`**

1. **`leafNodeProcessor` — early return for alias fields.**
   Add a guard before the existing `val newProps = props.toMutableMap()` line. When `props["type"] == "alias"`, return a triple with an empty `mutableMapOf()` as the properties value. This signals the accumulation loop that the field should be excluded from `updatedProperties`.

   ```kotlin
   // Guard added at entry of leafNodeProcessor:
   if (props["type"] == "alias") {
       return Triple(fieldName, fieldName, mutableMapOf())
   }
   ```

2. **`traverseMappingsAndUpdate` / `properties.forEach` accumulation loop — skip empty property maps.**
   After calling `leafNodeProcessor`, check whether the returned properties map is empty. If it is, do not write an entry to `updatedProperties`. This prevents a zero-property alias stub from being submitted in the `PUT _mapping` body.

   ```kotlin
   // In the forEach loop that builds updatedProperties:
   val (newKey, _, newProps) = leafNodeProcessor(...)
   if (newProps.isNotEmpty()) {
       updatedProperties[newKey] = newProps
   }
   ```

These two changes together ensure that alias fields never appear in the `PUT _mapping` request body sent to the query index.

**No changes to any public API, monitor data model, or query execution path.** The query index schema for all non-alias field types is unaffected.

---

## Testing notes

1. **Unit test — alias field excluded from query index mappings.**
   Create a `MappingMetadata` fixture that includes a `type: alias` field alongside a concrete keyword field. Assert that after `traverseMappingsAndUpdate` the returned map contains the concrete field and does not contain the alias field.

2. **Integration test — monitor creation succeeds on index with alias fields.**
   - Create an index with at least one `type: alias` field and one concrete field.
   - Create a doc-level monitor targeting that index.
   - Assert monitor creation returns HTTP 200 and monitor status is `ACTIVE`.
   - Verify the query index (`.opensearch-alerting-queries-*`) mapping does not contain any field with `"type": "alias"`.

3. **Regression test — idempotency across delete/re-create cycle.**
   - Create a doc-level monitor, delete it, and re-create it against the same index.
   - Assert the second creation succeeds (HTTP 200) with no `PUT _mapping` 500 error in logs.

4. **Existing test suite.**
   Run the full `DocLevelMonitorQueries` test class and the doc-level monitor integration tests. No existing test should regress: the guard only affects fields with `"type": "alias"`, which the existing tests do not exercise.

---

## Out of scope

The SA plugin (`opensearch-project/security-analytics`) writes `type: alias` fields to backing indices on detector creation and does not remove them on detector deletion. That cleanup is tracked separately. This PR makes the alerting plugin resilient to alias fields regardless of their origin, which addresses the immediate breakage independently of when the SA cleanup lands.

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/util/DocLevelMonitorQueries.kt
@@ -64,6 +64,21 @@ class DocLevelMonitorQueries(private val client: Client, private val clusterServ
         const val TYPE = "type"
         const val INDEX_PATTERN_SUFFIX = "-000001"
         const val QUERY_INDEX_BASE_FIELDS_COUNT = 8 // 3 fields we defined and 5 builtin additional metadata fields
+
+        /**
+         * Returns a mutable deep copy of [map] so that callers can mutate the result
+         * without affecting the original (e.g. cached cluster-state data).
+         */
+        @Suppress("UNCHECKED_CAST")
+        fun deepCopyMap(map: Map<String, Any>): MutableMap<String, Any> =
+            map.entries.associate { (k, v) ->
+                k to when (v) {
+                    is Map<*, *> -> deepCopyMap(v as Map<String, Any>)
+                    is List<*> -> (v as List<Any>).toMutableList()
+                    else -> v
+                }
+            }.toMutableMap()
+
         @JvmStatic
         fun docLevelQueriesMappings(): String {
             return DocLevelMonitorQueries::class.java.classLoader.getResource("mappings/doc-level-queries.json").readText()
@@ -297,15 +312,18 @@ class DocLevelMonitorQueries(private val client: Client, private val clusterServ
                 if (clusterState.routingTable.hasIndex(concreteIndexName)) {
                     val indexMetadata = clusterState.metadata.index(concreteIndexName)
                     if (indexMetadata.mapping()?.sourceAsMap?.get("properties") != null) {
-                        val properties = (
-                            (indexMetadata.mapping()?.sourceAsMap?.get("properties"))
-                                as MutableMap<String, Any>
-                            )
+                        @Suppress("UNCHECKED_CAST")
+                        val properties = deepCopyMap(
+                            indexMetadata.mapping()?.sourceAsMap?.get("properties") as Map<String, Any>
+                        )
                         // Node processor function is used to process leaves of index mappings tree
                         //
                         val leafNodeProcessor =
                             fun(fieldName: String, fullPath: String, props: MutableMap<String, Any>):
                                 Triple<String, String, MutableMap<String, Any>> {
+                                if (props["type"] == "alias") {
+                                    return Triple(fieldName, fieldName, mutableMapOf())
+                                }
                                 val newProps = props.toMutableMap()
                                 if (monitor.dataSources.queryIndexMappingsByType.isNotEmpty()) {
                                     val mappingsByType = monitor.dataSources.queryIndexMappingsByType
@@ -338,6 +356,13 @@ class DocLevelMonitorQueries(private val client: Client, private val clusterServ
                                 it.value is Map<*, *> &&
                                 (it.value as Map<String, Any>).containsKey("type") &&
                                 (it.value as Map<String, Any>)["type"] == NESTED
+                            ) {
+                            } else if (
+                                it.value is Map<*, *> &&
+                                (
+                                    (it.value as Map<String, Any>)["type"] == "alias" ||
+                                        (it.value as Map<String, Any>).isEmpty()
+                                    )
                             ) {
                             } else {
                                 if (updatedProperties.containsKey(it.key) && updatedProperties[it.key] != it.value) {
@@ -526,7 +551,7 @@ class DocLevelMonitorQueries(private val client: Client, private val clusterServ
             return Pair(updateMappingResponse, targetQueryIndex)
         } catch (e: Exception) {
             val unwrappedException = ExceptionsHelper.unwrapCause(e) as Exception
-            log.debug("exception after rollover queryIndex index: $targetQueryIndex exception: ${unwrappedException.message}")
+            log.warn("PUT mapping failed on queryIndex $targetQueryIndex [${unwrappedException.javaClass.simpleName}]: ${unwrappedException.message}")
             // If we reached limit for total number of fields in mappings, do a rollover here
             if (unwrappedException.message?.contains("Limit of total fields") == true) {
                 try {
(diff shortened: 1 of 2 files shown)


thecodingshrimp:
@toepkerd-zz @eirsep what do you think of this PR?

Issue #2086, opened 2026-04-05
[BUG] Deleting a Monitor while it is being executed orphans Alerts


toepkerd:
**What is the bug?**
Given a Monitor that has an ACTIVE ongoing Alert, if a user deletes that Monitor in the middle of one of its executions, a race condition causes Alert duplication/orphaning where the ongoing Alert continues to exist in the `.opendistro-alerting-alerts` index as ACTIVE, but the same Alert is written to the `.opendistro-alerting-alert-history*` index pattern as DELETED.

On the side, trying to acknowledge the orphaned ACTIVE alert leads to a stuck "acknowledge alert" task.

The longer monitor executions run, the wider the race condition window, and the more likely this bug is to occur.

**What is the expected behavior?**
The presence of the DELETED Alert is correct. However, the ACTIVE Alert should be deleted. This suggests an issue with AlertMover.kt's postDelete() flows, where the Alert is successfully copied but somehow it does not get deleted. Strangely, no "Failed to delete alerts" error logs or exceptions were found in the Elasticsearch logs.

**How can one reproduce the bug?**
1. Create an index and send continuous voluminous data to it. Example mappings:
```
"mappings": {
    "properties": {
        # basic mock fields
        "@timestamp": {"type": "date"},
        "status": {"type": "text"},
        "application": {"type": "text"},
        "severity": {"type": "text"},
        "number": {"type": "integer"},

        # voluminous mock fields
        "log_message": {"type": "text"},
        "stack_trace": {"type": "text"},
        "request_payload": {"type": "text"},
        "response_payload": {"type": "text"},

        # metadata mock fields
        "host": {"type": "text"},
        "pod": {"type": "text"},
        "container": {"type": "text"},
    }
}
```
2. Create a query-level monitor on that index with an expensive query to make executions run longer. The longer you make a Monitor execution last, the easier the bug will be to reproduce. Example config:
```
{
    "monitor_type": "query_level_monitor",
    "name": "Any Results Monitor",
    "enabled": true,
    "schedule": {"period": {"interval": 1, "unit": "MINUTES"}},
    "inputs": [
        {
            "search": {
                "indices": <index_name]>,
                "query": {
                    "size": 10000,
                    "query": {"wildcard": {"log_message": {"value": "*e*"}}},
                    "aggs": {
                        "nested_agg": {
                            "terms": {"field": "host.keyword", "size": 5000},
                            "aggs": {
[...]

PR #2175, opened 2026-06-05
feat(alerting): Add OrphanedAlertSweeper and fix JobSweeper to desche…

toepkerd:
### Description

Add OrphanedAlertSweeper: a scheduled background job that periodically scans for alerts whose monitors no longer exist in .opendistro-alerting-config. Orphaned alerts are moved to the alert history index with state DELETED.

Fix JobSweeper.sweepShard to deschedule monitors that are in the in-memory sweptJobs map but no longer exist in the config index. Previously, the background sweep only handled new/updated jobs and shard ownership changes, relying entirely on postDelete for deletions. If postDelete failed to fire (e.g., due to IndexingOperationListener not triggering), the monitor would remain scheduled indefinitely, producing repeated 'Can't find monitor' errors.

Note: this is not a bug fix for postDelete, but an anti entropy mechanism that maintains correct Alerting state while the bug gets root caused and fixed.

### Related Issues
https://github.com/opensearch-project/alerting/issues/2086

### Check List
- [Y] New functionality includes testing.
- [Y] Commits are signed per the DCO using `--signoff`.

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/alerts/OrphanedAlertSweeper.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/alerts/OrphanedAlertSweeper.kt
@@ -5,8 +5,11 @@
 
 package org.opensearch.alerting.alerts
 
+import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withTimeout
 import org.apache.logging.log4j.LogManager
@@ -18,6 +21,7 @@ import org.opensearch.alerting.settings.AlertingSettings
 import org.opensearch.cluster.ClusterChangedEvent
 import org.opensearch.cluster.ClusterStateListener
 import org.opensearch.cluster.service.ClusterService
+import org.opensearch.common.lifecycle.LifecycleListener
 import org.opensearch.common.settings.Settings
 import org.opensearch.commons.alerting.model.Alert
 import org.opensearch.commons.alerting.model.ScheduledJob
@@ -51,10 +55,13 @@ class OrphanedAlertSweeper(
     private val client: Client,
     private val threadPool: ThreadPool,
     private val clusterService: ClusterService
-) : ClusterStateListener {
+) : ClusterStateListener, LifecycleListener() {
 
     private val logger = LogManager.getLogger(OrphanedAlertSweeper::class.java)
-    private val scope = CoroutineScope(Dispatchers.IO)
+    private val exceptionHandler = CoroutineExceptionHandler { _, exception ->
+        logger.error("Uncaught exception in OrphanedAlertSweeper coroutine", exception)
+    }
+    private val scope = CoroutineScope(SupervisorJob() + Dispatchers.IO + exceptionHandler)
 
     @Volatile private var sweepPeriod = SWEEP_PERIOD.get(settings)
     @Volatile private var enabled = ENABLED.get(settings)
@@ -63,6 +70,7 @@ class OrphanedAlertSweeper(
 
     init {
         clusterService.addListener(this)
+        clusterService.addLifecycleListener(this)
         clusterService.clusterSettings.addSettingsUpdateConsumer(SWEEP_PERIOD) {
             sweepPeriod = it
             if (clusterService.state().nodes.isLocalNodeElectedClusterManager && enabled) reschedule()
@@ -73,6 +81,22 @@ class OrphanedAlertSweeper(
                 if (enabled) reschedule() else cancel()
             }
         }
+    }
+
+    override fun afterStart() {
+        if (clusterService.state().nodes.isLocalNodeElectedClusterManager && enabled) {
+            isClusterManager = true
+            reschedule()
+        }
+    }
+
+    override fun beforeStop() {
+        cancel()
+        clusterService.removeListener(this)
+    }
+
+    override fun beforeClose() {
+        scope.cancel()
     }
 
     override fun clusterChanged(event: ClusterChangedEvent) {
--- a/alerting/src/main/kotlin/org/opensearch/alerting/alerts/OrphanedAlertSweeper.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/alerts/OrphanedAlertSweeper.kt
@@ -1,0 +1,180 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.opensearch.alerting.alerts
+
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withTimeout
+import org.apache.logging.log4j.LogManager
+import org.opensearch.action.search.SearchRequest
+import org.opensearch.action.search.SearchResponse
+import org.opensearch.alerting.alerts.AlertIndices.Companion.ALERT_INDEX
+import org.opensearch.alerting.opensearchapi.suspendUntil
+import org.opensearch.alerting.settings.AlertingSettings
+import org.opensearch.cluster.ClusterChangedEvent
+import org.opensearch.cluster.ClusterStateListener
+import org.opensearch.cluster.service.ClusterService
+import org.opensearch.common.settings.Settings
+import org.opensearch.commons.alerting.model.Alert
+import org.opensearch.commons.alerting.model.ScheduledJob
+import org.opensearch.index.query.QueryBuilders
+import org.opensearch.search.aggregations.bucket.terms.Terms
+import org.opensearch.search.aggregations.bucket.terms.TermsAggregationBuilder
+import org.opensearch.search.builder.SearchSourceBuilder
+import org.opensearch.threadpool.Scheduler
+import org.opensearch.threadpool.ThreadPool
+import org.opensearch.transport.client.Client
+
+/**
+ * An anti-entropy background job that periodically scans for alerts whose monitors no longer
+ * exist in .opendistro-alerting-config ("orphaned alerts") and moves them to the alert history
+ * index with state = DELETED.
+ *
+ * Normally, when a monitor is deleted, the JobSweeper's IndexingOperationListener.postDelete
+ * callback fires, which deschedules the monitor and moves its alerts to history via AlertMover.
+ * However, postDelete does not always fire reliably — particularly when a monitor deletion
+ * occurs while one of its executions is in flight. In this case,
+ * alerts remain stranded in the active alerts index indefinitely.
+ *
+ * This sweeper is not a bug fix for postDelete — it is a safety net. It runs independently on the
+ * cluster manager node and cleans up any orphaned alerts that postDelete missed, ensuring the
+ * system eventually converges to a consistent state.
+ *
+ * Runs only on the cluster manager node to avoid duplicate work.
+ */
+class OrphanedAlertSweeper(
+    settings: Settings,
+    private val client: Client,
+    private val threadPool: ThreadPool,
+    private val clusterService: ClusterService
+) : ClusterStateListener {
+
+    private val logger = LogManager.getLogger(OrphanedAlertSweeper::class.java)
+    private val scope = CoroutineScope(Dispatchers.IO)
+
+    @Volatile private var sweepPeriod = SWEEP_PERIOD.get(settings)
+    @Volatile private var enabled = ENABLED.get(settings)
+    private var scheduledSweep: Scheduler.Cancellable? = null
+    private var isClusterManager = false
+
+    init {
+        clusterService.addListener(this)
+        clusterService.clusterSettings.addSettingsUpdateConsumer(SWEEP_PERIOD) {
+            sweepPeriod = it
+            if (clusterService.state().nodes.isLocalNodeElectedClusterManager && enabled) reschedule()
+        }
+        clusterService.clusterSettings.addSettingsUpdateConsumer(ENABLED) {
+            enabled = it
+            if (clusterService.state().nodes.isLocalNodeElectedClusterManager) {
+                if (enabled) reschedule() else cancel()
+            }
+        }
+    }
+
+    override fun clusterChanged(event: ClusterChangedEvent) {
+        if (this.isClusterManager != event.localNodeClusterManager()) {
+            this.isClusterManager = event.localNodeClusterManager()
+            if (this.isClusterManager && enabled) {
+                reschedule()
+            } else {
+                cancel()
+            }
+        }
+    }
+
+    private fun reschedule() {
+        cancel()
+        scheduledSweep = threadPool.scheduleWithFixedDelay(
+            { scope.launch { withTimeout(SWEEP_TIMEOUT_MS) { sweepOrphanedAlerts() } } },
+            sweepPeriod,
+            ThreadPool.Names.MANAGEMENT
+        )
+        logger.info("Orphaned alert sweeper scheduled with period: $sweepPeriod")
+    }
+
+    private fun cancel() {
+        scheduledSweep?.cancel()
+        scheduledSweep = null
+    }
+
+    internal suspend fun sweepOrphanedAlerts() {
+        try {
+            if (!clusterService.state().routingTable().hasIndex(ALERT_INDEX)) {
+                logger.debug("Alert index does not exist, skipping sweep")
+                return
+            }
+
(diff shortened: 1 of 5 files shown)


alerting/src/main/kotlin/org/opensearch/alerting/alerts/OrphanedAlertSweeper.kt
```diff
@@ -0,0 +1,180 @@
...
+ * cluster manager node and cleans up any orphaned alerts that postDelete missed, ensuring the
+ * system eventually converges to a consistent state.
+ *
+ * Runs only on the cluster manager node to avoid duplicate work.
+ */
+class OrphanedAlertSweeper(
+    settings: Settings,
+    private val client: Client,
+    private val threadPool: ThreadPool,
+    private val clusterService: ClusterService
```
lezzago, line 57: **Coroutine scope can be permanently broken by an unhandled failure**

This uses the default `Job()`. If any uncaught exception escapes `withTimeout(...)` in `reschedule()` (e.g. a `TimeoutCancellationException` thrown after the inner `try/catch` in `sweepOrphanedAlerts` returns, or any throwable raised before the try block is entered), the scope's job is cancelled and **all future scheduled sweeps silently no-op** — `scope.launch { ... }` on a cancelled scope is a no-op, and the user gets no signal that the sweeper has stopped working.

Suggestions:
- Use `SupervisorJob()`: `CoroutineScope(SupervisorJob() + Dispatchers.IO)` so a single failure doesn't tear down the scope.
- Or wrap the `withTimeout` call in a try/catch at the launch site so timeout/cancellation never propagates out of the coroutine.
- Add a `CoroutineExceptionHandler` for visibility either way.

alerting/src/main/kotlin/org/opensearch/alerting/alerts/OrphanedAlertSweeper.kt
```diff
@@ -0,0 +1,180 @@
...
+    private val scope = CoroutineScope(Dispatchers.IO)
+
+    @Volatile private var sweepPeriod = SWEEP_PERIOD.get(settings)
+    @Volatile private var enabled = ENABLED.get(settings)
+    private var scheduledSweep: Scheduler.Cancellable? = null
+    private var isClusterManager = false
+
+    init {
+        clusterService.addListener(this)
+        clusterService.clusterSettings.addSettingsUpdateConsumer(SWEEP_PERIOD) {
```
lezzago, line 74: **Boot-time race: `init` doesn't schedule a sweep**

Scheduling is driven solely by `clusterChanged`. If the local node is already the elected cluster manager at the moment this listener is registered — uncommon but possible during a fast restart of a single-CM cluster, or when the plugin is installed on an already-running CM — `clusterChanged` may not fire a transition for that election and `isClusterManager` stays `false` forever, so the sweeper never starts.

`JobSweeper` avoids this via `LifecycleListener.afterStart()` (see `JobSweeper.kt:124`). I'd mirror that pattern here:
- Implement `LifecycleListener`, register with `clusterService.addLifecycleListener(this)`, and bootstrap from `afterStart()` by checking `clusterService.state().nodes.isLocalNodeElectedClusterManager`.
- Also pair it with `beforeStop`/`beforeClose` to cancel the schedule, remove the cluster listener, and cancel `scope` — currently nothing tears these down on node shutdown.

alerting/src/main/kotlin/org/opensearch/alerting/alerts/OrphanedAlertSweeper.kt
toepkerd: Updated to use SupervisorJob + CoroutineExceptionHandler

alerting/src/main/kotlin/org/opensearch/alerting/alerts/OrphanedAlertSweeper.kt
toepkerd: Updated to make the sweeper also implement LifecycleListener

alerting/src/main/kotlin/org/opensearch/alerting/alerts/OrphanedAlertSweeper.kt
```diff
@@ -0,0 +1,204 @@
...
+            logger.error("Error during orphaned alert sweep", e)
+        }
+    }
+
+    // retrieves the monitor IDs referenced in the .opendistro-alerting-alerts index's alerts
+    // the possibility here is that some of these monitor IDs might no longer exist
+    private suspend fun getAlertMonitorIds(): Set<String> {
+        val agg = TermsAggregationBuilder("monitor_ids")
+            .field(Alert.MONITOR_ID_FIELD)
+            .size(1000)
```
AWSHurneyt, line 175: Is there a reason size is 1000 here? 

This is an edge case, but the plugin does support the `ALERTING_MAX_MONITORS` cluster setting to potentially have more than 1000 monitors.
https://github.com/opensearch-project/alerting/blob/main/alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt#L35

core/src/main/kotlin/org/opensearch/alerting/core/JobSweeper.kt
```diff
@@ -351,6 +353,13 @@ class JobSweeper(
             }
             searchAfter = response.hits.lastOrNull()?.seqNo
         }
+
+        // Deschedule any jobs that are in sweptJobs but no longer exist in the config index
+        currentJobs.keys.filter { shardNodes.isOwningNode(it) && !seenJobIds.contains(it) }.forEach {
```
engechas, line 358: Is there a possibility that we deschedule valid jobs due to a partial shard failure in the search with this logic?

There should be an existing mechanism that handles descheduling jobs once they are deleted - I'm wondering why we need this extra layer of handling in the sweeper rather than relying on the DeleteMonitor flow

alerting/src/main/kotlin/org/opensearch/alerting/alerts/OrphanedAlertSweeper.kt
```diff
@@ -0,0 +1,204 @@
...
+        )
+        logger.info("Orphaned alert sweeper scheduled with period: $sweepPeriod")
+    }
+
+    private fun cancel() {
+        scheduledSweep?.cancel()
+        scheduledSweep = null
+    }
+
+    internal suspend fun sweepOrphanedAlerts() {
```
engechas, line 128: Do we know how alerts are getting orphaned? We shouldn't have orphans if the delete logic is working correctly so it seems like a bandaid to add a sweeper rather than closing the gaps in the delete logic (if any)

I would advocate that we should fix any bugs that lead to orphaned alerts rather than relying on an out-of-band cleanup process

alerting/src/main/kotlin/org/opensearch/alerting/alerts/OrphanedAlertSweeper.kt
```diff
@@ -0,0 +1,204 @@
...
+        val searchRequest = SearchRequest(ScheduledJob.SCHEDULED_JOBS_INDEX).source(searchSource)
+
+        val response: SearchResponse = client.suspendUntil { search(searchRequest, it) }
+        return response.hits.hits.map { it.id }.toSet()
+    }
+
+    companion object {
+        val SWEEP_PERIOD = AlertingSettings.ORPHANED_ALERT_SWEEP_PERIOD
+        val ENABLED = AlertingSettings.ORPHANED_ALERT_SWEEP_ENABLED
+        const val SWEEP_TIMEOUT_MS = 60_000L
```
AWSHurneyt, line 202: Would it make sense to use one of the cluster settings here? E.g., `BULK_TIMEOUT`?
https://github.com/opensearch-project/alerting/blob/main/alerting/src/main/kotlin/org/opensearch/alerting/settings/AlertingSettings.kt#L120

alerting/src/main/kotlin/org/opensearch/alerting/alerts/OrphanedAlertSweeper.kt
```diff
@@ -0,0 +1,204 @@
...
+            if (orphanedMonitorIds.isEmpty()) {
+                logger.debug("No orphaned alerts found")
+                return
+            }
+
+            logger.info("Found ${orphanedMonitorIds.size} orphaned monitor(s) with active alerts: $orphanedMonitorIds")
+
+            // Move orphaned alerts to history
+            for (monitorId in orphanedMonitorIds) {
+                try {
```
lezzago, line 158: 🚫 **BLOCKING — sweeper can incorrectly delete active workflow (chained) alerts**

Per `AlertService.kt:949`, chained alerts can carry `monitor_id=""` and use `workflow_id` instead. The terms aggregation in `getAlertMonitorIds()` will return `""` as a valid bucket key. That empty string is then:

1. Looked up via `idsQuery().addIds("")` against `SCHEDULED_JOBS_INDEX` — returns no hits.
2. Therefore added to `orphanedMonitorIds`.
3. Passed to `AlertMover.moveAlerts(client, "", null)`, which runs `termQuery(MONITOR_ID_FIELD, "")` (`AlertMover.kt:62`) — matching **every chained alert in the cluster whose monitor_id field is empty**.

The result: chained alerts for **active**, non-deleted workflows get moved to history with `state=DELETED` on the very first sweep cycle. This is a regression risk, not just a coverage gap.

**Required fix (one of):**
- Filter the terms agg to non-empty monitor IDs (e.g. add a `must(existsQuery)` + `mustNot(termQuery("monitor_id", ""))` pre-filter), AND
- Add a parallel pass that aggregates `Alert.WORKFLOW_ID_FIELD`, reconciles against workflow IDs in `SCHEDULED_JOBS_INDEX`, and invokes the workflow variant of `AlertMover.moveAlerts` for genuine workflow orphans.

At minimum, please verify with a test that creates a workflow with chained alerts, leaves the workflow alive, runs the sweeper, and asserts the chained alerts are NOT moved to history.

core/src/main/kotlin/org/opensearch/alerting/core/JobSweeper.kt
lezzago, line 358: **Nit: redundant `isOwningNode` check**

Lines 304-307 earlier in `sweepShard` already remove non-owning jobs from `currentJobs`:
```kotlin
currentJobs.keys.filterNot { shardNodes.isOwningNode(it) }.forEach {
    scheduler.deschedule(it)
    currentJobs.remove(it)
}
```
By the time control reaches this new block, every key in `currentJobs` is owned by this node. The `shardNodes.isOwningNode(it) &&` predicate is dead. Either drop it for clarity or add a comment noting it's defensive.

alerting/src/main/kotlin/org/opensearch/alerting/alerts/OrphanedAlertSweeper.kt
```diff
@@ -0,0 +1,204 @@
...
+        scheduledSweep = threadPool.scheduleWithFixedDelay(
+            { scope.launch { withTimeout(SWEEP_TIMEOUT_MS) { sweepOrphanedAlerts() } } },
+            sweepPeriod,
+            ThreadPool.Names.MANAGEMENT
+        )
+        logger.info("Orphaned alert sweeper scheduled with period: $sweepPeriod")
+    }
+
+    private fun cancel() {
+        scheduledSweep?.cancel()
```
lezzago, line 124: **Concurrent sweeps possible if cluster-manager-ship oscillates**

`cancel()` only cancels the `Scheduler.Cancellable` — it doesn't track or cancel the in-flight coroutine `Job` returned by `scope.launch { ... }` at line 117. If CM-ship flips off and back on (rare but possible during partition flapping or network blips), the previous sweep's coroutine continues running while a new schedule is armed. Two sweeps can call `AlertMover.moveAlerts` for the same monitor IDs in parallel.

`AlertMover` uses `versionType=EXTERNAL_GTE` so the worst case is duplicated history-index writes / version-conflict log noise — no data corruption — but it's worth tightening:

```kotlin
private var inFlightSweep: Job? = null

private fun reschedule() {
    cancel()
    scheduledSweep = threadPool.scheduleWithFixedDelay(
        { inFlightSweep = scope.launch { withTimeout(SWEEP_TIMEOUT_MS) { sweepOrphanedAlerts() } } },
        sweepPeriod,
        ThreadPool.Names.MANAGEMENT
    )
}

private fun cancel() {
    scheduledSweep?.cancel()
    scheduledSweep = null
    inFlightSweep?.cancel()
    inFlightSweep = null
}
```

alerting/src/main/kotlin/org/opensearch/alerting/alerts/OrphanedAlertSweeper.kt
```diff
@@ -0,0 +1,204 @@
...
+                    logger.error("Failed to clean up orphaned alerts for monitor [$monitorId]", e)
+                }
+            }
+        } catch (e: Exception) {
+            logger.error("Error during orphaned alert sweep", e)
+        }
+    }
+
+    // retrieves the monitor IDs referenced in the .opendistro-alerting-alerts index's alerts
+    // the possibility here is that some of these monitor IDs might no longer exist
```
lezzago, line 171: **60s timeout may be too tight for the first sweep after an outage**

`AlertMover.moveAlerts` is called sequentially in a `for` loop, all under one `withTimeout(SWEEP_TIMEOUT_MS)`. Each call is roughly 1 search + 1 bulk index + 1 bulk delete (3 round-trips). On the first sweep after an extended outage where many orphans accumulated, ~100 orphans could blow the 60s budget and lose progress mid-loop — and the next cycle starts from scratch with the same overload.

Options:
- Cap orphans-per-sweep (e.g. process up to N per cycle, drain across cycles).
- Make timeout proportional to orphan count.
- Process moves with bounded parallelism (e.g. `coroutineScope { orphans.chunked(K).forEach { ... } }`).
- Or expose `SWEEP_TIMEOUT_MS` as a cluster setting alongside the period/enabled flags so operators can tune it without a code change.

core/src/main/kotlin/org/opensearch/alerting/core/JobSweeper.kt
toepkerd: There is an existing mechanism to deschedule monitors on monitor deletion, here in [JobSweeper](https://github.com/opensearch-project/alerting/blob/b7dcc1c70abc9564ed8258599d50213998c6ceb3/core/src/main/kotlin/org/opensearch/alerting/core/JobSweeper.kt#L186). 

The goal of this PR is to temporarily remediate a bug where this callback simply does not happen for a reason that is still being root caused. This is meant to be a short-medium term band-aid fix that maintains correct Alerting state for Alerting users while we root cause the true bug. Please see the description if you'd like more details.

toepkerd: To address the partial shard failure case, in the while loop above this, during the paginated shard search, the entire sweep returns early if any shards fail ([reference](https://github.com/opensearch-project/alerting/blob/b7dcc1c70abc9564ed8258599d50213998c6ceb3/core/src/main/kotlin/org/opensearch/alerting/core/JobSweeper.kt#L336)). This means this descheduling block at the end is only reached if all shards succeed. If any shard at all failed, the retry is deferred to the next sweep cycle.

alerting/src/main/kotlin/org/opensearch/alerting/alerts/OrphanedAlertSweeper.kt
toepkerd: The proposed mitigation to cap orphans per sweep is already in place, just not very explicitly. `moveAlerts()` runs a search on the active alerts index without explicitly passing a size param ([reference](https://github.com/opensearch-project/alerting/blob/b7dcc1c70abc9564ed8258599d50213998c6ceb3/alerting/src/main/kotlin/org/opensearch/alerting/alerts/AlertMover.kt#L68)). This means it defaults to returning only 10 orphaned Alerts even within a single zombie Monitor. In this respect, this sweeper already cleans at most 10 alerts per orphaned monitor per sweep.

Even if within a 10 orphaned Alerts cleanup batch, the timeout can still be hit in the worst case but that should not be an issue:
- case: timeout hits after both the copy and the delete bulk requests are sent. In this case, the timeout simply prevents the thread from reading in the result, but the task is done. On the next sweep, the alerts will have already been cleaned up, and the sweeper moves on without issue
- case: timeout happens between the copy and the delete bulk requests. In this case, there are now 2 copies of the orphaned alerts in the active and history indices. On the next sweep, these are found again, but with EXTERNAL_GTE version checks, idempotency is enforced and no duplication happens, progress is still made. The only way this case scenario can stall progress is if on every single sweep, the timeout hits between the copy and delete bulk requests, which are back to back lines.
- case: timeout happens before copy or delete requests, stalls on the search for orphaned Active alerts. If the search to the active alerts concrete index takes more than 60s, the cluster has bigger problems than the alert sweeper not catching all Alerts. Once the cluster can breathe again, the above 2 cases kick in and make progress.

PR #2179, opened 2026-06-16
fix(alerting): Avoid ScriptService fallback when multi-tenant trigger eval enabled

eirsep:
When multiTenantTriggerEvalEnabled=true and remote trigger evaluation cannot run (due to empty input results from a failed search), do not fall back to local ScriptService.compile. Instead, directly construct the trigger result with the search error attached. Background execution sets triggered=true (error alert), dryrun sets triggered=false.


eirsep:
CI failure is unrelated to the PR's change. It's a flaky integration test — DocumentMonitorRunnerIT failing because the .opendistro-alerting-alert* index shards aren't ready when the test tries to search them (500 "all 
shards failed" during a getAlerts call). It's a timing/race condition in the test cluster setup, not a code logic failure.

AWSHurneyt (approved):

vikhy-aws (approved):

PR #2225, opened 2026-08-27
Apply Index Monitor API input validation to the Execute Monitor API

jmsusanto:
### Description
Add validation to Execute Monitor API

### Behavior change
The Execute Monitor API's pre-flight input check now fails the request on any search failure, not only security exceptions. Previously, an inline dry-run against a nonexistent (or otherwise unqueryable) index returned HTTP 200 with the error captured in input_results; it now returns an error response. API consumers relying on the old dry-run debugging behavior should be aware of this.

### Check List
- [ ] New functionality includes testing.
- [ ] Commits are signed per the DCO using `--signoff`.

### DIFF
--- a/alerting/src/main/kotlin/org/opensearch/alerting/resthandler/RestExecuteMonitorAction.kt
+++ b/alerting/src/main/kotlin/org/opensearch/alerting/resthandler/RestExecuteMonitorAction.kt
@@ -9,8 +9,10 @@ import org.apache.logging.log4j.LogManager
 import org.opensearch.alerting.AlertingPlugin
 import org.opensearch.alerting.action.ExecuteMonitorAction
 import org.opensearch.alerting.action.ExecuteMonitorRequest
+import org.opensearch.alerting.alerts.AlertIndices
 import org.opensearch.common.unit.TimeValue
 import org.opensearch.commons.alerting.model.Monitor
+import org.opensearch.commons.alerting.model.ScheduledJob
 import org.opensearch.commons.alerting.util.AlertingException
 import org.opensearch.core.xcontent.XContentParser.Token.START_OBJECT
 import org.opensearch.core.xcontent.XContentParserUtils.ensureExpectedToken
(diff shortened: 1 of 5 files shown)


riysaxen-amzn:
1. **Behavior change for release notes**: the pre-flight search fails the request for *any* search failure, not just security exceptions — an inline dry-run against a nonexistent index now returns an error response instead of a 200 with the error in `input_results`. Worth a line in the PR description since it changes dry-run debugging behavior for API consumers.

2. **Follow-up question**: `validateDataSources` checks `queryIndex`/`findingsIndex`/`alertsIndex` but not `findingsIndexPattern`/`alertsHistoryIndex`/`alertsHistoryIndexPattern` (consistent with the existing create-path check). If that was assessed as acceptable during verification, ignore; otherwise might deserve a tracking issue covering both handlers.

riysaxen-amzn (approved):

jmsusanto:
> 1. **Behavior change for release notes**: the pre-flight search fails the request for _any_ search failure, not just security exceptions — an inline dry-run against a nonexistent index now returns an error response instead of a 200 with the error in `input_results`. Worth a line in the PR description since it changes dry-run debugging behavior for API consumers.
> 2. **Follow-up question**: `validateDataSources` checks `queryIndex`/`findingsIndex`/`alertsIndex` but not `findingsIndexPattern`/`alertsHistoryIndex`/`alertsHistoryIndexPattern` (consistent with the existing create-path check). If that was assessed as acceptable during verification, ignore; otherwise might deserve a tracking issue covering both handlers.

1. Good catch, this is intentional. The pre-flight search now surfaces any search failure (not just security
  exceptions), so an inline dry-run against a nonexistent index returns an error response instead of a 200 with the error in input_results. I'veadded a note to the PR description calling this out for consumers.
  
  2. This was intentional to stay consistent with the existing create-path check in RestIndexMonitorAction, which validates the same three indices (queryIndex/findingsIndex/alertsIndex). Extending both handlers to cover findingsIndexPattern/alertsHistoryIndex/alertsHistoryIndexPattern is worth doing separately

eirsep (approved):

